PR #34, opened 2017-01-20
Adds PUBLIC_KEY_SCHEMA and PUBLIC_KEYVAL_SCHEMA

lukpueh:
These schemas require the private portion of the key value to be
either omitted or an empty string.


vladimir-v-diaz:
LGTM!

PR #48, opened 2017-08-10
Implement TAP 9 (Mandatory metadata signing schemes)

vladimir-v-diaz:
[TAP 9](https://github.com/theupdateframework/taps/blob/master/tap9.md) removes the `method` attribute from signed metadata to prevent malicious actors from controlling it.  Instead, the signature scheme is now specified in the "signed" portion of metadata.

This pull request refactors a few functions to accept a signature scheme argument.  In addition, the signature scheme is now included in the "keys" field of signed metadata.

TAP 9 and this implementation are backwards incompatible.


securesystemslib/pyca_crypto_keys.py
```diff
@@ -264,113 +265,130 @@ def create_rsa_signature(private_key, data):
...
+      # decoded successfully.
+      except ValueError: #pragma: no cover
+        raise securesystemslib.exceptions.CryptoError('The private key'
+          ' (in PEM format) could not be deserialized.')
+
+      # 'TypeError' raised if a password was given and the private key was not
+      # encrypted, or if the key was encrypted but no password was supplied.
+      # Note: A passphrase or password is not used when generating
+      # 'private_key', since it should not be encrypted.
+      except TypeError: #pragma: no cover
```
lukpueh: Any reason for the `#pragma: no cover` here and with the `ValueError` above?

securesystemslib/pyca_crypto_keys.py
```diff
@@ -264,113 +265,130 @@ def create_rsa_signature(private_key, data):
...
+        raise securesystemslib.exceptions.CryptoError('The private key was'
+          ' unexpectedly encrypted.')
+
+      # 'cryptography.exceptions.UnsupportedAlgorithm' raised if the serialized
+      # key is of a type that is not supported by the backend, or if the key is
+      # encrypted with a symmetric cipher that is not supported by the backend.
+      except cryptography.exceptions.UnsupportedAlgorithm: #pragma: no cover
+        raise securesystemslib.exceptions.CryptoError('The private key is'
+          ' encrypted with a unsupported algorithm.')
```
lukpueh: +n

PR #57, opened 2017-09-15
PEM format quickfix #54

lukpueh:
This quickfix changes the PEM format from PKSC8 to PKSC5 (TraditionalOpenSSL) in pyca_crypto's variant of `create_rsa_encrypted_pem`.
PKSC5 has the PEM headers expected by other PEM parsing functions, e.g. `is_pem_private` and `extract_pem`, which would break without this fix. The PR also updates the unittests.

However, we should consider a smarter solution as discussed in #54.

### DIFF
--- a/securesystemslib/pyca_crypto_keys.py
+++ b/securesystemslib/pyca_crypto_keys.py
@@ -540,7 +540,7 @@ def create_rsa_encrypted_pem(private_key, passphrase):
 
   encrypted_pem = \
     private_key.private_bytes(encoding=serialization.Encoding.PEM,
-    format=serialization.PrivateFormat.PKCS8,
+    format=serialization.PrivateFormat.TraditionalOpenSSL,
     encryption_algorithm=serialization.BestAvailableEncryption(passphrase.encode('utf-8')))
 
   return encrypted_pem.decode()
--- a/securesystemslib/pyca_crypto_keys.py
+++ b/securesystemslib/pyca_crypto_keys.py
@@ -473,11 +473,13 @@ def verify_rsa_signature(signature, signature_scheme, public_key, data):
 def create_rsa_encrypted_pem(private_key, passphrase):
   """
   <Purpose>
-    Return a string in PEM format, where the private part of the RSA key is
-    encrypted.  The private part of the RSA key is encrypted by the Triple
-    Data Encryption Algorithm (3DES) and Cipher-block chaining (CBC) for the
-    mode of operation.  Password-Based Key Derivation Function 1 (PBKF1) + MD5
-    is used to strengthen 'passphrase'.
+    Return a string in PEM format (TraditionalOpenSSL), where the
+    private part of the RSA key is encrypted using the best available
+    encryption for a given key's backend. This is a curated encryption choice
+    and the algorithm may change over time.
+
+    c.f. cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/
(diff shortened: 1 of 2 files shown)


securesystemslib/pyca_crypto_keys.py
```diff
@@ -540,7 +540,7 @@ def create_rsa_encrypted_pem(private_key, passphrase):
 
   encrypted_pem = \
     private_key.private_bytes(encoding=serialization.Encoding.PEM,
-    format=serialization.PrivateFormat.PKCS8,
+    format=serialization.PrivateFormat.TraditionalOpenSSL,
```
vladimir-v-diaz: The docstring for this function should be updated, not only to mention the new PEM format but to correct it. The docstring incorrectly references PyCrypto and 3DES, etc. (this appears to be a copy+paste that wasn't edited).

tests/test_keys.py
```diff
@@ -682,6 +683,13 @@ def test_extract_pem(self):
                                    private_pem=False)
     self.assertTrue(securesystemslib.formats.PEMRSA_SCHEMA.matches(public_pem))
 
+    # Test encrypted private pem
+    private_pem_encrypted = KEYS.create_rsa_encrypted_pem(private_pem, "pw")
+    private_pem_encrypted = KEYS.extract_pem(private_pem_encrypted,
+                                   private_pem=True)
```
vladimir-v-diaz: I know that there are still many instances of this style of indentation in the codebase, but we should try to use the recently added indentation rules of the code style guidelines from now on.

Here is what the guidelines says about indentation for this particular case:

> Make sure to indent the continued line appropriately; double-indentation (4 spaces) is prefered for line continuation, to distinguish it from functional indentation.

Correct me if I'm wrong, but this line should look as follows:
```
KEYS.extract_pem(private_pem_encrypted,
    private_pem=True)
```

tests/test_keys.py
```diff
@@ -737,9 +745,11 @@ def test_is_pem_private(self):
     # Test for a valid PEM string.
     private_pem = self.rsakey_dict['keyval']['private']
     private_pem_ec = self.ecdsakey_dict['keyval']['private']
+    private_pem_encrypted = KEYS.create_rsa_encrypted_pem(private_pem, "pw")
```
vladimir-v-diaz: Maybe name it `private_pem_encrypted_rsa` to make it clear that it's an RSA PEM, similar to `private_pem_ec`?

securesystemslib/pyca_crypto_keys.py
```diff
@@ -473,11 +473,13 @@ def verify_rsa_signature(signature, signature_scheme, public_key, data):
...
   """
   <Purpose>
-    Return a string in PEM format, where the private part of the RSA key is
-    encrypted.  The private part of the RSA key is encrypted by the Triple
-    Data Encryption Algorithm (3DES) and Cipher-block chaining (CBC) for the
-    mode of operation.  Password-Based Key Derivation Function 1 (PBKF1) + MD5
-    is used to strengthen 'passphrase'.
+    Return a string in PEM format (TraditionalOpenSSL), where the
+    private part of the RSA key is encrypted using the best available
+    encryption for a given key’s backend. This is a curated encryption choice
```
vladimir-v-diaz: A non-ASCII character is used on this line, but these modules don't declare a source code encoding.  Probably easiest not to remove it.

lukpueh:
Oops, removed non-ascii character, fixed wrong variable, rebased and force-pushed. Should be fine now.

PR #58, opened 2017-09-15
Discontinue unittests in Python2.6 (tox)

lukpueh:
It seems that Travis CI has discontinued out-of-the box Python2.6 builds in a [recent update of their trusty boxes](https://docs.travis-ci.com/user/build-environment-updates/2017-09-06/). Let's seize the opportunity to discontinue Python2.6 unittesting.

This should probably entail an update of the [version specification in TUF](https://github.com/theupdateframework/tuf/blob/develop/setup.py#L98).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py26, py27, py35
+envlist = py27, py35
 
 [testenv]
 changedir = tests
@@ -16,13 +16,5 @@ commands =
 
 deps =
     -r{toxinidir}/requirements.txt
-
-install_command = pip install --pre {opts} {packages}
-
-[testenv:py26]
-deps =
-    -r{toxinidir}/requirements.txt
-    unittest2
-    importlib
 
 install_command = pip install --pre {opts} {packages}


vladimir-v-diaz:
On the [Travis page](https://docs.travis-ci.com/user/build-environment-updates/2017-09-06/) that you link, I can't seem to find the change where Python 2.6 is removed.  Their documentation stills includes [instructions on how to specify python 2.6](https://docs.travis-ci.com/user/languages/python/#Specifying-Python-versions), and I wonder if the reason why we're failing to install python2.6 on Travis is because we don't do so correctly in the `.travis.yml` file.  [We only specify "3.5"](https://github.com/secure-systems-lab/securesystemslib/blob/master/.travis.yml).

Nevertheless, [pip recently dropped python 2.6](https://github.com/pypa/pip/blob/ef5317aad5fba1b82f22cd8b2b0f3da90855ae82/news/4343.removal), so we should follow suit.

lukpueh:
I didn't find anything in the changelog either, I just assumed that the fails were related to the travis update. I didn't look for the cause any further, since I figured dropping 2.6 was overdue anyway.

PR #61, opened 2017-09-22
Upgrade pyca/cryptography version to >= 2.0.3

vladimir-v-diaz:


baloyan:
Pretty big PR, it is very hard to review (kidding).
I just would like to know is there any reason why you do not use the latest version of `cryptography` library ? I found it here: https://pypi.python.org/pypi/cryptography

If there is a reason, I think it would be better to provide it in a summary.

vladimir-v-diaz:
Thanks for the review!

cryptography 2.0.3 was the latest version when this pull request was merged.  As you've noticed, they have since released version 2.1.2.

Fortunately, the pyup.io badge correctly indicates that we have one pending update: https://github.com/secure-systems-lab/securesystemslib/blob/master/README.rst

However, I am waiting for the pyup.io bot to open a pull request to upgrade our dev dependencies.

PR #68, opened 2017-10-13
Add pyup.io badge to README.rst

vladimir-v-diaz:
**Fixes issue #**:

The issue tracker does not have an issue for this task.

Display a pyup.io badge on the README.  We use pyup.io to make sure all of our dependencies are up to date.

This pull request also fixes a typo in the overview section.

requirements**:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,6 +7,10 @@ Secure Systems Library
 .. image:: https://coveralls.io/repos/github/secure-systems-lab/securesystemslib/badge.svg?branch=master
    :target: https://coveralls.io/github/secure-systems-lab/securesystemslib?branch=master
 
+.. image:: https://pyup.io/repos/github/secure-systems-lab/securesystemslib/shield.svg
+     :target: https://pyup.io/repos/github/secure-systems-lab/securesystemslib/
+     :alt: Updates
+
 A library that provides cryptographic and general-purpose functions for Secure
 Systems Lab projects at NYU.  The routines are general enough to be usable by
 other projects.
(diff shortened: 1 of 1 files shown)


vladimir-v-diaz:
`dev-requirements.txt` pins our project's dependencies and ensures a consistent installation environment amongst the developers and contributors.  All other requirement files, including `setup.py`, are expected to result in the installation of the latest version of a dependency (we usually specify a minimum version, e.g., cryptography >= 1.7, for a dependency that we know works with our project).

It is usually advised for Python projects to install the latest versions of its dependencies, so that's why we choose not to pin them in `setup.py`, `requirements.txt`, and `ci-requirements.txt`, etc.

In practice, what this means is that we as developers work on the code according to `dev-requirements.txt`, and use pyup.io to make sure that we are working with the latest versions of dependencies.  Since production code should not be restricted in terms of using the latest versions of dependencies, we have Travis replicate this via `ci-requirements.txt`, which tests for an installation that closely matches what someone in the wild will experience.  

This setup should allow us to catch situations where one of our dependencies introduces a backwards-incompatible change, either via a build failure with Travis or `pyup.io` letting us know that one of our dependencies is out of date.

Let me know if this is clear, or if you think something is wrong with our setup.

vladimir-v-diaz:
We should also consider a note to developers/contributors where we advice them to periodically refresh their virtual environments and sync with `dev-requirements.txt`

lukpueh:
Thanks for the clarification! I misread your `setup.py` and my python environment... sorry for that.  Maybe we could add this to a best practices repo?

Issue #122, opened 2018-03-07
Load non-encrypted private keys and toggle password prompt


lukpueh:
**Description of issue or feature request**:
The [`interface`](https://github.com/secure-systems-lab/securesystemslib/blob/v0.10.10/securesystemslib/interface.py) module provides functions to import `RSA`, `ed25510` and `ECDSA` private keys from files.

The functions take an optional password argument and if no password is passed, the caller is prompted for a password. In either case the received password is used to decrypt the key. Hence, there is no way to load a non-password-encrypted private key without getting a prompt, or to fail if the key is encrypted but no password is passed, rather than prompting for a password.

This ticket proposes the addition of an optional boolean `prompt` argument to the functions:
`import_rsa_privatekey_from_file(filepath, password=None,  scheme='rsassa-pss-sha256')`
`import_ed25519_privatekey_from_file(filepath, password=None)`
`import_ecdsa_privatekey_from_file(filepath, password=None)`

**Current behavior**:
If a password is passed, that password is used to decrypt the key.
If no password is passed, the user will be prompted for a password, which is used to decrypt the key.
Fail if the key can't be decrypted.

**Expected behavior**:
If a password is passed and prompt is False, the passed password is used to decrypt the key. 
If prompt is True, the caller is prompted for a password, which is used to decrypt the key.
If no password is passed and prompt is False, the key will be loaded as non-encrypted key.
Fail if no password is passed and the key is decrypted.
Fail if a password is passed and the key can't be decrypted.

lukpueh:
UPDATE:
The current behavior actually does allow loading non-encrypted keys. This can be achieved by passing an empty string as password, which does not trigger the prompt, but skips decryption.

IMHO, this behavior is not obvious. I think it is better to fail if an empty password is passed. If a caller wants to use the function to load a non-encrypted key, s/he should just not pass a password.

However, a user, who is prompted for a password, can only enter either a non-empty or an empty password. To give that user the option to load a non-encrypted key, we should skip decryption if the user does not enter a password, i.e. the password is an empty string.

PR #124, opened 2018-03-08
Add prompt argument to import RSA private key function

lukpueh:
**Fixes issue #**:
Fixes RSA part of #122 

Add an optional boolean `prompt` argument to `interface.import_rsa_privatekey_from_file` and change the behavior of the function like so:

- If `password` is passed use passed `password` for decryption.
- If `prompt` is `True` use entered password for decryption.
- If no password is passed or entered, or if the entered password is an empty string, omit decryption.
- Passing and prompting for a password is not possible.

This PR further:
- adds `mock` as dev/ci requirement for supported Python versions <3.3, i.e. 2.7, to mock password prompts, and
- adopts and adds related unit tests.

See commit messages, code comments or secure-systems-lab/securesystemslib#122 for more details.

requirements**:

### DIFF
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -237,11 +237,20 @@ def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
 
 
 def import_rsa_privatekey_from_file(filepath, password=None,
-    scheme='rsassa-pss-sha256'):
+    scheme='rsassa-pss-sha256', prompt=False):
   """
   <Purpose>
-    Import the encrypted PEM file in 'filepath', decrypt it, and return the key
-    object in 'securesystemslib.formats.RSAKEY_SCHEMA' format.
+    Import the PEM file in 'filepath' containing the private key.
+
+    If password is passed use passed password for decryption.
+    If prompt is True use entered password for decryption.
+    If no password is passed or entered, or if the entered password is an empty
+    string, omit decryption.
+
+    Passing and prompting for a passowrd is not possible.
+
+    The returned key is an object in the
+    'securesystemslib.formats.RSAKEY_SCHEMA' format.
 
   <Arguments>
     filepath:
@@ -254,18 +263,35 @@ def import_rsa_privatekey_from_file(filepath, password=None,
     scheme:
       The signature scheme used by the imported key.
 
+    prompt:
+      If True the user is prompted for a passphrase to decrypt 'filepath'.
+      Default is False.
+
   <Exceptions>
+    ValueError, if 'password' is passed and 'prompt' is True.
+
+    ValueError, if 'password' is passed and it is an empty string.
+
     securesystemslib.exceptions.FormatError, if the arguments are improperly
     formatted.
 
-    securesystemslib.exceptions.CryptoError, if 'filepath' is not a valid
-    encrypted key file.
+    securesystemslib.exceptions.FormatError, if the entered password is
+    improperly formatted.
+
+    IOError, if 'filepath' can't be loaded.
+
+    securesystemslib.exceptions.CryptoError, if a password is available
+    and 'filepath' is not a valid key file encrypted using that password.
+
+    securesystemslib.exceptions.CryptoError, if no password is available
+    and 'filepath' is not a valid non-encrypted key file.
 
   <Side Effects>
-    The contents of 'filepath' is read, decrypted, and the key stored.
+    The contents of 'filepath' are read, optionally decrypted, and returned.
 
   <Returns>
     An RSA key object, conformant to 'securesystemslib.formats.RSAKEY_SCHEMA'.
+
   """
 
   # Does 'filepath' have the correct format?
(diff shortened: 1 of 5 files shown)


awwad:
I'll review.

It would be good to have a clear sense of what the status quo pre-PR was in the PR description. (My TUF fork in Uptane is still pretty remote, so I think the behavior is somewhat different.)

That might be too demanding. ' dunno. (:

awwad:
Consider the following alternative (which I don't say is superior or anything -- just worth considering):

optional argument 'encrypt' instead of optional argument 'prompt'


prompt vs encrypt

lukpueh:
Summarizing our conversation on the whiteboard:
- I'd like to keep `prompt` because, it should be explicit whether a function expects user input,  
- If `prompt` is True and a password is passed, we could (1) not prompt and use the passed pw, (2) prompt and override the passed pw with the entered pw, (3) fail with an argument error. Any of these make sense, so we can stick with solution (3).

lukpueh:
Maybe we should put it up for vote? @vladimir-v-diaz, @SantiagoTorres what do you guys think?

awwad:
(I think it's fine.)

vladimir-v-diaz:
Approach (3) is fine with me.

If we make `prompt` optional, how do we deal with guaranteeing consistent behavior across projects?  For instance, the TUF project might choose to set prompt=True, whereas in-toto uses prompt=False.

I suppose we can recommend that the CLI tools (that use securesystemslib) handle the option of prompting for a password.  In that case, an integrator who uses both in-toto and TUF can expect consistent behavior
by using, for example, `$ in-toto --sign --prompt-pw` and `$ repo.py --sign --prompt-pw`.

tests/test_interface.py
```diff
@@ -42,6 +42,13 @@
 else:
   import unittest2 as unittest
 
+# Use external backport 'mock' on versions under 3.3
+if sys.version_info >= (3, 3):
+  import unittest.mock as mock
+
+else:
+  import mock
+
```
awwad: (This question in no way interferes with this PR. As a point of general curiosity:
Is it better practice to test Python version numbers this way when determining what to import, or to try the first import and fail over to the second one? What if, say, Python 4.0 had been released before mock was added to unittest in Python 3.3? Should one just see if the duck quacks first?)

securesystemslib/interface.py
```diff
@@ -237,11 +237,20 @@ def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
 
 
 def import_rsa_privatekey_from_file(filepath, password=None,
-    scheme='rsassa-pss-sha256'):
+    scheme='rsassa-pss-sha256', prompt=False):
```
awwad: NBD: Has there been some discussion about whether or not False should be the default prompt value used?

- TUF tutorial text -- and some code, probably -- will need to change.
- (What are the security implications of passing the passphrase in through a function call in the interpreter rather than using the prompt functionality from `getpass`?)

securesystemslib/interface.py
```diff
@@ -277,37 +303,51 @@ def import_rsa_privatekey_from_file(filepath, password=None,
...
+    if not len(password):
+      raise ValueError('Password must be 1 or more characters')
+
+  elif prompt:
+    # Password confirmation disabled here, which should ideally happen only
+    # when creating encrypted key files (i.e., improve usability).
     # It is safe to specify the full path of 'filepath' in the prompt and not
     # worry about leaking sensitive information about the key's location.
     # However, care should be taken when including the full path in exceptions
     # and log files.
```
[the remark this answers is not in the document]
lukpueh: AFAIK, Trishank and Vlad made this decision, because in order for the prompt to show the full path, the full path must be known to the caller anyway, whereas an exception/log might reveal that information to someone who normally does not know it.

tests/test_interface.py
lukpueh: IMHO both are fine. In this case I lean towards the version checking approach, because according to Python docs, mock is "*available as unittest.mock in Python 3.3 onwards"*. So if it fails with e.g. 4.0, it would be a Python (docs) bug.
It is important though that the unittests take all supported Python versions into consideration.

There might be a minimal performance difference between exception handling and tuple comparison, but I think we can disregard that.

lukpueh:
@awwad, thanks for the review! It would be great if you could make the changes you proposed.\

@vladimir-v-diaz, `--prompt-pw` makes sense. An alternative that I've often come across is to provide a `-p/--password [<password>]` option, which can be used to either pass the password as argument or, if the password is not passed, signals the cli tool to prompt for a password.

securesystemslib/interface.py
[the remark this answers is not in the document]
vladimir-v-diaz: > because in order for the prompt to show the full path, the full path must be known to the caller anyway, whereas an exception/log might reveal that information to someone who normally does not know it.

That's correct.  Revealing the full path via an exception or log file should be avoided elsewhere. 

We want to avoid vulnerabilities like [Information exposure through an error message](https://cwe.mitre.org/data/definitions/209.html) and [full path disclosure](https://www.owasp.org/index.php/Full_Path_Disclosure).

vladimir-v-diaz:
> @vladimir-v-diaz, --prompt-pw makes sense. An alternative that I've often come across is to provide a -p/--password [<password>] option, which can be used to either pass the password as argument or, if the password is not passed, signals the cli tool to prompt for a password.

That's the approach we used in TUF's CLI tool.
https://github.com/theupdateframework/tuf/blob/develop/docs/CLI.md#cli-examples

awwad:
Recommendations applied. Vlad, does merging this affect you?

vladimir-v-diaz:
Not now, but I'll have to update the TUF code once a new version of securesystemslib is released.

PR #148, opened 2018-05-04
No (en|de)cryption of ed25519 private keys when empty password

trishankatdatadog:
**Fixes issue #**:

Presently, the library is consistent about how it handles empty passwords with regards to RSA, Ed25519, and ECDSA. The functions for RSA handles empty passwords as a signal that private keys are not to be encrypted / decrypted, whereas the functions for Ed25519 and ECDSA currently "encrypts / decrypts" them with empty passwords.

The functions for Ed25519 are now consistent with the RSA functions in the sense that empty passwords are a signal that private keys are not to be encrypted / decrypted.

requirements**:

### DIFF
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -645,10 +645,12 @@ def import_ed25519_privatekey_from_file(filepath, password=None, prompt=False):
         ' file \'' + Fore.RED + filepath + Fore.RESET + '\': ',
         confirm=False) or None
 
-    with open(filepath, 'rb') as file_object:
-      jsons = file_object.read()
-      return securesystemslib.keys.\
-             import_ed25519key_from_private_json(jsons, password=password)
+  # Finally, regardless of password, try decrypting the key, if necessary.
+  # Otherwise, load it straight from the disk.
+  with open(filepath, 'rb') as file_object:
+    json_str = file_object.read()
+    return securesystemslib.keys.\
+           import_ed25519key_from_private_json(json_str, password=password)
 
 
 
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -472,10 +472,18 @@ def generate_and_write_ed25519_keypair(filepath=None, password=None):
         confirm=True)
 
   else:
-    logger.debug('The password has been specified.  Not prompting for one')
+    logger.debug('The password has been specified. Not prompting for one.')
 
   # Does 'password' have the correct format?
   securesystemslib.formats.PASSWORD_SCHEMA.check_match(password)
+
+  # Encrypt the private key if 'password' is set.
+  if len(password):
+    ed25519_key = securesystemslib.keys.encrypt_key(ed25519_key, password)
+
+  else:
+    logger.debug('An empty password was given. '
+                 'Not encrypting the private key.')
 
   # If the parent directory of filepath does not exist,
   # create it (and all its parent directories, if necessary).
@@ -507,8 +515,7 @@ def generate_and_write_ed25519_keypair(filepath=None, password=None):
 
   # Raise 'securesystemslib.exceptions.CryptoError' if 'ed25519_key' cannot be
   # encrypted.
-  encrypted_key = securesystemslib.keys.encrypt_key(ed25519_key, password)
-  file_object.write(encrypted_key.encode('utf-8'))
+  file_object.write(ed25519_key.encode('utf-8'))
   file_object.move(filepath)
 
   return filepath
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -575,7 +575,7 @@ def import_ed25519_publickey_from_file(filepath):
 
 
 
-def import_ed25519_privatekey_from_file(filepath, password=None):
+def import_ed25519_privatekey_from_file(filepath, password=None, prompt=False):
   """
   <Purpose>
     Import the encrypted ed25519 key file in 'filepath', decrypt it, and return
@@ -593,6 +593,10 @@ def import_ed25519_privatekey_from_file(filepath, password=None):
       The password, or passphrase, to import the private key (i.e., the
       encrypted key file 'filepath' must be decrypted before the ed25519 key
       object can be returned.
+
+    prompt:
+      If True the user is prompted for a passphrase to decrypt 'filepath'.
+      Default is False.
 
   <Exceptions>
     securesystemslib.exceptions.FormatError, if the arguments are improperly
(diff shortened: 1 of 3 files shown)


trishankatdatadog:
Related to https://github.com/in-toto/in-toto/pull/205

vladimir-v-diaz:
Note: two of the unit tests are failing.
https://travis-ci.org/secure-systems-lab/securesystemslib/builds/375092872?utm_source=github_status&utm_medium=notification

trishankatdatadog:
Who knows what's going on. I'll take a look later when I find some downtime...

vladimir-v-diaz:
You can easily click on the [job failure](https://travis-ci.org/secure-systems-lab/securesystemslib/jobs/375092873) to figure what's going on.

The loaded private key doesn't match the ED25519KEY_SCHEMA.

trishankatdatadog:
I know this :) I meant I have to find the time.

securesystemslib/interface.py
```diff
@@ -607,46 +619,36 @@ def import_ed25519_privatekey_from_file(filepath, password=None):
...
-  # different keyids can be associated using supported keyid_hash_algorithms.
-  key_object['keyid_hash_algorithms'] = \
-      securesystemslib.settings.HASH_ALGORITHMS
+    # NOTE: A user who gets prompted for a password, can only signal that the
+    # key is not encrypted by entering no password in the prompt, as opposed
+    # to a programmer who can call the function with or without a 'password'.
+    # Hence, we treat an empty password here, as if no 'password' was passed.
+    password = get_password('Enter a password for an encrypted RSA'
+        ' file \'' + Fore.RED + filepath + Fore.RESET + '\': ',
+        confirm=False) or None
```
vladimir-v-diaz: This `or None` is confusing.  We have a guideline (although specific for outputting text) that hints against this sort of short circuit evaluation.

https://github.com/secure-systems-lab/code-style-guidelines#never-use-short-circuit-evaluation-to-produce-output

vladimir-v-diaz: `password` should be None if the user doesn't enter a password, correct?

securesystemslib/keys.py
```diff
@@ -1586,6 +1589,40 @@ def is_pem_private(pem, keytype='rsa'):
...
+    # 'securesystemslib.exceptions.CryptoError' if the decryption fails.
+    key_object = securesystemslib.keys.\
+                 decrypt_key(json_str.decode('utf-8'), password)
+
+  else:
+    logger.debug('No password was given. Attempting to import an'
+        ' unencrypted file.')
+    key_object = json.loads(json_str.decode('utf-8'))
+
+  # Raise an exception if an unexpected key type is imported.
```
vladimir-v-diaz: Do we have an ed25519 schema for this check?
https://github.com/secure-systems-lab/securesystemslib/blob/1c4161e0a8b249e42f4f8d02ffef49a45a2e4006/securesystemslib/formats.py#L308-L315

securesystemslib/interface.py
trishankatdatadog: I'm afraid this code is already there for RSA, too...

https://github.com/secure-systems-lab/securesystemslib/blob/1c4161e0a8b249e42f4f8d02ffef49a45a2e4006/securesystemslib/interface.py#L328-L330

vladimir-v-diaz: Hmm, true.  I see that @awwad  had a similar comment against it [here](https://github.com/secure-systems-lab/securesystemslib/pull/124#discussion_r174238179).  I recommend we avoid copying it.

trishankatdatadog: Got it. What do you recommend?

vladimir-v-diaz: I'm not sure why @lukpueh added the `or None`.  Is it even needed?  If the user doesn't enter anything, the empty string is considered None, right?

@awwad, what do you think?

trishankatdatadog: Just my 0.02 BTC, but on a separate note, there's a lot of code repeated between both of these functions for RSA and Ed25519 (and, in the future, for ECDSA) that really should be abstracted out and shared...

awwad: get_password() would return an empty string, and later code tests for `None`, so yes, the `or None` (or some other, more intuitive construction) is necessary.

vladimir-v-diaz: Necessary?  I'm confused, isn't the empty string and None similar in Boolean equivalence?  Is this really necessary, or is it needed for clarity (which I don't find clear at all)?

awwad: IIRC, there are specific `is None` tests in later code, but it's been a few weeks, so perhaps I'm misremembering.

vladimir-v-diaz: @trishankatdatadog You can add a line to set password to None if len(password) == 0.  That should make it clear.

securesystemslib/interface.py
```diff
@@ -643,7 +643,12 @@ def import_ed25519_privatekey_from_file(filepath, password=None, prompt=False):
     # Hence, we treat an empty password here, as if no 'password' was passed.
     password = get_password('Enter a password for an encrypted RSA'
         ' file \'' + Fore.RED + filepath + Fore.RESET + '\': ',
-        confirm=False) or None
+        confirm=False)
+
+    # If user sets an empty string for the password, explicitly set the
+    # password to None, because some functions may expect this later.
+    if len(password) == 0: # pragma: no cover
+      password = None
```
vladimir-v-diaz: 👍

securesystemslib/keys.py
```diff
@@ -1605,7 +1605,15 @@ def import_ed25519key_from_private_json(json_str, password=None):
...
-    key_object = json.loads(json_str.decode('utf-8'))
+    try:
+      key_object = json.loads(json_str.decode('utf-8'))
+    # If the JSON could not be decoded, it is very likely, but not necessarily,
+    # due to a non-empty password.
+    except ValueError:
+      raise securesystemslib.exceptions\
+            .CryptoError('Malformed Ed25519 key JSON, '
+                         'possibly due to encryption, '
+                         'but no password provided?')
```
vladimir-v-diaz: This is a good error message for now, and can address the root cause in a future pull request.  One way to avoid parsing difficulty with these  key files is to include additional fields in the JSON that indicate whether the key data is encrypted, which type of encryption is used, etc.

For example: 

```Python
{
"encrypted": True
"algo": "AES"
"hashing": sha3
"key data": "XDSFSD#$@cx..."
}
```

securesystemslib/interface.py
```diff
@@ -472,7 +472,7 @@ def generate_and_write_ed25519_keypair(filepath=None, password=None):
         confirm=True)
 
   else:
-    logger.debug('The password has been specified.  Not prompting for one')
+    logger.debug('The password has been specified. Not prompting for one.')
```
vladimir-v-diaz: Oh, you're one of those single space propagandists. :P
https://arstechnica.com/science/2018/05/two-spaces-after-period-are-better-than-one-except-maybe-they-arent-study-finds/

vladimir-v-diaz:
Thanks for the pull request and for fixing the unit test failures!

PR #156, opened 2018-09-26
Support normalization of line endings

ofek:
that is provided, the better.

**Fixes issue #**:

No cross-platform support

requirements**:

### DIFF
--- a/securesystemslib/hash.py
+++ b/securesystemslib/hash.py
@@ -41,6 +41,7 @@ import securesystemslib.formats
 # Import securesystemslib logger to log warning messages.
 logger = logging.getLogger('securesystemslib.hash')
 
+DEFAULT_CHUNK_SIZE = 4096
 DEFAULT_HASH_ALGORITHM = 'sha256'
 DEFAULT_HASH_LIBRARY = 'hashlib'
 SUPPORTED_LIBRARIES = ['hashlib']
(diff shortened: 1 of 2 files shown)


trishankatdatadog:
Looks great, thanks, @ofek! Does this address newlines on boundaries, do you know?

ofek:
Yes

trishankatdatadog:
Dumb question: what if the nth chunk ends with `\r` and the (n+1)th chunk begins with `\n` and we're on Windows?

ofek:
Yes and that case is tested (full coverage)

awwad:
Mechanically:
On a quick glance, it looks right mechanically. Also, there are no resulting TUF or in-toto test failures I've seen (locally, 2.7 and 3.6 TUF, 2.7 in-toto). I doubt the tests encounter these cases (not sure if we should add some), but it means we didn't break anything, so that's good.

Conceptually, I'm a little worried, since this means that the hashes of these two things hash equivalently. If someone knows this, they can swap in one for the other.... (LMK if my thinking is off.)

turtle
turtle-equivalent

This might be too low-level a place to put this code?

SantiagoTorres:
So, the motivation for this change follows the git usecase, in which windows uses \r\n line-endings to files, whereas *nix uses \n. This would happen, which is troublesome, but it's up to the applications to make sure the replacement only happens on *text* files, where `\n` and `\r\n` have a distinctive usecase.

awwad:
I'll think about it a little more and look at the stacks this lives in this afternoon, but I think that normalizing data is something that should happen at a higher level, or, at the least, be non-default and well-documented.

trishankatdatadog:
Our thinking is that in-toto (and, by extension, securesystemslib) should handle this, but we also do not think that this should happen automagically for everyone, which is why we are advocating for the flag approach. We do not think that in-toto integrators should be explicitly aware, by default, of "interesting" edge cases such as this.

@ofek Please let me know if I misunderstood anything.

awwad:
My bad: somehow managed to read that the default was to normalize. It's not, so I'm OK with this. :) I'll take one more look and I should be able to merge it shortly this afternoon.

PR #160, opened 2018-11-20
WIP: sphincs+ support, for post-quantum crypto

awwad:
This PR is to create a place to discuss the changes that @cryptojedi and @joostrijneveld have recommended for adding sphincs+ post-quantum crypto support to secure-systems-lib as a means to allow TUF and in-toto to make use of it.


awwad:
I think that it's important that sphincs support and the use of pyspx (which is a bit new and could use more testing) be optional (not a requirement for ssl installation).  What do you think, @SantiagoTorres?

securesystemslib/interface.py
```diff
@@ -900,6 +900,260 @@ def import_ecdsa_privatekey_from_file(filepath, password=None):
   return key_object
 
 
+def generate_and_write_spx_keypair(filepath=None, password=None):
```
awwad: Sidenote (doesn't have to be done for this PR): much of the code in this function is duplicated in multiple places: generate_and_write_..._keypair.  This makes it clear that we should modularize the generate functions and do things like argument checking, password prompting and format validation, temp file creation, writing, etc. in one place used by all the generate_ functions.

securesystemslib/interface.py
```diff
@@ -900,6 +900,260 @@ def import_ecdsa_privatekey_from_file(filepath, password=None):
...
+  # encrypted.
+  file_object.write(spx_key.encode('utf-8'))
+  file_object.move(filepath)
+
+  return filepath
+
+
+
+
+def import_spx_publickey_from_file(filepath):
```
awwad: Not necessary for this PR: this function can be generalized across all keytypes.

SantiagoTorres:
ACK,

How time-sensitive is this review? :)

Thanks!
-Santiago.

awwad:
Just wanted to know if you have thoughts on optional-vs-required.  No rush on that question.

awwad:
@cryptojedi and @joostrijneveld: This looks good.  I'll make more comments later, but I very much appreciate the effort to make this fit the style of the code and include good docstrings.

If you don't mind, I think I have to make a few edits to make the use of the sphincs+ implementation and pyspx optional before I can merge this.  That'll entail a few edits after Thanksgiving (back on Monday).  The thinking there is that pyspx is still pretty fresh and not heavily tested, and I don't think I should require folks using TUF to include it -- just allow them to use it optionally.

joostrijneveld:
@awwad Sounds great, thanks! We didn’t really give making the dependency optional much thought (other than checking for potentially unsuccessful imports), but that definitely sounds like a good idea.

I completely agree w.r.t the status of pyspx; I’ll make sure to add some more comprehensive tests.

securesystemslib/formats.py
```diff
@@ -227,7 +235,7 @@
 # Supported TUF key types.
 KEYTYPE_SCHEMA = SCHEMA.OneOf(
   [SCHEMA.String('rsa'), SCHEMA.String('ed25519'),
-   SCHEMA.String('ecdsa-sha2-nistp256')])
+   SCHEMA.String('ecdsa-sha2-nistp256'),SCHEMA.String('spx')])
```
alyptik: Shouldn't there be a space after the "," to be consistent with the previous line?

securesystemslib/interface.py
```diff
@@ -900,6 +900,260 @@ def import_ecdsa_privatekey_from_file(filepath, password=None):
...
+  securesystemslib.formats.PATH_SCHEMA.check_match(filepath)
+
+  # If the caller does not provide a password argument, prompt for one.
+  if password is None: # pragma: no cover
+
+    # It is safe to specify the full path of 'filepath' in the prompt and not
+    # worry about leaking sensitive information about the key's location.
+    # However, care should be taken when including the full path in exceptions
+    # and log files.
+    password = get_password('Enter a password for the SPX'
```
alyptik: Multi-line strings like this are kind of confusing to mentally parse when combined with string concatenation. Is there maybe a more compact way to write these?

securesystemslib/interface.py
```diff
@@ -900,6 +900,260 @@ def import_ecdsa_privatekey_from_file(filepath, password=None):
...
+    # It is safe to specify the full path of 'filepath' in the prompt and not
+    # worry about leaking sensitive information about the key's location.
+    # However, care should be taken when including the full path in exceptions
+    # and log files.
+    # NOTE: A user who gets prompted for a password, can only signal that the
+    # key is not encrypted by entering no password in the prompt, as opposed
+    # to a programmer who can call the function with or without a 'password'.
+    # Hence, we treat an empty password here, as if no 'password' was passed.
+    password = get_password('Enter a password for an encrypted RSA'
+        ' file \'' + Fore.RED + filepath + Fore.RESET + '\': ',
```
alyptik: What's the reason you chose single quotes for these literals instead of using double quotes such as in `" file '" + ...` vs. `'file \''`?

securesystemslib/keys.py
```diff
@@ -1637,6 +1726,46 @@ def import_ed25519key_from_private_json(json_str, password=None):
...
+  else:
+    logger.debug('No password was given. Attempting to import an'
+        ' unencrypted file.')
+    try:
+      key_object = \
+               securesystemslib.util.load_json_string(json_str.decode('utf-8'))
+    # If the JSON could not be decoded, it is very likely, but not necessarily,
+    # due to a non-empty password.
+    except securesystemslib.exceptions.Error:
+      raise securesystemslib.exceptions\
```
alyptik: How come you broke the line before the "." instead of after like in previous statements?

securesystemslib/spx_keys.py
```diff
@@ -0,0 +1,307 @@
...
+  securesystemslib.formats.SPX_SIG_SCHEMA.check_match(scheme)
+
+  # Signing the 'data' object requires a seed and public key.
+  # spx.signing.SigningKey.sign() generates the signature.
+  public = public_key
+  private = private_key
+
+  signature = None
+
+  # An if-clause is not strictly needed here, since 'spx' is the only
```
alyptik: Since we are bothering to handle the error with an `else:` clause, wouldn't you agree that the `if` check actually does matter? Maybe the comment could be reworded to reflect this if so.

JustinCappos:
Looks like the remaining issues are very minor code changes.  @lukpueh would you kindly take a look?  If it's very minor edits, please just make them and we can merge...

lukpueh:
Continued in #169. Closing here.

PR #164, opened 2019-03-30
Fix hash and hex schema regex pattern

adityasaky:
The regex patterns used to match `HASH_SCHEMA` and `HEX_SCHEMA` are incomplete - they're not strict about matching only what's in the character class. This PR adds a fix for that.

[This](
https://github.com/adityasaky/in-toto/commit/aa579917a029f7f1253404d96dbf063549e7bb46) is a commit on a branch which loads an in-toto metadata file which has key IDs with invalid characters. However, the validation which uses `HASH_SCHEMA` (indirectly) passes.

### DIFF
--- a/securesystemslib/formats.py
+++ b/securesystemslib/formats.py
@@ -99,7 +99,7 @@ ISO8601_DATETIME_SCHEMA = SCHEMA.RegularExpression(r'\d{4}-\d{2}-\d{2}T\d{2}:\d{
 UNIX_TIMESTAMP_SCHEMA = SCHEMA.Integer(lo=0, hi=2147483647)
 
 # A hexadecimal value in '23432df87ab..' format.
-HASH_SCHEMA = SCHEMA.RegularExpression(r'[a-fA-F0-9]+')
+HASH_SCHEMA = SCHEMA.RegularExpression(r'^[a-fA-F0-9]+$')
 
 # A dict in {'sha256': '23432df87ab..', 'sha512': '34324abc34df..', ...} format.
 HASHDICT_SCHEMA = SCHEMA.DictOf(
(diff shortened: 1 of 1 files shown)


lukpueh:
Good thinking, @adityasaky, but I fear you jumped to the wrong conclusion. 

The patterns are actually fine without leading `^` (see [Python's Regular Expression Syntax](https://docs.python.org/3/library/re.html#regular-expression-syntax) for details). 

As for the terminal `$`, you are indeed right, it is required in order to not match strings that have unspecified trailing characters. But take a look at [`securesystemslib.schema.RegularExpression`](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/schema.py#L945-L948) and you'll see that every pattern is auto-suffixed with a `$`.

Could it be that the test in your  `poc-ssl-keyid-regex-broken`-branch passes (although you expect it to fail) because you don't run it? Don't forget that our [TestLoader requires test methods to start with `test`](https://docs.python.org/3/library/unittest.html#unittest.TestLoader.testMethodPrefix).

Some more advice:
- You can run test methods individually, e.g.:
```
python tests/test_verifylib.py TestKeyIdVerificationSSL.test_verify_keyid_validation_ssl
```

- If you expect a test case to raise an exception, then assert for it, e.g.:

```python
  def test_verify_keyid_validation_ssl(self):
    invalid_layout_path = os.path.join(
        os.path.dirname(os.path.realpath(__file__)), "demo_files",
        "demo-invalid.layout.template")

    with self.assertRaises(securesystemslib.exceptions.FormatError):
      metadata = Metablock.load(invalid_layout_path)
```

adityasaky:
Indeed I had. Thank you for the detailed response and the clarification. I didn't look at the constructor at all, which in retrospect I should have. Also a stupid mistake not beginning the method with test and only the class haha.

Thanks again!

Issue #35, opened 2017-01-27
Create public high-level interface for securesystemslib.keys


lukpueh:
~~Currently `securesystemslib.keys` provides PEM import functions for public and private ecdsa and rsa keys. 
Let's also support ed25519 and add JSON import functions for all three protocols.~~

UPDATE:
Below discussion has moved the focus of this issue to add/factor out securesystemslib high-level functions from the lower-level `securesystemslib.keys` module, i.e. what goes into `securesystemslib.keys` and what goes into `securesystemslib.interface`.

vladimir-v-diaz:
I'm confused about exactly what you want.  Let's use the import function for RSA keys as an example...
import_rsakey_from_private_pem(pem, password=None)

The above function accepts a PEM argument and returns JSON:

https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/keys.py#L1061-L1109

Do you want to accept JSON argument and return what?

lukpueh:
First of all I want to add three functions:
- `import_ed25519_from_pem`
- `import_ed25519_from_private_pem`
- `import_ed25519_from_public_pem`
which should return python dictionaries compliant to `formats.ANYKEY_SCHEMA`, like `import_rsakey_from_private_pem` does (unless we say, we never store ed25519 keys in PEM format).

Additionally I want to add or rather move functions to import ed25519, ecdsa and RSA key, JSON serialized in above format and returning Python dictionaries in above format, like [tuf.import_ed25519_publickey_from_file] (https://github.com/lukpueh/tuf/blob/develop/tuf/repository_lib.py#L1120) or [`securesystemslib.import_ed25519_publickey_from_file`](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/interface.py#L402) (which are copy-pated btw.)

The rationale is, to provide all key loading functions in `securesystemslib.keys`, which IMHO was why we separated `securesystemslib` from TUF in the first place.

lukpueh:
I think we should add all functions from `securesystemslib.interface` and relevant functions from `tuf.repository_lib` to `securesystemslib.keys`, which according to its docstring is an " easy-to-use public interface" for "(RSA, ECDSA and Ed25519)-related function".

vladimir-v-diaz:
[...]

PR #170, opened 2019-06-19
External Signing using CCID/PIV interface.

tanishqjasoria:
This aims to add support for hardware security modules, to generate signatures using keys stored in the modules.  

**Fixes issue #**: [tuf/issues/864](https://github.com/theupdateframework/tuf/issues/864)

1. Add PKCS11LIB to securesystemslib/settings.py. It servers as the PATH of PKCS11 library for 
    cryptographic tokens
2. Add hsm.py, to support hardware security modules through the PKCS#11 standard. This module use 
    PyKCS11, a python wrapper (SWIG) for PKCS#11 modules to communicate with the cryptographic 
    tokens using CCID/PIV interface
3. Add hsm_keys.py, to provide a high-level API for using hardware security modules for various 
    cryptographic operations

requirements**:


JustinCappos:
Glad to see this moving along!

securesystemslib/hsm.py
```diff
@@ -0,0 +1,448 @@
...
+    if PyKCS11.CKK[key_type] == 'CKK_RSA':
+
+      # RSA-PSS with SHA256 hash to be used for signature generation
+      RSA_PSS = PyKCS11.CKM_SHA256_RSA_PKCS_PSS
+      # SHA256 hash to be used to digest the data
+      HASH = PyKCS11.CKM_SHA256
+      # Mask generating function
+      MGF = PyKCS11.CKG_MGF1_SHA256
+      # length of salt to be used for hashing
+      salt_length = 32
```
mnm678: You define this same value a few times. I would move the variable to the top of the file and make it all caps.

securesystemslib/hsm_keys.py
```diff
@@ -0,0 +1,213 @@
...
+    library is not specified in 'settings.py' or the library is corrupt
+
+  <Returns>
+    list of dictionaries corresponding to all the available HSMs
+  """
+
+  # All the functions must use the same object of the HSM class,
+  # to use same session for all the operations.
+  global smartcard
+  smartcard = HSM(PKCS11LIB)
```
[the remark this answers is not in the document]
tanishqjasoria: Sorry, I forgot to remove that!
I would modify it.

securesystemslib/settings.py
```diff
@@ -46,3 +46,6 @@
 
 # The algorithm(s) in HASH_ALGORITHMS are used to generate key IDs.
 HASH_ALGORITHMS = ['sha256', 'sha512']
+
+# The PATH of PKCS11 library for hardware security modules
+PKCS11LIB = "/usr/lib/opensc-pkcs11.so"
```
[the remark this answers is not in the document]
tanishqjasoria: The other way is for the user to set a environment variable before using the module.

securesystemslib/hsm.py
tanishqjasoria: If we plan to add other RSA signing mechanism in future, then it might be a problem. If not, I can move the variables to the top.

securesystemslib/hsm_keys.py
```diff
@@ -0,0 +1,212 @@
...
+    securesystemslib.exceptions.NotFoundError, if the path of PKCS#11
+    library is not specified in 'settings.py' or the library is corrupt
+
+  <Returns>
+    list of dictionaries corresponding to all the available HSMs
+  """
+
+  # All the functions must use the same object of the HSM class,
+  # to use same session for all the operations.
+  global smartcard
```
lukpueh: No need to use `global` here.

securesystemslib/hsm_keys.py
```diff
@@ -0,0 +1,212 @@
...
+
+  <Returns>
+    list of handle of all the available private keys
+  """
+
+  # Use the HSM with the corresponding 'slot_info'
+  smartcard.get_HSM_session(HSM_info['slot_id'])
+
+  _user_pin = securesystemslib.interface.get_password(
+    prompt="Enter the User Pin for the SmartCard")
```
lukpueh: I think `*_keys.py` modules are meant to be non-interactive. A function that requires a user to input a password or pin is a better fit for `securesystemslib.interface`.

securesystemslib/hsm_keys.py
```diff
@@ -0,0 +1,212 @@
...
+
+  <Arguments>
+    HSM_info:
+      element from the list generated by 'load_HSMs()'
+
+  <Exceptions>
+    securesystemlib.exceptions.InvaliNameError, if the requested
+    HSM is either not present or cannot be used.
+
+    securesystemslib.exceptions.BasPasswordError, if the entered
```
lukpueh: s/BasPasswordError/BadPasswordError/

securesystemslib/hsm.py
```diff
@@ -0,0 +1,446 @@
...
+    """
+    <Purpose>
+      User Login into the HSM. Required to access private objects.
+
+    <Arguments>
+      user_pin:
+        PIN for the CKU_USER login.
+
+    <Exceptions>
+      securesystemslib.exceptions.BasPasswordError, if the entered
```
lukpueh: s/BasPasswordError/BadPasswordError/

securesystemslib/hsm_keys.py
```diff
@@ -0,0 +1,212 @@
...
+  <Purpose>
+    Get list of handles of private keys stored in the HSM
+    corresponding to the 'HSM_info'
+
+  <Arguments>
+    HSM_info:
+      element from the list generated by 'load_HSMs()'
+
+  <Exceptions>
+    securesystemlib.exceptions.InvaliNameError, if the requested
```
lukpueh: s/InvaliNameError/InvalidNameError/

securesystemslib/hsm_keys.py
```diff
@@ -0,0 +1,212 @@
...
+  <Purpose>
+    Get list of handles of public keys stored in the HSM
+    corresponding to the 'HSM_info'
+
+  <Arguments>
+    HSM_info:
+      element from the list generated by 'load_HSMs()'
+
+  <Exceptions>
+    securesystemlib.exceptions.InvaliNameError, if the requested
```
lukpueh: s/InvaliNameError/InvalidNameError/

securesystemslib/hsm.py
```diff
@@ -0,0 +1,446 @@
...
+  hsm.py
+
+<Author>
+  Tanishq Jasoria <jasoriatanishq@gmail.com>
+
+<Purpose>
+  The goal of this module is to support hardware security modules through
+  the PKCS#11 standard.
+
+  This module use PyKCS11, a python wrapper (SWIG) for PKCS#11 modules
```
lukpueh: s/use/uses/

securesystemslib/hsm_keys.py
```diff
@@ -0,0 +1,212 @@
...
+"""
+<Program Name>
+  hsm_keys.py
+
+<Author>
+  Tanishq Jasoria <jasoriatanishq@gmail.com>
+
+<Purpose>
+  This module provides a high-level API for using hardware security modules
+  for various cryptographic operations
```
lukpueh: What's the rationale for having both `hsm_keys` and `hsm` modules? The high-level interface that may be called by a developer, who uses `securesystemslib`, should be in the `keys` module.

Please also make sure that the high-level interface provide data such es created signatures or exported public keys in an expected format (see `securesystemslib.formats.{*KEY_SCHEMA, *SIGNATURE_SCHEMA}`).

tests/test_hsm.py
```diff
@@ -0,0 +1,369 @@
...
+
+  def test_initialization(self):
+
+    self.assertRaises(securesystemslib.exceptions.NotFoundError,
+        self.HSM, None)
+
+    try:
+      self.HSM('/NO LIBRARY HERE')
+    except:
+      pass
```
lukpueh: Why not use `assertRaises` here as well?

securesystemslib/hsm.py
tanishqjasoria: I made the required changes!

dev-requirements.txt
```diff
@@ -5,6 +5,8 @@ tox==3.2.1
 coveralls==1.3.0
 coverage==4.5.1
 colorama==0.3.9
+PyKCS11==1.5.5
+PyKCS11==1.5.5
```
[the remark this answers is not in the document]
tanishqjasoria: Yes I had a look at python-pkcs11, but OpenSC recommends using PyKCS11 to add supoort for PKCS11.
And also, I was having a hard time to decide which library to use, there are many option available out there, so I just started with what I felt was good.

tests/.coveragerc
```diff
@@ -12,3 +12,5 @@ exclude_lines =
 
 omit =
   */securesystemslib/_vendor/*
+  */securesystemslib/hsm.py
+  */securesystemslib/hsm_keys.py
```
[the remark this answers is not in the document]
tanishqjasoria: I did that for the time being as I was not able to figure out how to carry out the required tests without having the HSM attached to machine.
I would be adding tests for both the modules.

securesystemslib/hsm.py
```diff
@@ -0,0 +1,446 @@
...
+    """
+    <Purpose>
+      To logout and terminate sessions with the HSM completely.
+    """
+
+    try:
+      self.logout()
+      self.close_session()
+    except:
+      # Exception is raised when user already logged out
```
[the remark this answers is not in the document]
tanishqjasoria: I made the required changes to make it more readable!

securesystemslib/hsm_keys.py
tanishqjasoria: When I started I was not sure which library to use to interact with the hardware tokens. So, I thought of having two modules, one which implement basic function to interact with the tokens, while hiding the library specific details and the other module will use these function to incorporate required functionality in securesystemslib. 
Now if we decide not to use PyKCS11, we can simply modify the hsm.py module without making major changes in the hsm_keys.py module.

lukpueh: Fair enough. Please feel free to chime in on the discussion about `securesystemslib`'s API in #35. IMO the different layers of abstraction should be more intuitive or more clearly defined (or both). But that's not something we need to fix here.

lukpueh:
Thanks for the changes, @tanishqjasoria! Please ping me when you've fixed the tests and addressed all of my comments, and I'll re-review. And let me know if you need any help.

lukpueh:
Hi @tanishqjasoria, are you planning on finalizing this PR? It would be very much appreciated. :)

tanishqjasoria:
Yeah sure, I would try to finalize this PR by this weekend!

lukpueh:
It would be good to rebase on top of master, because the recent changes enhance automated testing and coverage visualization. 

I do suggest though to first address the inline comments in my last review, as GitHub understandably has a hard time mapping the comments to the code after a rebase. :)

tests/test_hsm.py
```diff
@@ -34,7 +37,10 @@
 # To initialize SoftHSM for testing purposes!
 import PyKCS11
 # Library to interact with SoftHSM.
-PKCS11LIB = './tests/lib/libsofthsm2.so'
+PKCS11LIB = '/usr/local/lib/softhsm/libsofthsm2.so'
+
+# Path where SoftHSM is created and stored
+TOKENS_PATH = '/var/lib/softhsm/tokens'
```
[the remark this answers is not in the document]
tanishqjasoria: The token path is taken from a .configure file. If we have to use a custom directory for the token, we need to provide a .configure file and before execution we would have to set a env variable to point to the directory of new config file!

tests/test_hsm_keys.py
```diff
@@ -0,0 +1,276 @@
...
+from cryptography.hazmat.primitives.asymmetric.ec import EllipticCurvePublicKey
+from cryptography.x509 import Certificate
+
+# To initialize SoftHSM for testing purposes!
+import PyKCS11
+# Library to interact with SoftHSM.
+PKCS11LIB = '/usr/local/lib/softhsm/libsofthsm2.so'
+
+# Path where SoftHSM is created and stored
+TOKENS_PATH = '/var/lib/softhsm/tokens'
```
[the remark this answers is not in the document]
tanishqjasoria: I can put these variables in the __init__.py file?

.travis.yml
```diff
@@ -9,8 +9,18 @@ env:
...
+  - sudo apt-get install -y swig
+
 before_script:
   - pip install -U pip wheel tox
+  - sudo wget https://dist.opendnssec.org/source/softhsm-2.3.0.tar.gz
+  - sudo tar -xzf softhsm-2.3.0.tar.gz
+  - cd softhsm-2.3.0
+  - sudo ./configure
+  - sudo make
+  - sudo make install
```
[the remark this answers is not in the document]
tanishqjasoria: Yes, sudo is needed for all the commands!
Sure, I would keep that in mind while resolving the conflict!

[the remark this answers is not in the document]
SantiagoTorres: This is most likely *not* the case.  You need sudo because you wget'ed with sudo and then untar'd with sudo, so the resulting files are owned by root.  The only command that should need sudo here is probably the make install

securesystemslib/hsm.py
```diff
@@ -431,14 +431,16 @@ def close(self):
...
     except PyKCS11.PyKCS11Error as error:
-      if error.__str__() == 'CKR_USER_NOT_LOGGED_IN (0x00000101)':
-        # When the user is already logged out.
-        logger.info(str(error))
-        self.close_session()
-      else:
-        # When the session does not exists.
-        logger.info(str(error))
+      # Error is raised when there is no active session with the HSM.
+      logger.info(str(error))
```
SantiagoTorres: I think this should be a higher level of logging than `info`. Could you please change?

.travis.yml
```diff
@@ -15,8 +15,8 @@ install:
   - sudo apt-get update
   - sudo apt-get install -y swig
   - pip install -U tox coveralls
-  - sudo wget https://dist.opendnssec.org/source/softhsm-2.3.0.tar.gz
-  - sudo tar -xzf softhsm-2.3.0.tar.gz
+  - wget https://dist.opendnssec.org/source/softhsm-2.3.0.tar.gz
+  - tar -xzf softhsm-2.3.0.tar.gz
   - cd softhsm-2.3.0
   - sudo ./configure
```
SantiagoTorres: I don't think now you  need sudo for configure and make,  only for install. You used to need sudo for these because the tar was sudo'd.

securesystemslib/hsm.py
tanishqjasoria: Sure, I will change that!

.travis.yml
tanishqjasoria: I tried it without sudo (when I didnt use sudo for wget and tar), but it didn't work. 
There were some erros, regarding Permission Denied!

SantiagoTorres: I'm pretty sure you don't need it:
```
FROM debian:stable

RUN apt-get update && apt-get install -y swig python3-pip wget libssl-dev
RUN pip3 install -U tox coveralls

# drop privs
RUN useradd mortal && mkdir /mundaneopolis && chown mortal:mortal /mundaneopolis
USER mortal:mortal
WORKDIR /mundaneopolis

RUN wget https://dist.opendnssec.org/source/softhsm-2.3.0.tar.gz && tar -zxvf softhsm-2.3.0.tar.gz
WORKDIR /mundaneopolis/softhsm-2.3.0/
RUN ./configure --disable-gost && make

# RUN make install # will fail
USER root:root
RUN make install
```

tanishqjasoria: Yup, It does not need sudo. I did a clean install, without sudo and it worked.
Apparently, there is a .log file which is modified by the .configure script. Due to my previous installation, root was the owner of the file.

trishankatdatadog:
Finally --- good job, guys!!

setup.py
```diff
@@ -96,7 +96,7 @@
     'Topic :: Software Development'
   ],
   install_requires = ['six>=1.11.0', 'subprocess32; python_version < "3"',
-      'python-dateutil==2.8.0', 'colorama>=0.3.9'],
+      'python-dateutil==2.8.0', 'PyKCS11==1.5.5; python_version > "3"', 'colorama>=0.3.9'],
```
Silvanoc: You're pinning here a version of PyKCS11 that is not available anymore on PyPI.

trishankatdatadog: Sorry, maybe I'm missing something, but in general we should be using `>=` not `==`, no?

tanishqjasoria: Sorry, I forgot to change that.

securesystemslib/hsm.py
```diff
@@ -0,0 +1,446 @@
...
+    mechanism = None
+    key_type = self.session.getAttributeValue(private_key_handle,
+        [PyKCS11.CKA_KEY_TYPE])[0]
+
+    if PyKCS11.CKK[key_type] == 'CKK_RSA':
+      mechanism = PyKCS11.RSA_PSS_Mechanism(RSA_PSS_MECH,
+          RSA_PSS_HASH_SHA256, RSA_PSS_MGF_SHA256, RSA_PSS_SALT_LENGTH)
+
+    elif PyKCS11.CKK[key_type] == 'CKK_EC' or PyKCS11.CKK[key_type] == 'CKK_ECDSA':
+      mechanism = PyKCS11.Mechanism(PyKCS11.CKM_ECDSA)
```
Silvanoc: The mechanism to be used should be negotiated with the slot (provided by [PyKCS11Lib.getMechanismList](https://pkcs11wrap.sourceforge.io/api/api.html#PyKCS11.PyKCS11Lib.getMechanismList)) to ensure that a supported one gets used. Mine, for example, doesn't support any of the PSS mechanisms.

I'd rather provide a sorted list (being the first the preferred mechanism) and automatically select the first matching one.

BTW this is exactly the same code as in lines 384-389, right? I'd rather extract that code to a function, not only for code economy, but in order to ensure that the same negotiation algorithm gets used.

setup.py
Silvanoc: BTW, you have the same version pinning in [ci-requirements.txt](https://github.com/secure-systems-lab/securesystemslib/pull/170/files?file-filters%5B%5D=.py&file-filters%5B%5D=.txt#diff-e513740081fff179cd93aa695e63334dR5) and [test-requirements.txt](https://github.com/secure-systems-lab/securesystemslib/pull/170/files?file-filters%5B%5D=.py&file-filters%5B%5D=.txt#diff-913fedaed3cb05fe362c0ceb35e2d827R8).

lukpueh: Thanks for pointing this out. The various *-requirements.txt are due for revision (see #182).

securesystemslib/hsm.py
tanishqjasoria: To get started with the work, We kept it simple at first (worked on the issue according to the mechanisms available with the YubiKey). I can make the required changes to get it supported with various HSM available. 

How would we decide the list of mechanisms to be made available?

securesystemslib/hsm.py
```diff
@@ -0,0 +1,446 @@
...
+from cryptography import x509
+
+# RSA-PSS with SHA256 hash to be used for signature generation.
+RSA_PSS_MECH = PyKCS11.CKM_SHA256_RSA_PKCS_PSS
+# SHA256 hash to be used to digest the data.
+RSA_PSS_HASH_SHA256 = PyKCS11.CKM_SHA256
+# Mask generating function for SHA256 Hash.
+RSA_PSS_MGF_SHA256 = PyKCS11.CKG_MGF1_SHA256
+# Length of salt to be used for hashing.
+RSA_PSS_SALT_LENGTH = 32
```
tanishqjasoria: Due to various mechanisms available, I just declared the mechanism at the start for consistency within various operations.

lukpueh:
Hi @tanishqjasoria, you have a couple of unnecessary merge commits in your branch. The preferred way to sync a feature branch with upstream master is to regularly rebase on master (instead of merging master into the feature branch). 

I  cleaned up your commit history by cherry-picking your relevant commits on top of the up-to-date upstream master (resolving many conflicts). You can replace your branch with my cleaned up version using these commands (try to understand what they are doing):
```bash
# Fetch my version of your branch
git fetch git@github.com:lukpueh/securesystemslib.git hsm_support_signing-clean:hsm_support_signing-clean

# Go to your branch (if you're not already on it)
git checkout hsm_support_signing

# You can diff yours with mine to see that mine does not remove any of your changes
# You should only see additions that come from syncing with upstream master
git diff hsm_support_signing hsm_support_signing-clean

# Replace yours with mine
git reset --hard hsm_support_signing-clean

# Force push to update the branch on GitHub
# (you might have to change "origin" to whatever points to git@github.com:tanishqjasoria/securesystemslib.git)
git push origin hsm_support_signing -f
```


In case you are interested how I cleaned up your history, here's what I did. You don't need to do this, if you used above commands, but it helps to know these commands.

```bash

# Note down the relevant commit hashes from your PR
#  `--no-merges` shows the log without merge commits
# `--first-parent` shows the log without commits that you merged in
git log --oneline --no-merges  --first-parent master..hsm_support_signing

# Create a new branch on top of master
# (make sure that master is the laster version of upstream/master !!) 
git checkout -b hsm_support_signing-clean &&  git reset --hard master

# Now use `git cherry-pick <commit ID>` for all commits that you noted above, 
# starting at the bottom and going to the top.
# This is actually the hard part, because it requires fixing quite a few conflicts, due
# to upstream master having moved on, which in turn requires knowledge of the
# changes in upstream master.

```

For future references, please try to avoid merge commits on your feature branch! :)

tanishqjasoria:
Thanks a lot for your help @lukpueh!
I will keep this in mind from now on.

joshuagl:
In Issue #179 we're working to ensure that any functionality with a native dependency presents meaningful user feedback when the native dependency is not available.

It would be great if this PR could implement functionality similar to #200 to ensure the `hsm_keys` module can be imported even if the native dependency isn't available and that any functions which rely on that native dependency throw an `UnsupportedLibraryError`.

trishankatdatadog:
Also, how do we test that this works? Docker has tests that somehow tests Yubikeys, not sure how they do it, maybe using emulators.

tanishqjasoria:
> Also, how do we test that this works? Docker has tests that somehow tests Yubikeys, not sure how they do it, maybe using emulators.

I am testing it using softhsm to emulate security tokens!
https://www.opendnssec.org/softhsm/

trishankatdatadog:
> I am testing it using softhsm to emulate security tokens!

Very nice! Sorry, I haven't closely, but is it part of the test suite?

lukpueh:
@SantiagoTorres just pointed out that there's a [new OpenSC release (0.20.0)](https://github.com/OpenSC/OpenSC/releases/tag/0.20.0), which fixes a bunch of CVEs. Let's make sure to use that.

tanishqjasoria:
@joshuagl I would try to implement the same functionality here also. @trishankatdatadog Yes, softhsm is already a part of test suit for hsm.py and hsm_keys.py.

securesystemslib/settings.py
```diff
@@ -50,3 +50,6 @@
 # Used in securesystemslib.process, to raise a subprocess.TimeoutExpired if
 # a started subprocess does not terminate before the here specified seconds
 SUBPROCESS_TIMEOUT = 3
+
+# The PATH of PKCS11 library for hardware security modules
+PKCS11LIB = "/usr/lib/opensc-pkcs11.so"
```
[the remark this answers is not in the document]
tanishqjasoria: The idea here was to let the user decide which library he want so use for the purpose. There are various vendor specific libraries available for the same.

tox.ini
```diff
@@ -16,7 +16,7 @@ deps =
 
 commands =
     coverage run tests/aggregate_tests.py
-    coverage report -m --fail-under 99
+    coverage report -m --fail-under 89
```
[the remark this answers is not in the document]
tanishqjasoria: PyKCS11 (the library used to interact with the HSM) is not supported on Python 2. The related tests are skipped in Python 2, which results in low coverage. This was just done to check if the tests were running successfully.

I discussed this with @lukpueh and we agreed to tackle it by using different coverage rules for different environments, as discussed [here](https://github.com/secure-systems-lab/securesystemslib/pull/200#discussion_r368568724).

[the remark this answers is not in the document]
tanishqjasoria: This won't be merged with low threshold for all of securesystemslib.

securesystemslib/hsm_keys.py
```diff
@@ -0,0 +1,216 @@
...
+  """
+  <Purpose>
+    To load a custom library to interact with the hardware tokens.
+
+  <Exceptions>
+    securesystemslib.exceptions.NotFoundError, if the path of PKCS#11
+    library is not specified or the library is corrupt.
+  """
+
+  global smartcard
```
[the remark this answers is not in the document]
tanishqjasoria: I was under the impression that using `global` would define it as a module level constant.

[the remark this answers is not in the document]
joshuagl: You're right, I had misremembered the rules around `global`. Apologies for that.

securesystemslib/settings.py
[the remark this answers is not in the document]
joshuagl: That makes sense, thanks for clarifying. Re-reading over the code do we even need this variable in `settings`? The only place it appears to be used is in test_hsm_keys.py where the variable is assigned a new value.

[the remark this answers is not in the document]
tanishqjasoria: It's meant to be used by the user.
If this is defined in `settings`, the user would just have to make changes here to update/change the library for all the instantiations.

.travis.yml
```diff
@@ -19,8 +19,18 @@ matrix:
     - python: "3.8"
       env: TOXENV=purepy38
 
+before_install:
+  - sudo apt-get update
+  - sudo apt-get install -y swig
+
 install:
   - pip install -U tox coveralls
+  - wget https://dist.opendnssec.org/source/softhsm-2.3.0.tar.gz
```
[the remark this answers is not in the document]
tanishqjasoria: Yes, we can use the version from the host OS. I will update this.

trishankatdatadog:
We should get @ woodruffw to review this too, since he plans to use a HSM to perform the key ceremony for PyPI

trishankatdatadog:
We should get @woodruffw to review this, too, since he plans to use an HSM to perform the PyPI key ceremony

trishankatdatadog:
Has anyone tested whether the code here works with something like the [YubiHSM](https://www.yubico.com/product/yubihsm-2)? @JustinCappos 

If not, I might be able to arrange for testing with one, but no guarantees

JustinCappos:
> Has anyone tested whether the code here works with something like the [YubiHSM](https://www.yubico.com/product/yubihsm-2)? @JustinCappos
> 
> If not, I might be able to arrange for testing with one, but no guarantees

I don't recall.  I'll let @lukpueh and @SantiagoTorres chime in...

SantiagoTorres:
Unfortunately, I don't think we've had access to a yubihsm. I think we've been successfully able to test on a yubikey pro 4(?) and my nitrokey pro

trishankatdatadog:
> I don't recall. I'll let @lukpueh and @SantiagoTorres chime in...

@lukpueh @SantiagoTorres Do you guys need a YubiHSM to test this with?

securesystemslib/settings.py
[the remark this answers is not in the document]
joshuagl: Unless I'm misreading/misunderstanding the variable `PKCS11LIB` in settings isn't used anywhere other than the tests, test_hsm.py and test_hsm_keys.py. In the latter the variable is immediately assigned a new value.

hsm.py takes a `PKCS11LIB_path` in the HSM constructor and assigns it to the instance variable `self.PKCS11LIB` of the HSM object. That variable is used later in the `refresh()` function.

hsm_keys.py has a function, `load_library()` which takes an argument `PKCS11LIB_USER` and uses that as the argument to a call to `PKCS11.load()` in the `hsm_keys.refresh()` function.

`settings.PKCS11LIB` appears to be unused outside of the test code?

securesystemslib/hsm.py
```diff
@@ -0,0 +1,446 @@
...
+
+logger = logging.getLogger('securesystemslib_hsm')
```
woodruffw: Mostly a nitpick, but `getLogger(__name__)` is a common practice for initializing module-level loggers. `__name__` also makes it easier to enable or disable particular log-levels in client code that consumes this library.

securesystemslib/hsm.py
```diff
@@ -0,0 +1,446 @@
...
+      library is not specified or the library is corrupt
+    """
+
+    # Try to load the PKCS11 library
+    try:
+      # Load the PKCS#11 library and simultaneously update the list
+      # of available HSM.
+      self.PKCS11.load(self.PKCS11LIB)
+    except PyKCS11.PyKCS11Error as error:
+      logger.info('PKS11 Library not found or is corrupt!')
```
woodruffw: Nitpick: `logger.error` for a more semantic message?

securesystemslib/hsm.py
```diff
@@ -0,0 +1,446 @@
...
+      securesystemslib.exceptions.NotFoundError, if the path of PKCS#11
+      library is not specified or the library is corrupt
+
+    <Returns>
+      A list of dictionaries consisting of relevant information
+      regarding all the available tokens
+    """
+
+    # Refresh the list of available slots for HSM
+    self.refresh()
```
woodruffw: This is a more general design question: what is the intended lifetime behavior of `HSM` objects?

My intuition as a consumer of this library is that the list of available HSMs available for use in an `HSM` instance is fixed at initialization time, but calling `refresh` here makes it possible that the list has changed between object creation and use. If the intended behavior is that any operation that needs to access HSMs always operates on a refreshed list, it might make sense to turn `refresh` into a decorator and explicitly document that as the behavior.

securesystemslib/hsm.py
```diff
@@ -0,0 +1,446 @@
...
+    <Exceptions>
+      securesystemslib.exceptions.BadPasswordError, if the entered
+      user pin is invalid.
+    """
+
+    try:
+      self.session.login(user_pin)
+    except PyKCS11.PyKCS11Error as error:
+      if error.__str__() == 'CKR_USER_ALREADY_LOGGED_IN (0x00000100)':
+        logger.info('Already logged in as CKU_USER.')
```
woodruffw: Nitpick: `logger.warning` might be more semantic.

Also, checking `__str__` is likely to be fragile. It looks like `error.value` contains the numeric error code, which you can map to a symbolic value like `CKR_USER_ALREADY_LOGGED_IN` via `PyKCS11.CKR` dictionary.

securesystemslib/hsm.py
```diff
@@ -0,0 +1,446 @@
...
+    result = self.session.verify(public_key_handle, signed_bytes, signature_bytes,
+        mechanism)
+
+    return result
+
+
+
+
+
+  def logout(self):
```
woodruffw: Nit: Do `logout` and `close_session` form part of the public API, or are they just abstracted here for calling in `close`? If the latter, it might make sense to document them as private and prefix them with underscores (e.g. `_logout` and `_close_session`) to discourage misuse.

tests/test_hsm_keys.py
```diff
@@ -0,0 +1,307 @@
...
+  import securesystemslib.hsm_keys
+  HSM = securesystemslib.hsm_keys
+
+# Library to interact with SoftHSM.
+PKCS11LIB = '/usr/local/lib/softhsm/libsofthsm2.so'
+
+# Path where SoftHSM is created and stored
+TOKENS_PATH = '/var/lib/softhsm/tokens'
+
+logger = logging.getLogger('securesystemslib_test_hsm_keys')
```
woodruffw: Same nitpick as the other `getLogger` call.

.coveragerc
```diff
@@ -7,6 +7,7 @@ omit =
   */tests/*
   */site-packages/*
 
+
```
trishankatdatadog:
```suggestion
```

lukpueh:
> > I don't recall. I'll let @lukpueh and @SantiagoTorres chime in...
> 
> @lukpueh @SantiagoTorres Do you guys need a YubiHSM to test this with?

It would be nice if NYU could get one to test it. I only have a YubiKey 5 nano, but haven't tested it with this PR yet.

securesystemslib/settings.py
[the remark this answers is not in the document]
tanishqjasoria: Yes, `settings.PKCS11LIB` is unused outside of the test code. The idea was, that the user could just initialize the HSM using `load_library(settings.PKCS11LIB)`. I can remove this from `settings` if this is not necessary.

tanishqjasoria:
I have tested this PR using the Yubikey NEO.

securesystemslib/hsm.py
tanishqjasoria: The intended behaviour of the `HSM` object is something like, get the list of available HSM -> establish a session with the required HSM -> do the intended job or jobs -> close the session.
It's advisable to refresh the list before doing a series of operations as SlotID may change on removal and reinsertion of the HSM.
In `hsm_keys`, `load_HSM` calls `refresh` and get the list of available HSMs to the user. The user chooses from the list of slot available and then pass it as the argument `HSM_info` to various functions. Adding `refresh` as a decorator won't give the user the functionality to choose from the available HSMs.

securesystemslib/settings.py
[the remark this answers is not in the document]
joshuagl: I don't think the variable in `settings` provides much value here. As it is a value that will almost certainly have to be changed before use it is no more convenient than having a local variable to pass around. Let's drop it.

trishankatdatadog:
> It would be nice if NYU could get one to test it. I only have a YubiKey 5 nano, but haven't tested it with this PR yet.

Let me test this PR with a YubiHSM we have around here. NYU should get one of their own, it's not very expensive (~$650).

securesystemslib/hsm.py
```diff
@@ -23,29 +23,40 @@
...
+try:
+  from cryptography.hazmat.primitives import serialization
+  from cryptography.hazmat.backends import default_backend
+  from cryptography import x509
+except ImportError:
+  CRYPTO = False
+
+# Import python wrapper for PKCS#11 to communicate with the tokens
+HSM_SUPPORT = True
+NO_HSM_MSG = "HSM support required PyKCS11 library"
```
joshuagl:
```suggestion
NO_HSM_MSG = "HSM support requires PyKCS11 library"
```

securesystemslib/hsm.py
```diff
@@ -23,29 +23,40 @@
...
+  # SHA256 hash to be used to digest the data.
+  RSA_PSS_HASH_SHA256 = PyKCS11.CKM_SHA256
+  # Mask generating function for SHA256 Hash.
+  RSA_PSS_MGF_SHA256 = PyKCS11.CKG_MGF1_SHA256
+  # Length of salt to be used for hashing.
+  RSA_PSS_SALT_LENGTH = 32
+except ImportError:
+  HSM_SUPPORT = False
+
+logger = logging.getLogger('__securesystemslib_hsm__')
```
joshuagl:
```suggestion
logger = logging.getLogger(__name__)
```
There was a recent change in 5c1a1d953cb00b481d277e6f48c5003d1c9e9d0c to standardise on the pattern `logger = logging.getLogger(__name__)`

securesystemslib/hsm.py
```diff
@@ -236,9 +253,15 @@ def get_public_key_value(self, public_key_handle):
       public_key_handle:
         element of the list returned by get_public_key_objects().
 
+    <Exceptions>
+      securesystemslib.exceptions.UnsupportedLibraryError, if the cryptography
+      module id not available.
```
joshuagl:
```suggestion
      module is not available.
```

securesystemslib/hsm.py
```diff
@@ -279,9 +302,15 @@ def get_X509_value(self, x509_handle):
       x509_handle:
         element from the list returned by get_X509_objects().
 
+    <Exceptions>
+      securesystemslib.exceptions.UnsupportedLibraryError, if the cryptography
+      module id not available
```
joshuagl:
```suggestion
      module is not available
```

securesystemslib/hsm_keys.py
```diff
@@ -179,10 +185,17 @@ def export_public_key_PEM(public_key_handle):
     public_key_handle:
       element from the list generated by 'load_public_keys()'.
 
+  <Exceptions>
+      securesystemslib.exceptions.UnsupportedLibraryError, if the cryptography
+      module id not available
```
joshuagl:
```suggestion
      module is not available
```

securesystemslib/hsm_keys.py
```diff
@@ -0,0 +1,238 @@
...
+  """
+  <Purpose>
+    To get list of all the available HSMs
+
+  <Exceptions>
+    securesystemslib.exceptions.NotFoundError, if the path of PKCS#11
+    library is not specified in 'settings.py' or the library is corrupt
+
+    securesystemslib.exceptions.Error, if the function call is made
+    before the inilialization call using load_library.
```
lukpueh:
```suggestion
    before the initialization call using load_library.
```

securesystemslib/hsm_keys.py
```diff
@@ -0,0 +1,238 @@
...
+
+  <Returns>
+    list of dictionaries corresponding to all the available HSMs
+  """
+
+  # All the functions must use the same object of the HSM class,
+  # to use same session for all the operations.
+
+  # Get information reagarding the available HSM
```
lukpueh:
```suggestion
  # Get information regarding the available HSM
```

securesystemslib/hsm_keys.py
```diff
@@ -0,0 +1,238 @@
...
+  from cryptography.hazmat.backends import default_backend
+except ImportError:
+  CRYPTO = False
+
+
+def load_library(PKCS11LIB_USER):
+  """
+  <Purpose>
+    To load a custom library to interact with the hardware tokens.
+
```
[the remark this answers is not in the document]
lukpueh: I see the `HSM` class has more info.  Maybe you can transfer/copy some of it to the public facing function, and flesh it out a bit.

trishankatdatadog:
Note to self: test code with YubiHSM sooner than later...

securesystemslib/hsm_keys.py
```diff
@@ -0,0 +1,239 @@
...
+    securesystemslib.exceptions.BadPasswordError, if the entered
+    user pin is invalid.
+
+  <Returns>
+    list of handle of all the available private keys
+  """
+
+  # Use the HSM with the corresponding 'slot_info'
+  smartcard.get_HSM_session(HSM_info)
```
lukpueh: If you call this function before `load_library` you will get `NameError: name 'smartcard' is not defined`

lukpueh: Same comment applies to other functions that call `smartcard`

requirements-test.txt
```diff
@@ -1,5 +1,6 @@
 # test runtime dependencies  (see 'tests_require' field in setup.py)
 mock; python_version < "3.3"
+PyKCS11>=1.5.5; python_version > '3'
```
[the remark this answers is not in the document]
tanishqjasoria: Actually, I already did a rebase and resolved the conflicts.

securesystemslib/hsm.py
```diff
@@ -0,0 +1,473 @@
...
+    <Purpose>
+      Get object handle of the X509 certificates stored on the HSM.
+
+    <Returns>
+      List of all the available certificate handles.
+    """
+
+    x509_objects = self.session.findObjects([(PyKCS11.CKA_CLASS,
+        PyKCS11.CKO_CERTIFICATE)])
+    return x509_objects
```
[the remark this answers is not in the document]
tanishqjasoria: We don't need the certificates. I would remove the X509 certificate functionalities.

tanishqjasoria:
@lukpueh I agree on the change of the architecture. We should remove the two levels of abstraction. Apart from all the other benefits, it would make the interaction with HSM more transparent to the user.
We can use python-pkcs11 to make it more **pythonic** and easier to understand.

securesystemslib/hsm_keys.py
tanishqjasoria: Yes, there is a particular flow that needs to be followed while using this. 
It should give a more meaningful error to the user, I would change that.

lukpueh:
*Notes from an out-of-band conversation with @tanishqjasoria about the architecture:*

----


The API should look similar to securesystemslib.functions.gpg , that is:
- `create_signature` -- Create a signature using a key from the smartcard (via `pykcs11`), and return it in a securesystemslib-like format (see  `securesystemslib.formats.*SIGNATURE_SCHEMA`)
- `export_pubkey` -- Export a pubkey from the smartcard (via `pykcs11`), and return it in an securesystemslib-like format (see `securesystemslib.formats.*PUBKEY_SCHEMA`)
- `verify_signature` -- Take the signed data, and a pubkey and signature in securesystemslib-like format, and verify it (via `cryptography`). This should not require a smartcard or pykcs11.  

In addition we will need a public function that lists all the available keys from all the available hsms that function should give the user the necessary hsm/key identifiers required by  `create_signature` or `export_pubkey`.
I suggest that each call to those functions operates in a separate hsm session, that is they should be wrapped by setup and teardown (including login+authentication/logout if necessary). This could be implemented in a decorator but it is okay to just call some private functions, i.e. `_create_session`, `_login` on top, and `_logout`, `_destroy_session` on the bottom of those functions.

We might want to check how costly (computationally, memory-wise) these operations are, but I guess it’s feasible and definitely makes the interaction a lot nicer for the user.

Similarly, loading the  `pkcs11.so` library could be done automatically on import of `securesystemslib.hsm`. `PyKCS11`  just defaults to the `PYKCS11LIB`  environment variable if no path is passed. I suggest we just try using that. We can assign the library object to a module level global and use it in the functions where we need it. If `PYKCS11LIB`  is not set or the load fails for a different reason, we don’t do anything right away. But later, if the user calls a function for which the library is needed we inform them to either set the environment variable or load the library manually.

I picture something like this:

```python
# in module securesystemslib.hsm
import PyKCS11
PKCS11 = PyKCS11.PyKCS11Lib()
# Per default we try loading from PYKCS11LIB environment variable
try:
  PKCS11.load()
except PyKCS11.PyKCS11Error as e:
  # TODO: maybe add some more debug info
  log.debug(e)

def load_pkcs11_library(path=None):
  # TODO: Wrap in try/except and add nice error message/user feedback,
  # and re-raise as suited exception
  PKCS11.load(path)

def create_signature():
  # TODO: figure out a way to tell that PKCS11 has not been loaded
  if not PKCS11.loaded:
    # TODO: Raise suited exception informing user about the PYKCS11LIB envvar
    # or the option to use `load_pkcs11_library`
    raise
  # TODO:
  # open session
  # login
  # sign
  # logout
  # close session

# TODO: add other functions
```

trishankatdatadog:
Looks _great_ to me, thanks, Lukas and @tanishqjasoria!

securesystemslib/hsm.py
```diff
@@ -0,0 +1,473 @@
...
+    self.refresh()
+
+    # Get the list of slots on which HSMs are available
+    slot_list = self.PKCS11.getSlotList()
+    slot_info_list = []
+
+    # For all the available HSMs available, add relevant information
+    # to the slots dictionary
+    for slot in slot_list:
+      slot_dict = dict()
```
woodruffw: More of a style nit, but literal construction is a little easier to read and less repetitive:

```python
slot_dict = {
    'slot_id': slot,
    'slot_info': self.PKCS11.getSlotInfo(slot),
    # ...
}
slot_info_list.append(slot_dict)
```

securesystemslib/hsm.py
```diff
@@ -0,0 +1,473 @@
...
+    <Exceptions>
+      securesystemslib.exceptions.BadPasswordError, if the entered
+      user pin is invalid.
+    """
+
+    try:
+      self.session.login(user_pin)
+    except PyKCS11.PyKCS11Error as error:
+      if PyKCS11.CKR[error.value] == "CKR_USER_ALREADY_LOGGED_IN":
+        logger.warning('Already logged in as CKU_USER.')
```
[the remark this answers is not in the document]
woodruffw: +1 to this.

lukpueh:
Superseded by #229. Thanks for the solid groundwork, @tanishqjasoria! Closing here...

PR #163, opened 2019-03-05
Support gpg signatures (copied from in-toto project)

danixeee:
This PR relates to #55. I have mostly just copied and modified (where needed) _gpg_ signing implementation from [in-toto](https://github.com/in-toto/in-toto) project and kept it in separate module. Relevant tests are also added with some new ones to increase test coverage to 100%.

I have slightly changed [SIGNATURE_SCHEMA](https://github.com/openlawlibrary/securesystemslib/blob/danixeee/integrate-gpg-from-in-toto/securesystemslib/formats.py#L357) in order to support adding _gpg_ signatures with [append_signature](https://github.com/theupdateframework/tuf/blob/develop/tuf/repository_tool.py#L3137) function to TUF metadata.

I need some directions, do we want _gpg_ as a separate module, or `gpg/formats.py` should be added to existing `formats.py`, etc. ?

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #174, opened 2019-08-13
Add gpg support and custom (sub)process module

### LINKED ISSUE #55: Support GPG and SSH keys
*securesystemslib* provides an API to import and export public and private keys in PEM (RSA, ECDSA, also cf. #54) or in a proprietary format (Ed25519).

It would be convenient to extend the API so that users can load, e.g. their existing GPG or SSH keys. Note that *securesystemslib* does not necessarily have to provide an API to generate those keys. As @aaaaalbert has mentioned, user might even prefer to use their own well-known toolchain.

lukpueh:
**Fixes issue #**:

Supersedes  #163 
Closes #55 and https://github.com/in-toto/in-toto/issues/275

This PR adds a Python interface for GnuPG providing functions for:
- signature creation (using private keys in a local gpg keyring)
- public key export (from a local gpg keyring into a *pythonic* format),
- gpg signature verification function that uses the `cryptography` library and does not require the gpg command line tool

Additionally, this PR adds a:
- custom (sub)process convenience module, to i.a. hide Python version compatibility details

Both, the gpg sub-package and the process module, were originally developed as part of the `in-toto` project, and were transferred at [in-toto@9c374a9c970f0e8d985cbf06a7e6c7a7df3017c3](https://github.com/in-toto/in-toto/commit/9c374a9c970f0e8d985cbf06a7e6c7a7df3017c3), preserving parts of the original git history.  ~See commit messages of 62ba23974b74ef76cf82cbc40832c88d368066ff and de817db4afaffe2693249d7c21675c65915a5866 for git-technical details about `.~

For review only consider commits in the range of https://github.com/lukpueh/securesystemslib/compare/47a9d9a...lukpueh:add-gpg, earlier commits have already been reviewed as part of the in-toto project.

This PR further:
- drops Python 3.4 support in favor of the process module (9116205a7ab040ab00fcbda1cc9b898efdac34e9),
- lowers the test code coverage threshold to 99%, due to a hard-to-test gpg code branch (a79667a89ac846ce47dc4b000cf8c7d1e60dbaa5),
- fixes an unrelated docs bug that was introduced in #162 (ad996ea18e8b4c84a8b86df81056730f057f99c6) 

*DISCLAIMER:* This PR does not integrate the public gpg functions with securesystemslib's public function modules, i.e. `keys` or `interface`.
This also means, that the generic `keys.verify_signature` function can not be used to verify gpg signatures (use `gpg.functions.verify_signature` instead).

While such an integration seems desirable, I suggest to do this in a separate PR, e.g. in the course of a general API revision (see #35).


lukpueh:
*I just rebased onto master, force pushed and lost the merge commits that contained messages about how I used `git filter-branch` with `--subdirectory-filter` and `--index-filter` to preserve the git history of the gpg subpackage (and the process module).*

For documentation purposes here is what I did to pull in the gpg subpackage together with its history:

```bash
# Grab in-toto repo and create a new branch at tip of development
  git clone git@github.com:in-toto/in-toto.git
  cd in-toto
  git checkout -b in-toto-gpg

  # Make "in_toto/gpg" subdir the root of the repo
  git filter-branch --prune-empty --subdirectory-filter in_toto/gpg HEAD

  # Grab securesystemslib repo and fetch the unrelated in-toto branch that
  # only has the contents of in-toto's gpg subdirectory
  cd ..
  git clone git@github.com:secure-systems-lab/securesystemslib.git
  cd securesystemslib
  git remote add in-toto ../in-toto
  git fetch in-toto
  git checkout in-toto-gpg

  # Move all gpg files back into a corresponding securesystemslib/gpg subdir
  # NOTE: sed does not understand \t so we use a literal tab (Ctrl-v + TAB)
  git filter-branch --index-filter \
    'git ls-files -s | sed "s/$(printf \\t)/&\securesystemslib\/gpg\//" |
      GIT_INDEX_FILE=$GIT_INDEX_FILE.new \
        git update-index --index-info &&
     mv "$GIT_INDEX_FILE.new" "$GIT_INDEX_FILE"' HEAD

  # Merge the unrealted gpg branch into a new working branch based off of
  # securesystemslib's master
  git checkout master
  git checkout -b add-gpg
  git merge in-toto-gpg --allow-unrelated-histories

```

lukpueh:
Okay, I just cleaned up the history here. This is ready for reviewing. As mentioned above in the PR description, only https://github.com/lukpueh/securesystemslib/compare/47a9d9a...lukpueh:add-gpg is new code. The rest of the history was transferred from the in-toto repository.

@rennergade, do you want to give it a try?

securesystemslib/gpg/common.py
```diff
@@ -714,9 +712,18 @@ def parse_signature_packet(data, supported_signature_types=None,
   # conflict resolution scheme that makes more sense.
   # (see RFC4880 5.2.4.1.)
   # Below we only consider the last and favor hashed over unhashed subpackets
-  # TODO: Should we warn if a we use an unhashed subpacket?
-  for subpacket_type, subpacket_data in \
-      unhashed_subpacket_info + hashed_subpacket_info:
+  for idx, subpacket_tuple in \
+      enumerate(unhashed_subpacket_info + hashed_subpacket_info):
+
+    is_hashed = (idx >= len(unhashed_subpacket_info))
```
rennergade: As an outsider, it's a little unclear to me how `is_hashed` is verified. I think a comment is warranted on why the index is being compared to the length.

lukpueh:
Thanks for the quick review, @rennergade! 

We've been a little bit lenient regarding `import`-convention on the in-toto repo (where this code was written originally). There is a corresponding issue over there, but we labeled it *low-prio* (see https://github.com/in-toto/in-toto/issues/145).

securesystemslib/gpg/common.py
lukpueh: Let me know if 9906acd works for you.

PR #178, opened 2019-09-10
Move colorama to install_requires in setup.py

### LINKED ISSUE #153: Lack of colorama hard dependency
The current release blows up because `colorama` is not a dependency. Although it *should* get pulled in by either `cryptography` or `pynacl`, there is no information to describe that is the issue when using securesystemslib.

For example, this is the test log from `in-toto` using the latest securesystemslib (given that it doesn't hard depend on either library)

```
======================================================================
ERROR: tests.test_runlib (unittest.loader._FailedTest)
ImportError: Failed to import test module: tests.test_runlib
  File "/usr/lib64/python3.6/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/lib64/python3.6/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/home/santiago/Documents/personal/programas/in-toto/in-toto/tests/test_runlib.py", line 29, in <module>
    from in_toto.models.metadata import Metablock
  File "/home/santiago/Documents/personal/programas/in-toto/in-toto/in_toto/models/metadata.py", line 34, in <module>
    from in_toto.models.layout import Layout
  File "/home/santiago/Documents/personal/programas/in-toto/in-toto/in_toto/models/layout.py", line 49, in <module>
    import securesystemslib.interface
  File "/home/santiago/Documents/personal/programas/in-toto/in-toto/.tox/py36/lib/python3.6/site-packages/securesystemslib/interface.py", line 51, in <module>
    from colorama import Fore
ModuleNotFoundError: No module named 'colorama'
```

Needless to say, it's rather dangerous to assume that `cryptography` or `pynacl` will depend on colorama forever. Worse than that, I'm assuming that we may add support for other libraries which may or may not pull `colorama` in.

**Current behavior**:
Thing breaks unless you explicitly:

1. Install a dependency that pulls `colorama`, such as `pynacl` or `cryptography`.
2. Install colorama explicitly to compensate for this lack of hard-dependency

**Expected behavior**:
The library installs itself with everything it needs and doesn't make any assumptions of the packages that will be pulled and their upstream dependencies.

lukpueh:
**Fixes issue #**:
closes #153 

`colorama` was added to colorize output generated by some functions in `securesystemslib.interface`.

Prior to this PR `colorama` was listed as extra requirement in setup.py under `extras_require["crypto"]` alongside `cryptography`.

However, `securesystemslib.interface` does not properly handle an installation without `colorama`, as e.g. `secureystemslib.keys` does for an installation without `cryptography` (or `pynacl`), thus it cannot be an extra requirement.

Moreover, `colorama` does not require C extensions, which would be a good reason to keep it out of the standard installation.

requirements**:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,8 +96,9 @@ setup(
     'Topic :: Software Development'
   ],
   install_requires = ['six>=1.11.0', 'subprocess32; python_version < "3"',
-      'python-dateutil==2.8.0'],
-  extras_require = {'crypto': ['cryptography>=2.2.2', 'colorama>=0.3.9'],
+      'python-dateutil==2.8.0', 'colorama>=0.3.9'],
+  extras_require = {
+      'crypto': ['cryptography>=2.2.2'],
       'pynacl': ['pynacl>1.2.0']},
   packages = find_packages(exclude=['tests', 'debian']),
   scripts = []


SantiagoTorres:
Thanks, Lukas! This is something that always bothered me. Would it be worthwhile to later add support for non-colorama versions? (i.e., why pull output colorized on terminals that don't support it?)

lukpueh:
Yes absolutely, @SantiagoTorres! I think it would be useful to generally revise the code, related to optional dependencies (i.e. cyrptography, pynacl). IIRC, there is some unnecessary redundancy on one hand and then missing case handling on the other. I'll follow up with a separate issue.

But I think this is a reasonable quick fix to close #153.

SantiagoTorres:
Oh certainly, agreed. I think it'd be a possible first issue for new contributors. Thanks!

Issue #180, opened 2019-09-12
TempFile doesn't behave like a Python file object


joshuagl:
This is replicating the content of theupdateframework/tuf#160, as the `TempFile` class is now implemented in securesystemslib, not tuf.

**Description of issue or feature request**:

`TempFile` isn't a [file object](https://docs.python.org/3/glossary.html#term-file-object) and its methods aren't consistent with similar methods on a [file object](https://docs.python.org/3/glossary.html#term-file-object).

**Current behavior**:

There are inconsistencies between file operations on a `securesystemslib.util.TempFile` and [file object](https://docs.python.org/3/glossary.html#term-file-object)s from the Python standard library, for example:

```python
>>> import securesystemslib.util
>>> tf = securesystemslib.util.TempFile()
>>> import tempfile
>>> ptf = tempfile.TemporaryFile()
>>> f = open('tmpfile', 'w+')
>>> tf.write(b'Hello, world!')
>>> ptf.write(b'Hello, world!')
13
>>> f.write('Hello, world!')
13
>>> tf.read(3)
b''
>>> ptf.read(3)
b''
>>> f.read(3)
''
>>> tf.read()
b'Hello, world!'
>>> ptf.read()
b''
>>> f.read()
''
>>> tf.tell()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'TempFile' object has no attribute 'tell'
>>> ptf.tell()
13
>>> f.tell()
13
>>> tf.seek(0)
>>> ptf.seek(0)
0
>>> f.seek(0)
0
>>> tf.seek(0)
>>> ptf.seek(0)
0
>>> f.seek(0)
0
>>> tf.read(3)
b'Hel'
>>> ptf.read(3)
b'Hel'
>>> f.read(3)
'Hel'
>>> tf.read()
b'Hello, world!'
>>> ptf.read()
b'lo, world!'
>>> f.read()
'lo, world!'
```

**Expected behavior**:

Users familiar with Python will expect `TempFile` to behave like a [file object](https://docs.python.org/3/glossary.html#term-file-object)

It would be a usability improvement for Python developers if `TempFile` behaved like a [file object](https://docs.python.org/3/glossary.html#term-file-object). It should probably implement [`io.FileIO`](https://docs.python.org/3/library/io.html#io.FileIO).

joshuagl:
[...]

PR #181, opened 2019-09-19
Remove securesystemslib.util.TempFile

joshuagl:
**Fixes issue #**: #180 

**Description of the changes being introduced by the pull request**: remove the `securesystemslib.util.TempFile` class.

This was originally added to provide convenient handling of compressed and uncompressed file like objects however, those features of `securesystemslib.util.TempFile` are no longer used by tuf nor in-toto.
Furthermore `securesystemslib.util.TempFile` has some behavioural inconsistencies with a standard Python file object, which cause confusion for users of the class.

The following code changes are introduced:

1.  add a new helper `securesystemslib.util.persist_temp_file()` to create a copy of the temporary file object (first arg) at the path specified in the second argument in 4a36f97
2. remove all internal users of `securesystemslib.util.TempFile` in `securesystemslib.interface` in e594e66
3. remove the class itself and the associated unit tests in a6dca67

requirements**:


securesystemslib/util.py
```diff
@@ -395,6 +95,34 @@ def get_file_details(filepath, hash_algorithms=['sha256']):
   return file_length, file_hashes
 
 
+def persist_temp_file(temp_file, persist_path):
+  """
+  <Purpose>
+    Copies 'temp_file' (a file like object) to a newly created non-temp file at
+    'persist_path' and closes 'temp_file' so that it is removed.
+
+  <Arguments>
```
lukpueh: Would you mind adding `temp_file` to the argument section as well?

lukpueh: Just found out about GitHub suggestions. :) Does this work for you?

```suggestion
  <Arguments>
  temp_file:
    A file-like object, e.g. as returned by 'tempfile.TemporaryFile()'.
     
```

joshuagl: Great catch. I've added this to the first commit and force pushed an updated series.

lukpueh: Hooray!

joshuagl: > Just found out about GitHub suggestions. :) Does this work for you?

Apologies, I missed that. This change would have been fine.

lukpueh: No worries, yours is actually better. :)

joshuagl:
Excellent news, can't wait to see the releases. Do let me know if there's any way I can help. Otherwise I'll dive into some more tuf and ssl issues next week. 😊

PR #186, opened 2019-10-01
Fix a race condition in `process.run_duplicate_streams`

lukpueh:
**Fixes issue #**:
in-toto/in-toto#282

This PR fixes a race condition in a standard stream duplicating loop in `process.run_duplicate_streams`, where the subprocess would write to its standard streams after the last read in the parent process, and then terminate before another iteration would be started, and therefor leaving the last parts of the streams unread.

Also adds a note about buffered outputs in child process to the function's docstring.

requirements**:


lukpueh:
3cec8584f597ab07f734413bc275ef6a7d24a673 fixes another issue in `process.run_duplicate_streams` (see commit message and code comment for details). Kudos to @aaaaalbert for reminding me to look closer)

lukpueh:
Many thanks, @adityasaky!

Issue #194, opened 2019-12-16
Print expected type instead of its string representation


trishankatdatadog:
**Description of issue or feature request**:

Moved from the [TUF repo](https://github.com/theupdateframework/tuf/issues/243). Also see [this comment](https://github.com/theupdateframework/tuf/issues/243#issuecomment-530754200).

**Current behavior**:

Presently, if a user passes an object O other a list to a Schema ListOf object, the error message will print the string representation S of O. Unfortunately, the error message is obscured when the S is large (e.g. O is a set with hundreds of thousands of strings).

**Expected behavior**:

I think the error message would be just as effective if we simply printed the type of O.

PR #199, opened 2020-01-06
Make check_match() errors more readable for ListOf objects

joshuagl:
that is provided, the better.

**Fixes issue #194**:

When the ListOf object's check_match() fails due to an incompatible object
having been passed, the salient information for the error message is the
type of the passed object and the type of the object against which the
check_match() method was called, therefore print those types in the error
message and not the string representation of the objects.

requirements**:

### DIFF
--- a/securesystemslib/schema.py
+++ b/securesystemslib/schema.py
@@ -523,7 +523,8 @@ class ListOf(Schema):
   def check_match(self, object):
     if not isinstance(object, (list, tuple)):
       raise securesystemslib.exceptions.FormatError(
-          'Expected ' + repr(self._list_name) + ' but got ' + repr(object))
+          'Expected object of type {} but got type {}'.format(
+            self._list_name, type(object).__name__))
 
 
     # Check if all the items in the 'object' list


securesystemslib/schema.py
```diff
@@ -523,7 +523,8 @@ def __init__(self, schema, min_count=0, max_count=sys.maxsize, list_name='list')
   def check_match(self, object):
     if not isinstance(object, (list, tuple)):
       raise securesystemslib.exceptions.FormatError(
-          'Expected ' + repr(self._list_name) + ' but got ' + repr(object))
+          'Expected object of type %s but got type %s'
+          % (type(self._list_name), type(object)))
```
lukpueh: According to the docstring `_list_name` is *"A string identifier for the ListOf object."* So I'd...

```suggestion
          % (self._list_name, type(object)))
```
Otherwise you'll always get something like `<class 'str'>`.

lukpueh: Maybe it would be nicer to only show the name of the object's class, e.g.`str` instead of `<class 'str'>`...
```suggestion
          % (self._list_name, type(object).__name__))
```
What do you think?

lukpueh: I think the new-style  string formatting with `.format()` is generally preferable and we tend to use it in new code. But it's not a blocker.

joshuagl: I do prefer the new-style string formatting, but when I couldn't see string formatting in the style guide I decided to err on the side of caution. Especially as I think that `.format()` is only available in Python3  (whilst ssl and its users still want to support Python2)?

joshuagl:
Thanks for the incisive feedback. I've force-pushed an updated patch to address your two primary comments.

As I understand it we can not switch to the new style string formatting until we drop Python 2 support, which I understood to be something we don't want to rush into.

securesystemslib/schema.py
lukpueh: `format()` was added in 2.6 and we don't support <2.7 so that should be fine. We already use `format()` it in many other places in securesystemslib and in-toto. The code style guideline is a bit outdated in that regard (sorry!). Feel free to update your patch, otherwise I'll merge.

joshuagl: Thanks @lukpueh, I've updated to use `format()`

Issue #179, opened 2019-09-12
Revise extra dependency handling


lukpueh:
*[Updated on Jan 23, 2020]*

**Description of issue or feature request**:

securesystemslib lists some dependencies that require C-code (`cryptography` requires`openssl`, `pynacl` requires `libsodium`) as optional to allow for a pure-python installation.  The runtime handling of missing optional dependencies should be revised.

**Current behavior**:
`cryptography` and `pynacl` are listed as optional (extra) dependencies, but `securesystemslib` does not  fare (consistently) well, if installed without them.

**Expected behavior**:
- ~Public facing modules (e.g. `interface.py` and `keys.py`) must be importable, even if the optional dependencies are not installed.~ Fixed with #200
- ~Each public facing function always should be callable and present meaningful user-feedback if an optional dependency that is required for that function is not installed.~ Fixed with #200

- Also address or keep in mind recently merged or pending functionality, that has non-pure Python dependencies (#174, #169, #170). 

-~*Optional*: `colorama` was made a strict dependency  in #178 to quickfix #155. @SantiagoTorres, to consider making it a optional again (with respect to required adoptions as outlined above).~ Fixed with #200 

- It would be nice to fine-tune code coverage measurement (see https://github.com/secure-systems-lab/securesystemslib/pull/200#discussion_r368568724 ff.)

joshuagl:
When looking at available options for testing for the presence of optional dependencies it seems there's an old mechanism [`imp`](https://docs.python.org/3/library/imp.html), which was deprecated in Python 3.4 and a new mechanism [`importlib`](https://docs.python.org/3/library/importlib.html#module-importlib), which is available since Python 3.1.

securesystemslib's tox environment tests Python 2.7, 3.5 and 3.6 - but that doesn't necessarily mean those are the only versions we care to support. PyPI lists securesystemslib-0.11.3 as supporting Python 2.7, 3.4, 3.5 and 3.6.

It may be reasonable to take the same approach as Ansible (see Ansible commit [2732cde](https://github.com/ansible/ansible/commit/2732cde031a80636dfda31e7f3491a680c01b0bb)) and use `imp` for Python versions < 3 and `importlib` for anything else.

However with Python 2's EOL being [close](https://pythonclock.org) I wanted to get some feedback on whether it's even desirable to continue to maintain Python2 support in securesystemslib?
[...]

PR #206, opened 2020-02-06
Handle missing native dependencies in securesystemslib.gpg

joshuagl:
**Fixes issue #**: 
#179 

* Handle missing `gpg[2]` command in all functions that would call it and raise `UnsupportedLibraryError`
* Wrap native module imports in `securesystemslib.gpg` in `try/except` logic so that the modules can always be imported cleanly
* Handle missing native module dependencies in `securesystemslib.gpg` by raising `UnsupportedLibraryError` when a function requiring a native module dependency is called
* Test the `securesystemslib.gpg` interfaces requiring `gpg[2]` or native module dependencies from our pure python test environment

requirements**:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ matrix:
       env: TOXENV=py27
     - python: "2.7"
       env: TOXENV=purepy27
+      before_install:
+        - sudo apt-get remove -y --allow-remove-essential gnupg gnupg2
     - python: "3.5"
       env: TOXENV=py35
     - python: "3.6"
@@ -18,6 +20,8 @@ matrix:
       env: TOXENV=py38
     - python: "3.8"
       env: TOXENV=purepy38
+      before_install:
+        - sudo apt-get remove -y  --allow-remove-essential gnupg gnupg2
 
 install:
   - pip install -U tox coveralls
--- a/securesystemslib/gpg/util.py
+++ b/securesystemslib/gpg/util.py
@@ -24,6 +24,7 @@ from distutils.version import StrictVersion # pylint: disable=no-name-in-module,
 import cryptography.hazmat.backends as backends
 import cryptography.hazmat.primitives.hashes as hashing
 
+import securesystemslib.exceptions
 import securesystemslib.gpg.exceptions
 import securesystemslib.process
 import securesystemslib.gpg.constants
@@ -294,10 +295,18 @@ def get_version():
 
     The executed base command is defined in constants.GPG_VERSION_COMMAND.
 
+  <Exceptions>
+    securesystemslib.exceptions.UnsupportedLibraryError:
+            If the gpg command is not available
+
   <Returns>
     Version number string, e.g. "2.1.22"
 
   """
+  if not securesystemslib.gpg.constants.HAVE_GPG: # pragma: no cover
+    raise securesystemslib.exceptions.UnsupportedLibraryError(
+        securesystemslib.gpg.constants.NO_GPG_MSG)
+
   command = securesystemslib.gpg.constants.GPG_VERSION_COMMAND
   process = securesystemslib.process.run(command,
       stdout=securesystemslib.process.PIPE,
--- a/securesystemslib/gpg/util.py
+++ b/securesystemslib/gpg/util.py
@@ -21,8 +21,13 @@ import logging
 
 from distutils.version import StrictVersion # pylint: disable=no-name-in-module,import-error
 
-import cryptography.hazmat.backends as backends
-import cryptography.hazmat.primitives.hashes as hashing
+CRYPTO = True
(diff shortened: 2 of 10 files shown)


joshuagl:
Need to fix the `.travis.yml` changes to correctly remove (and possibly re-add) gpg et al.

.travis.yml
```diff
@@ -8,6 +8,10 @@ matrix:
       env: TOXENV=py27
     - python: "2.7"
       env: TOXENV=purepy27
+      before_install:
+        - sudo apt-get remove -y --allow-remove-essential gnupg gnupg2
+      after_script:
+        - sudo apt-get install -y gnupg gnupg2
```
lukpueh: Out of curiosity, what's idea of restoring this? Isn't the worker just dumped after running? Is there caching? Anyways, removing gnupg [removes apt](https://travis-ci.org/secure-systems-lab/securesystemslib/jobs/647078797#L258) too, so  the after script [command fails](https://travis-ci.org/secure-systems-lab/securesystemslib/jobs/647078797#L334).

securesystemslib/gpg/util.py
```diff
@@ -294,10 +309,18 @@ def get_version():
...
+    securesystemslib.exceptions.UnsupportedLibraryError:
+            If the gpg command is not available
+
   <Returns>
     Version number string, e.g. "2.1.22"
 
   """
+  if not securesystemslib.gpg.constants.HAVE_GPG: # pragma: no cover
+    raise securesystemslib.exceptions.UnsupportedLibraryError(
+      securesystemslib.gpg.constants.NO_GPG_MSG)
```
lukpueh: Please double-indent.

tests/check_public_interfaces.py
```diff
@@ -194,6 +197,18 @@ def test_purepy_ed25519(self):
...
+
+    with self.assertRaises(securesystemslib.exceptions.UnsupportedLibraryError):
+      securesystemslib.gpg.functions.create_signature('bar')
+
+    with self.assertRaises(securesystemslib.exceptions.UnsupportedLibraryError):
+      securesystemslib.gpg.functions.export_pubkey('f00')
+
+    with self.assertRaises(securesystemslib.exceptions.UnsupportedLibraryError):
+      securesystemslib.gpg.util.get_version()
+
```
lukpueh: Could you add a `verify_signature` call too? IIUC it should also raise an `UnsupportedLibraryError`.

.travis.yml
joshuagl: Whoops, my bad. I wasn't sure whether there was a new host instance per matrix entry and meant to check before pushing the PR.

I'll remove this.

securesystemslib/gpg/util.py
joshuagl: 🤦‍♂ will do!

tests/check_public_interfaces.py
joshuagl: Yes I can. That will throw `UnsupportedLibraryError` eventually, from whichever handler's `verify_signature` function. Does it make sense to add an explicit check and error early? That probably only makes the tests cleaner/easier and doesn't affect usage, so perhaps now.

joshuagl:
Thanks for the review @lukpueh !

> Also, IIUC, we should be able to remove [this block](https://github.com/secure-systems-lab/securesystemslib/blob/283cea3da2e30ede974e1cb96c50b6064f02bb05/tests/aggregate_tests.py#L37-L54) in aggregate_tests.py now (the call in L57 too), and in test_gpg.py replace
> occurrences of
> `@unittest.skipIf(os.getenv("TEST_SKIP_GPG"), "gpg not found")`
> with
> `@unittest.skipIf(not securesystemslib.gpg.constants.HAVE_GPG, "gpg not found")`
> 
> Right?

Absolutely right. That will be a nice clean up. I'll update the PR with that and your other requested changes on Monday.

tests/check_public_interfaces.py
lukpueh: > Does it make sense to add an explicit check and error early? 

I wondered too. Probably doesn't hurt. :)

> That probably only makes the tests cleaner/easier and doesn't affect usage, so perhaps now.

Not sure I understand why it makes the tests cleaner.

joshuagl: > > Does it make sense to add an explicit check and error early?
> 
> I wondered too. Probably doesn't hurt. :)

👍 

> > That probably only makes the tests cleaner/easier and doesn't affect usage, so perhaps now.
> 
> Not sure I understand why it makes the tests cleaner.

Perhaps not cleaner. But easier/shorter to write, as the arguments can be any type rather than needing to be correct enough to pass the initial parts of the function.

joshuagl:
I've update the PR to incorporate your suggested changes @lukpueh . Please take a look when you have the time.

Issue #210, opened 2020-02-24
Module is logging warnings on import


FlorianVeaux:
Hello 👋 

**Description of issue or feature request**:

When `securesystemslib` is installed without the `colors` [extra](https://github.com/secure-systems-lab/securesystemslib/blob/master/setup.py#L105), warnings/error logs are shown.

**Current behavior**:

On Python3, it complains about the colorama module not being found [here](https://github.com/secure-systems-lab/securesystemslib/blob/bbcbeca09d5017070f6cd68d8c9d1731c2e980aa/securesystemslib/interface.py#L59)

On Python2, it tries to complain for the same thing but is failing with the `No handlers could be found for logger "securesystemslib_interface"`.

**Expected behavior**:
The library should not complain about colorama as it is an optional dependency (and should also initialize the logger correctly on Python2)

**How to reproduce**:

Python2: `docker run -it python:2 bash -c "pip install securesystemslib[crypto,pynacl] && python -c 'import securesystemslib.interface'"`

Python3: `docker run -it python:3 bash -c "pip install securesystemslib[crypto,pynacl] && python -c 'import securesystemslib.interface'"`

PR #211, opened 2020-02-24
Do not log a warning if colorama is not installed

FlorianVeaux:
**Fixes issue #**:
https://github.com/secure-systems-lab/securesystemslib/issues/210

Remove the warning

requirements**:

More details available on the issue #210

**Additional notes**:
This PR doesn't fix the issue that the `logger` has not been initialized correctly. (Shows only on Python2 with `No handler could be found ...`). This would require a bit more knowledge of the library and an easy workaround for library users is to use `logging.raiseExceptions = False`

### DIFF
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -56,7 +56,7 @@ try:
   TERM_RED = Fore.RED
   TERM_RESET = Fore.RESET
 except ImportError:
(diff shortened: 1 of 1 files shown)


lukpueh:
Thanks pointing out these issues and providing a fix, @FlorianVeaux. We added the warning only recently (see #200). But you are right, a debug level message seems like a better fit. I'll merge this and follow up with a fix for the missing handler.

PR #214, opened 2020-02-25
Allow verifying signatures without gpg

FlorianVeaux:
that is provided, the better.

**Issue**:
A regression was introduced with https://github.com/secure-systems-lab/securesystemslib/pull/206
Before that it was possible to verify keys even if `gpg` was missing in the used environment.

gpg is not required for verifying but a safe-guard was added in the `verify_signature` method which now prevents that possibility.

Removes the safe-guard introduced in #206 

requirements**:

### DIFF
--- a/tests/check_public_interfaces.py
+++ b/tests/check_public_interfaces.py
@@ -204,9 +204,6 @@ class TestPublicInterfaces(unittest.TestCase):
       securesystemslib.gpg.functions.create_signature('bar')
 
     with self.assertRaises(securesystemslib.exceptions.UnsupportedLibraryError):
-      securesystemslib.gpg.functions.verify_signature(None, 'f00', 'bar')
-
-    with self.assertRaises(securesystemslib.exceptions.UnsupportedLibraryError):
       securesystemslib.gpg.functions.export_pubkey('f00')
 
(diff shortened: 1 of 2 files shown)


tests/check_public_interfaces.py
```diff
@@ -203,9 +203,6 @@ def test_gpg_cmds(self):
     with self.assertRaises(securesystemslib.exceptions.UnsupportedLibraryError):
       securesystemslib.gpg.functions.create_signature('bar')
 
-    with self.assertRaises(securesystemslib.exceptions.UnsupportedLibraryError):
```
trishankatdatadog: Would it make sense to say that we do _not_ raise this exception (regression test)?

FlorianVeaux: AFAICT this would require running the `securesystemslib.gpg.functions.verify_signature(None, 'f00', 'bar')` function with correctly formatted values. I'm not really sure how that could be added but this is a great idea.

lukpueh:
A comprehensive fix including regression testing is available in https://github.com/secure-systems-lab/securesystemslib/pull/215.

lukpueh:
Fixed in #215.

PR #223, opened 2020-03-27
Add documentation about how to run the tests

MVrachev:
I am a new contributor to this project and initially it was not clear for me how to run the tests. 
It's good to have that information somewhere visible.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -337,6 +337,8 @@ Miscellaneous functions
     >>> ecdsa_key2 = import_ecdsakey_from_pem(public_pem)
 
 
+
+
 GnuPG interface
 ~~~~~~~~~~~~~~~
 
@@ -374,10 +376,8 @@ Testing is done with **tox**, which can be installed with pip:
 
 
 Secure Systems Library supports multiple versions of Python.
-
 For that reason, the project is tested against multiple virtual environments with different Python versions.
-
-If you run 
+If you run
 ::
 
 $ tox
--- a/README.rst
+++ b/README.rst
@@ -60,6 +60,8 @@ Python).  Support for RSA, ECDSA, and E25519 via the `cryptography` and
     $ pip install securesystemslib[crypto]
     $ pip install securesystemslib[pynacl]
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -334,7 +334,30 @@ Miscellaneous functions
...
+::
+
+    $ pip install tox
+
+
+Secure Systems Library supports multiple versions of Python.
+
+For that reason, the project is tested against multiple virtual environments with different Python versions.
+
+If you run
```
lukpueh: I think we can put "*Secure Systems Library supports [...]. If you run*" (L346-L350) all into one paragraph without blank lines.

Also, please remove the trailing whitespace after run. :)

MVrachev:
> Many thanks for the patch, @MVrachev! This indeed should be very helpful for new contributors.
> 
> Can I ask you a little favor? It would be great if you could restructure the document so that usage docs and development docs are a bit better separated.
> Could you maybe add two new h2-level headings , e.g. "Usage" and "Testing" (underlined with `+`). And put everything from "Create RSA Keys" downwards under "Usage", and your newly added section under "Testing"?

Yes, no problem, I will do it.

> Besides, since securesystemslib used to be part of [theupdateframework/tuf](https://github.com/theupdateframework/tuf) the development and testing setup is very similar (see [TUF's testing docs](https://github.com/theupdateframework/tuf/blob/develop/docs/CONTRIBUTORS.rst#testing)). Do you think it might be helpful to link to tufs testing docs here in your newly added testing section?

What I understand is that the idea of `securesystemslib` is to provide the needed functionality for TUF but as a library, so that it can be used from other projects.
From that perspective, it makes sense `securesyslib` to have independent documentation, even though now there could be a duplication, because tomorrow the maintainers of this library can decide to use another approach for testing which is not used by TUF.

lukpueh:
> Yes, no problem, I will do it.

Thanks! :)

> ... independent documentation ...

You are absolutely right. Let's not mention TUF.

PR #225, opened 2020-04-02
Add default keyid to format_metadata_to_key

mnm678:
Add an optional parameter to format_metadata_to_key that allows the caller to provide a default keyid. This change can be used for future changes that remove keyid_hash_algorithms from TUF as described in https://github.com/theupdateframework/tuf/issues/848.

### DIFF
--- a/securesystemslib/keys.py
+++ b/securesystemslib/keys.py
@@ -474,7 +474,7 @@ def format_keyval_to_metadata(keytype, scheme, key_value, private=False):
 
 
 
-def format_metadata_to_key(key_metadata):
+def format_metadata_to_key(key_metadata, default_keyid=None):
   """
   <Purpose>
     Construct a key dictionary (e.g., securesystemslib.formats.RSAKEY_SCHEMA)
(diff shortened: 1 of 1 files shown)


securesystemslib/keys.py
```diff
@@ -544,7 +544,8 @@ def format_metadata_to_key(key_metadata):
 
   # Convert 'key_value' to 'securesystemslib.formats.KEY_SCHEMA' and generate
   # its hash The hash is in hexdigest form.
-  default_keyid = _get_keyid(keytype, scheme, key_value)
+  if (default_keyid == None):
```
joshuagl, line 547: According to [PEP 8](https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations) the correct way to check if a condition is `None` is with the identity operator `is`:
```suggestion
  if default_keyid is None:
```

securesystemslib/keys.py
```diff
@@ -474,7 +474,7 @@ def format_keyval_to_metadata(keytype, scheme, key_value, private=False):
 
 
 
-def format_metadata_to_key(key_metadata):
+def format_metadata_to_key(key_metadata, default_keyid=None):
```
joshuagl, line 477: Could you document default_keyid in the Arguments section of the docstring?

Issue #220, opened 2020-03-06
Remove `keys.format_metadata_to_key()`'s dependency on `settings.HASH_ALGORITHMS`


joshuagl:
**Description of issue or feature request**:

It should be possible to control the hash algorithms used for generating keyids in `securesystemslib.keys.format_metadata_to_key()` without having to change package level settings in `securesystemslib.settings`

Related to #219 

**Current behavior**:

The tuf reference implementation has code like the following, i.e. [repository_tool.py#L3062](https://github.com/theupdateframework/tuf/blob/679348ffcf99c8e6736e257f4c533f55f438a028/tuf/repository_tool.py#L3062)
```
hash_algorithms = securesystemslib.settings.HASH_ALGORITHMS
securesystemslib.settings.HASH_ALGORITHMS = key_metadata['keyid_hash_algorithms']
key_object, keyids = securesystemslib.keys.format_metadata_to_key(key_metadata)
securesystemslib.settings.HASH_ALGORITHMS = hash_algorithms
```

**Expected behavior**:

format_metadata_to_key can be called without having to change `securesystemslib.settings.HASH_ALGORITHMS`, something like:
```
key_object, keyids = securesystemslib.keys.format_metadata_to_key(key_metadata, hash_algorithms)
```

MVrachev:
I will work to fix this one.

Issue #219, opened 2020-03-06
Remove securesystemslib.settings


joshuagl:
**NOTE: this is a meta-bug/milestone.**

**Description of issue or feature request**:

The settings module is a legacy of securesystemslib originating as part of the tuf reference implementation and results in awkward cohesion between securesystemslib and its users.

The settings module is a prime example of this, there are places in securesystemslib where modifying the behaviour of a function requires changing values in `securesystemslib.settings`, which is non-intuitive. Particularly as many of the settings are only documented as comments in that file.

This meta-bug tracks the goal/milestone of removing `securesystemslib.settings` and should be linked to smaller, more tractable and more specific issues.

**Current behavior**:

Users have to modify securesystemslib internal settings, rather than being able to pass all required information as arguments to library functions.

**Expected behavior**:

Users are not required to modify internal library settings and can pass all required configuration values as arguments to the functions they are using.

JustinCappos:
Makes sense to me.  Thanks for suggesting this!

PR #202, opened 2020-01-21
Add default timeout setting helper to process module

lukpueh:
**Fixes issue #**:
None

**Description of the changes being introduced by the pull request**:
Add helper to use `settings.SUBPROCESS_TIMEOUT` as default argument in `process.run*` functions, and still be able to modify the setting in the settingsmodule after the function definitions are evaluated, i.e. after the process module is imported for the first time.

**Please verify and check that the pull request fulfils the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #227, opened 2020-04-03
Refactor format_metadata_to_key

MVrachev:
**Fixes issue #**:

https://github.com/secure-systems-lab/securesystemslib/issues/220

Refactor format_metadata_to_key so that it won't depend
on settings.HASH_ALGORITHMS.
There are couple of code snippets in tuf where in order to provide
a custom list of hash algorithms, tuf is temporary changing
settings.HASH_ALGORITHMS and then returning it to its formal state.

This could lead to problems in future and one should
be able to control hash algorithms used for generating keyids in f
ormat_metadata_to_key() without having to change
package level settings in securesystemslib.settings.

requirements**:

### DIFF
--- a/securesystemslib/keys.py
+++ b/securesystemslib/keys.py
@@ -474,7 +474,8 @@ def format_keyval_to_metadata(keytype, scheme, key_value, private=False):
 
 
 
-def format_metadata_to_key(key_metadata, default_keyid=None):
+def format_metadata_to_key(key_metadata, default_keyid=None,
+    keyid_hash_algorithms=None):
   """
   <Purpose>
     Construct a key dictionary (e.g., securesystemslib.formats.RSAKEY_SCHEMA)
@@ -522,6 +523,10 @@ def format_metadata_to_key(key_metadata, default_keyid=None):
       provided, the keyid will be calculated by _get_keyid using the default
       hash algorithm. If provided, the default keyid can be any string.
 
+    keyid_hash_algorithms:
+      An optional list of hash algorithms to use when generating keyids.
+      Defaults to securesystemslib.settings.HASH_ALGORITHMS.
+
   <Exceptions>
     securesystemslib.exceptions.FormatError, if 'key_metadata' does not conform
     to 'securesystemslib.formats.KEY_SCHEMA'.
@@ -553,7 +558,10 @@ def format_metadata_to_key(key_metadata, default_keyid=None):
   keyids = set()
   keyids.add(default_keyid)
 
-  for hash_algorithm in securesystemslib.settings.HASH_ALGORITHMS:
+  if keyid_hash_algorithms is None:
+    keyid_hash_algorithms = securesystemslib.settings.HASH_ALGORITHMS
+
+  for hash_algorithm in keyid_hash_algorithms:
     keyid = _get_keyid(keytype, scheme, key_value, hash_algorithm)
     keyids.add(keyid)
 
@@ -562,7 +570,7 @@ def format_metadata_to_key(key_metadata, default_keyid=None):
   key_dict['keytype'] = keytype
   key_dict['scheme'] = scheme
   key_dict['keyid'] = default_keyid
-  key_dict['keyid_hash_algorithms'] = securesystemslib.settings.HASH_ALGORITHMS
+  key_dict['keyid_hash_algorithms'] = keyid_hash_algorithms
   key_dict['keyval'] = key_value
 
   return key_dict, keyids
--- a/tests/test_keys.py
+++ b/tests/test_keys.py
@@ -26,6 +26,7 @@ from __future__ import division
 from __future__ import unicode_literals
 
 import unittest
+import copy
 
 import securesystemslib.exceptions
 import securesystemslib.formats
@@ -164,36 +165,50 @@ class TestKeys(unittest.TestCase):
 
 
   def test_format_metadata_to_key(self):
-    # Reconfiguring rsakey_dict to conform to KEY_SCHEMA
-    # i.e. {keytype: 'rsa', keyval: {public: pub_key, private: priv_key}}
-    keyid = self.rsakey_dict['keyid']
-    del self.rsakey_dict['keyid']
+    # Copying self.rsakey_dict so that rsakey_dict remains
+    # unchanged during and after this test execution.
+    test_rsakey_dict = copy.copy(self.rsakey_dict)
+    del test_rsakey_dict['keyid']
 
-    rsakey_dict_from_meta, junk = KEYS.format_metadata_to_key(self.rsakey_dict)
+    # Call format_metadata_to_key by using the default value for keyid_hash_algorithms
+    rsakey_dict_from_meta_default, junk = KEYS.format_metadata_to_key(test_rsakey_dict)
 
-    # Check if the format of the object returned by this function corresponds
+    # Check if the format of the object returned by calling this function with
+    # default hash algorithms e.g. securesystemslib.settings.HASH_ALGORITHMS corresponds
     # to RSAKEY_SCHEMA format.
-    self.assertEqual(None,
-        securesystemslib.formats.RSAKEY_SCHEMA.check_match(rsakey_dict_from_meta),
+    self.assertTrue(
+        securesystemslib.formats.RSAKEY_SCHEMA.matches(rsakey_dict_from_meta_default),
         FORMAT_ERROR_MSG)
 
-    self.assertEqual(None,
-        securesystemslib.formats.KEY_SCHEMA.check_match(rsakey_dict_from_meta),
+    self.assertTrue(
+        securesystemslib.formats.KEY_SCHEMA.matches(rsakey_dict_from_meta_default),
         FORMAT_ERROR_MSG)
 
-    self.rsakey_dict['keyid'] = keyid
+    # Call format_metadata_to_key by using custom value for keyid_hash_algorithms
+    rsakey_dict_from_meta_custom, junk = KEYS.format_metadata_to_key(test_rsakey_dict,
+        keyid_hash_algorithms=['sha384'])
+
+    # Check if the format of the object returned by calling this function with
+    # custom hash algorithms corresponds to RSAKEY_SCHEMA format.
+    self.assertTrue(
+        securesystemslib.formats.RSAKEY_SCHEMA.matches(rsakey_dict_from_meta_custom),
+        FORMAT_ERROR_MSG)
+
+    self.assertTrue(
+        securesystemslib.formats.KEY_SCHEMA.matches(rsakey_dict_from_meta_custom),
+        FORMAT_ERROR_MSG)
+
+    test_rsakey_dict['keyid'] = self.rsakey_dict['keyid']
 
     # Supplying a wrong number of arguments.
     self.assertRaises(TypeError, KEYS.format_metadata_to_key)
-    args = (self.rsakey_dict, self.rsakey_dict)
+    args = (test_rsakey_dict, test_rsakey_dict)
     self.assertRaises(TypeError, KEYS.format_metadata_to_key, *args)
 
     # Supplying a malformed argument to the function - should get FormatError
-    keyval = self.rsakey_dict['keyval']
-    del self.rsakey_dict['keyval']
+    del test_rsakey_dict['keyval']
     self.assertRaises(securesystemslib.exceptions.FormatError,
-        KEYS.format_metadata_to_key, self.rsakey_dict)
-    self.rsakey_dict['keyval'] = keyval
+        KEYS.format_metadata_to_key, test_rsakey_dict)
 
 
 


securesystemslib/interface.py
```diff
@@ -572,7 +572,7 @@ def import_ed25519_publickey_from_file(filepath):
   # also includes the keyid.
   ed25519_key_metadata = securesystemslib.util.load_json_file(filepath)
   ed25519_key, junk = \
-    securesystemslib.keys.format_metadata_to_key(ed25519_key_metadata)
+    securesystemslib.keys.format_metadata_to_key(ed25519_key_metadata, securesystemslib.settings.HASH_ALGORITHMS)
```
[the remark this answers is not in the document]
MVrachev: Addressed in a follow on commit.

securesystemslib/interface.py
```diff
@@ -817,7 +817,7 @@ def import_ecdsa_publickey_from_file(filepath):
   # also includes the keyid.
   ecdsa_key_metadata = securesystemslib.util.load_json_file(filepath)
   ecdsa_key, junk = \
-    securesystemslib.keys.format_metadata_to_key(ecdsa_key_metadata)
+    securesystemslib.keys.format_metadata_to_key(ecdsa_key_metadata, securesystemslib.settings.HASH_ALGORITHMS)
```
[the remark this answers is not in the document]
MVrachev: Addressed in a follow on commit.

securesystemslib/keys.py
```diff
@@ -474,7 +474,7 @@ def format_keyval_to_metadata(keytype, scheme, key_value, private=False):
 
 
 
-def format_metadata_to_key(key_metadata):
+def format_metadata_to_key(key_metadata, hash_algorithms=securesystemslib.settings.HASH_ALGORITHMS):
```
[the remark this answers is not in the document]
MVrachev: Addressed in a follow on commit.

securesystemslib/keys.py
```diff
@@ -518,6 +519,10 @@ def format_metadata_to_key(key_metadata, hash_algorithms=securesystemslib.settin
        'keyval': {'public': '...',
                   'private': '...'}}
 
+    hash_algorithms:
+      An optional parameter used to provide custom hash algorithms when
+      creating the key dictionary.
+
```
joshuagl, line 525: One purpose of the docstring is to prevent the reader from having to read the code to understand how the function works and what arguments to pass. Therefore it should tell us the type of the argument, especially where it isn't obvious from the function signature.

```suggestion
    hash_algorithms:
      An optional list containing the names of hash algorithms to use when
      generating keyids. Defaults to securesystemslib.settings.HASH_ALGORITHMS

```

MVrachev: Addressed in a follow on commit.

tests/test_keys.py
```diff
@@ -186,7 +186,7 @@ def test_format_metadata_to_key(self):
     # Supplying a wrong number of arguments.
     self.assertRaises(TypeError, KEYS.format_metadata_to_key)
     args = (self.rsakey_dict, self.rsakey_dict)
-    self.assertRaises(TypeError, KEYS.format_metadata_to_key, *args)
+    self.assertRaises(TypeError, ['sha256', 'sha512'], KEYS.format_metadata_to_key, *args)
```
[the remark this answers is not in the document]
MVrachev: > Looks like asserting for a `TypeError` is not enough here. Your are getting a `TypeError: 'list' object is not callable` , which is not what you want. I recommend to check out the documentation of unittest's `assertRaises` and maybe read up on *args and **kwargs. :)

Yes, that was a dummy mistake.
Thank you!

securesystemslib/keys.py
```diff
@@ -522,6 +522,10 @@ def format_metadata_to_key(key_metadata, default_keyid=None):
       provided, the keyid will be calculated by _get_keyid using the default
       hash algorithm. If provided, the default keyid can be any string.
 
+    hash_algorithms:
+      An optional parameter used to provide custom hash algorithms when
```
joshuagl, line 526: I may be over-reading this, but I'm finding the term "custom" to be problematic. It's not a custom hash algorithm, it's choosing which of the supported hash algorithms you want to use.
```suggestion
      An optional list of hash algorithms to use when
```

MVrachev: Yes, not a bad idea and this will provide information about the type of that argument which is good.

I add and squashed my previous commits.

securesystemslib/keys.py
```diff
@@ -474,7 +474,7 @@ def format_keyval_to_metadata(keytype, scheme, key_value, private=False):
 
 
 
-def format_metadata_to_key(key_metadata, default_keyid=None):
+def format_metadata_to_key(key_metadata, default_keyid=None, hash_algorithms=securesystemslib.settings.HASH_ALGORITHMS):
```
joshuagl, line 477: The [coding style](https://github.com/secure-systems-lab/code-style-guidelines) requests we keep lines to 79 characters or less with a double (4 space) indentation for continuation. Could you wrap this line?

```suggestion
def format_metadata_to_key(key_metadata, default_keyid=None,
    hash_algorithms=securesystemslib.settings.HASH_ALGORITHMS):
```

MVrachev: Yes, you are right. It seems I overlooked it.

Fixed and rebased on the master branch.

joshuagl:
This change looks good, thank you.

I would like to hold off on merging it until we have a solid release out to support tuf changes related to the Warehouse work for pep 458, because I anticipate that this will be the first of several API changes as we work to remove the settings module (re: #219) and I would like us to include those in one release rather than spread them over multiple releases. 

We just had the 0.15.0 release with abstract files & directories support. Next we'll want a release with the HSM support in #229. Then we'll have all of the features required of PEP 458 and can start landing changes like this towards #219.

It would be really nice if we tested `format_metadata_to_key` with non-default parameters too, do you think you could work on a patch for that in this PR @MVrachev ?

joshuagl:
> Also, be aware that default values in functions are only evaluated once when the function definition is evaluated. This has bitten me before (see #202). As a consequence, constructs like these won't be possible anymore.
> 
> ```python
> # The default arg is evaluated together with the function definition on first import:
> # format_metadata_to_key(key_metadata, hash_algorithms=['sha256', 'sha512'])
> import securesystemslib.keys
> import securesystemslib.settings
> 
> securesystemslib.settings.HASH_ALGORITHMS = ["blake2"]
> 
> format_metadata_to_key(key_metadata)
> # the default for hash_algorithms still is ['sha256', 'sha512'] and not ["blake2"]
> ```

I've been wondering what the right pattern for this is. Maybe something like:
```python
def format_metadata_to_key(key_metadata, default_keyid=None,
  keyid_hash_algorithms=None):
  # docstrings and stuff ...
  if keyid_hash_algorithms is None:
    keyid_hash_algorithms = securesystemslib.settings.HASH_ALGORITHMS
  # ... rest of the function
```

That way we don't bind the default value at function evaluation and we allow users to continue with the current pattern of assigning `securesystemslib.settings.HASH_ALGORITHMS`, making this change less of an API break. We're also a little bit cleaner in our intent to deprecate the settings module as we don't encode it in the interface (by using a variable from the module as a default value).

trishankatdatadog:
> I've been wondering what the right pattern for this is. Maybe something like:

Good idea, I like it!

MVrachev:
I rebased and addressed Joshua's comments.I changed the following: 

1) Renamed has_algorithms to keyid_hash_algorithms 

2) Use the approach proposed from Joshua Lock about handling the default value of keyid_hash_algorithms 

3) Added a new commit adding unit tests using nondefault hash algorithms.

tests/test_keys.py
```diff
@@ -197,6 +197,29 @@ def test_format_metadata_to_key(self):
 
 
 
+  def test_format_metadata_to_key_non_default_hash_algorithms(self):
```
joshuagl, line 200: I think we should just roll these tests into `test_format_metadata_to_key()` – that way we keep all tests of this function together, as is the pattern throughout the rest of the tests.

tests/test_keys.py
```diff
@@ -197,6 +197,29 @@ def test_format_metadata_to_key(self):
...
+    # Reconfiguring rsakey_dict to conform to KEY_SCHEMA
+    # i.e. {keytype: 'rsa', keyval: {public: pub_key, private: priv_key}}
+    keyid = self.rsakey_dict['keyid']
```
joshuagl, line 203: As a matter of taste I would much prefer that, rather than saving the value of the keyid here and restoring it later, we just make a copy of `rsakey_dict` to use within the test. Then we don't have to worry about restoring the class member to its pristine state at the end of the test.

tests/test_keys.py
```diff
@@ -197,6 +197,29 @@ def test_format_metadata_to_key(self):
...
+    del self.rsakey_dict['keyid']
+
+    rsakey_dict_from_meta, junk = KEYS.format_metadata_to_key(self.rsakey_dict,
+      keyid_hash_algorithms=['sha384'])
+
+    # Check if the format of the object returned by this function corresponds
+    # to RSAKEY_SCHEMA format.
+    self.assertEqual(None,
+        securesystemslib.formats.RSAKEY_SCHEMA.check_match(rsakey_dict_from_meta),
```
joshuagl, line 212: Should we use `assertTrue()` and [`matches()`](https://github.com/secure-systems-lab/securesystemslib/blob/09cf6d4fe8c5131944a8cf0f109fae7d14ba29b2/securesystemslib/schema.py#L70), rather than `assertEqual()` and [`check_match()`](https://github.com/secure-systems-lab/securesystemslib/blob/09cf6d4fe8c5131944a8cf0f109fae7d14ba29b2/securesystemslib/schema.py#L86), here? 

`matches()` returns a boolean to indicate whether the schema is matched, whereas `check_match()` returns `None` on a match and throws an exception on a mismatch. I don't think we want tests throwing unhandled exceptions and testing for a `None`-value or an `Exception` requires more code.

tests/test_keys.py
```diff
@@ -197,6 +197,29 @@ def test_format_metadata_to_key(self):
...
+        FORMAT_ERROR_MSG)
+
+    self.assertEqual(None,
+        securesystemslib.formats.KEY_SCHEMA.check_match(rsakey_dict_from_meta),
```
joshuagl, line 216: Same question on `matches()` vs `check_match()` here.

tests/test_keys.py
```diff
@@ -197,6 +197,29 @@ def test_format_metadata_to_key(self):
...
+        FORMAT_ERROR_MSG)
+
+    self.rsakey_dict['keyid'] = keyid
```
joshuagl, line 219: As above, we can avoid this if we make a local copy of the dict to use during this test.

tests/test_keys.py
MVrachev: That's a good suggestion. I will add it.

tests/test_keys.py
MVrachev: Addressed your comment by making changes and amending them to my last commit.

tests/test_keys.py
MVrachev: Addressed your comment by making changes and amending them to my last commit.

tests/test_keys.py
```diff
@@ -164,36 +165,50 @@ def test_import_rsakey_from_public_pem(self):
 
 
   def test_format_metadata_to_key(self):
-    # Reconfiguring rsakey_dict to conform to KEY_SCHEMA
-    # i.e. {keytype: 'rsa', keyval: {public: pub_key, private: priv_key}}
-    keyid = self.rsakey_dict['keyid']
-    del self.rsakey_dict['keyid']
+    # Copying self.rsakey_dict so that rsakey_dict remains
+    # unchanged during and after this test execution.
+    test_rsakey_dict = copy.deepcopy(self.rsakey_dict)
```
joshuagl, line 170: Nit: I think `copy` would suffice here?

MVrachev: Yes, you are right. 

I overdo it a little to make sure the right values were copied but forgot to try with `copy.copy`.

I made the change and amended.

securesystemslib/keys.py
```diff
@@ -474,7 +474,8 @@ def format_keyval_to_metadata(keytype, scheme, key_value, private=False):
 
 
 
-def format_metadata_to_key(key_metadata, default_keyid=None):
+def format_metadata_to_key(key_metadata, default_keyid=None,
+  keyid_hash_algorithms=None):
```
lukpueh, line 478: Minor nit: Please use double-indentation (4 spaces) for line continuation.

tests/test_keys.py
```diff
@@ -164,36 +165,50 @@ def test_import_rsakey_from_public_pem(self):
...
 
-    self.assertEqual(None,
-        securesystemslib.formats.KEY_SCHEMA.check_match(rsakey_dict_from_meta),
+    self.assertTrue(
+        securesystemslib.formats.KEY_SCHEMA.matches(rsakey_dict_from_meta_default),
+        FORMAT_ERROR_MSG)
+
+    # Call format_metadata_to_key by using custom value for keyid_hash_algorithms
+    rsakey_dict_from_meta_custom, junk = KEYS.format_metadata_to_key(test_rsakey_dict,
+      keyid_hash_algorithms=['sha384'])
```
lukpueh, line 189: See comment above about line continuation.

MVrachev:
Fixed the indentation.

MVrachev:
Thanks for merging that @joshuagl!
I will update the corresponding pr https://github.com/theupdateframework/tuf/pull/1016 in tuf next week.

PR #221, opened 2020-03-23
External Signing using PIV interface.

tanishqjasoria:
This PR aims to enable support for HSM. This is a revision of PR #170, to change the API interface for the support.

**Fixes issue #**: [tuf/issues/864](https://github.com/theupdateframework/tuf/issues/864)

**Description of the changes being introduced by the pull request**:

1. Module **hsm.py**: to enable support of hardware security modules through the PKCS#11 standard. This module use PyKCS11, a python wrapper (SWIG) for PKCS#11 modules to communicate with the cryptographic tokens using CCID/PIV interface.

2. Test Module **test_hsm.py**: to test various functionality for the HSM support.

**Please verify and check that the pull request fulfils the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #229, opened 2020-04-06
Add PKCS11-based HSM interface

lukpueh:
**Fixes issue #**:
Supersedes #170 and #221  (big kudos to @tanishqjasoria for excellent groundwork!)
~Bases on / incorporates #228 (needs rebase here after #228 is merged)~ (rebased on #228)
 
Add public interface to:
- list available HSMs (`get_hsms`), and
- keys on a given HSM (`get_keys_on_hsm`),
- sign data with a private key on the HSM, exporting the signature into a securesystemslib-like format (`create_signature`), and to
- export a key from the HSM using a sslib-like format (`export_pubkey`) to be used for signature verification.

Signature verification can be performed with the existing `securesystemslib.keys.verify_signature` function. 

Currently the HSM interface supports ecdsa keys and the `"ecdsa-sha2-nistp256"` and `"ecdsa-sha2-nistp384"` signature schemes.

Any interaction with the HSM interface require a one-time call to `load_pkcs11_lib` passing it the path to a PKCS11 dynamic library.

The commit also adds a basic test loop (sign + export key -> verify) using SoftHSM, as well interface tests that check that the public functions error gracefully if optional dependencies are not installed.

TODO:
- error handling
  - PKCS11 seems brittle, make sure that there are no state problems (e.g. openSession on a slot requires prior call to et slots, etc.) and we always get back the values we expect, i.e. handle hsm return invalid / incomplete data (e.g. on getAttributeValue), and fail gracefully if not
     HINT: check `CKR_<RETURN VALUE TYPE>` constants
  - revise error messages and exception taxonomy
  - Add `HSM_INFO_SCHEMA`, `HSM_KEY_INFO_SCHEMA`, `HSM_KEY_ID_SCHEMA` and check_match on them.

- docs
  - flesh out function docstrings and add code comments
  - add links to PKCS11 specs
  - add installation and usage docs, e.g. replace test_hsm.TestECDSAOnLUKPUEHsYubiKey with some instructions for YubiKey in README.md

requirements**:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,10 +100,11 @@ setup(
     'Topic :: Software Development'
   ],
   install_requires = ['six>=1.11.0', 'subprocess32; python_version < "3"',
-      'python-dateutil>=2.8.0'],
+      'python-dateutil>=2.8.0', 'asn1crypto; python_version > "3"'],
   extras_require = {
       'colors': ['colorama>=0.3.9'],
       'crypto': ['cryptography>=2.6'],
+      'pkcs11': ['PyKCS11>=1.5.7; python_version > "3"'],
       'pynacl': ['pynacl>1.2.0']},
   tests_require = 'mock; python_version < "3.3"',
   packages = find_packages(exclude=['tests', 'debian']),
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,13 @@ matrix:
       before_install:
         - sudo apt-get remove -y --allow-remove-essential gnupg gnupg2
 
+before_install:
+  - if [[ $TOXENV == "py35" || $TOXENV == "py36" || $TOXENV == "py37" || $TOXENV == "py38" || $TOXENV == py38-no-gpg ]]; then
+      sudo apt-get update && sudo apt-get install -y swig libsofthsm2-dev;
+      # NOTE: Will be '/usr/lib/softhsm/libsofthsm2.so' in bionic
+      export PYKCS11LIB=/usr/lib/x86_64-linux-gnu/softhsm/libsofthsm2.so;
+    fi
+
 install:
   - pip install -U tox coveralls
 
--- a/requirements-min.txt
+++ b/requirements-min.txt
@@ -2,3 +2,4 @@
 six
 python-dateutil
 subprocess32; python_version < '3'
+asn1crypto; python_version > '3'
--- a/requirements-pinned.txt
+++ b/requirements-pinned.txt
@@ -1,9 +1,11 @@
+asn1crypto==1.3.0         ; python_version > "3"
 cffi==1.14.0              # via cryptography, pynacl
 colorama==0.4.3
 cryptography==2.9
-enum34==1.1.6             ; python_version < "3" # via cryptography
+enum34==1.1.10            ; python_version < "3" # via cryptography
 ipaddress==1.0.23         ; python_version < "3" # via cryptography
 pycparser==2.20           # via cffi
+pykcs11==1.5.7            ; python_version > "3"
 pynacl==1.3.0
 python-dateutil==2.8.1
 six==1.14.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -37,3 +37,5 @@ colorama
 six
 python-dateutil
 subprocess32; python_version < '3'
+PyKCS11>=1.5.7; python_version > '3'
+asn1crypto; python_version > '3'
--- a/securesystemslib/hsm.py
+++ b/securesystemslib/hsm.py
@@ -1,0 +1,482 @@
+#!/usr/bin/env python
+"""
+<Program Name>
+  hsm.py
+
+<Started>
+  June 19, 2019.
+
+<Author>
+  Tanishq Jasoria <jasoriatanishq@gmail.com>
+  Lukas Puehringer <lukas.puehringer@nyu.edu>
+
+<Purpose>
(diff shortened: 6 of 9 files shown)


lukpueh:
Snippets to test with YubiKey on PIV slot

```bash
# Create key and sign it (the latter is important!)
yubico-piv-tool -a generate -s 9c -A ECCP256 -k -o public.pem
yubico-piv-tool -a verify-pin -a selfsign -s 9c -i public.pem -S /CN=a.b.c/OU=abc/O=b.c/ -o cert.pem
yubico-piv-tool -k -a import-certificate -s 9c -i cert.pem
```

```python
import securesystemslib.hsm
import securesystemslib.keys

sslib_key_id = \
    "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
scheme = securesystemslib.hsm.ECDSA_SHA2_NISTP256
data = b"data to be signed"

user_pin = "USER PIN" # TODO: Replace
pkcs11_so_path = "/path/to/pkcs11/shared/object/or/dynamic/lib" # TODO: replace
# E.g. /usr/local/Cellar/yubico-piv-tool/2.0.0/lib/libykcs11.dylib when
# installing with brew

# Init PKCS11
securesystemslib.hsm.load_pkcs11_lib(pkcs11_so_path)

# Get YubiKey assuming it is on the 1st slot
hsm_info = securesystemslib.hsm.get_hsms().pop()

key_infos = securesystemslib.hsm.get_keys_on_hsm(hsm_info)

# Grab keyid by label (NOTE: I promise better tooling ;)
hsm_key_id = [key_info["key_id"] for key_info in key_infos if key_info["label"] == "Public key for Digital Signature"].pop()

# Export public key
pub = securesystemslib.hsm.export_pubkey(
    hsm_info, hsm_key_id, scheme, sslib_key_id)

# Create signature
sig = securesystemslib.hsm.create_signature(
    hsm_info, hsm_key_id, user_pin, data, scheme, sslib_key_id)

# Verify signature
sig_valid = securesystemslib.keys.verify_signature(pub, sig, data)

assert sig_valid == True

```

jku:
Continuing documenting my attempt at testing:

Yubico does not document which tool should be used for this but I've now switched to yubico-piv-tool instead of ykman to reproduce Lukas's steps, I have reset the "management key" and...

```
(ykman) jkukkonen@jkukkonen-l:~/src/securesystemslib$ yubico-piv-tool -a generate -s 9c -A ECCP256 -k -o public.pem
Enter management key: 
Successfully generated a new private key.
(ykman) jkukkonen@jkukkonen-l:~/src/securesystemslib$ yubico-piv-tool -a verify-pin -a selfsign -s 9c -i public.pem -S /CN=a.b.c/OU=abc/O=b.c/ -o cert.pem
Enter PIN: 
Successfully verified PIN.
Failed signing certificate.
```

Even full verbosity gives nothing useful. This UX is worse than pgp AND it also doesn't work. I am not sure if I can do anything more at the moment

jku:
So I was finally able to run the snippet succesfully (thanks lukas):
* yubico-piv-tool is 2.2.0  (on debian) apparently creates broken keys that fail self-sign: installing to 2.3.0 and creating th key there helped
* the code does **not** work with opensc-pkcs11.so: signing fails with CKR_USER_NOT_LOGGED_IN. It does owrk with the yubico module. I am able to use pkcs11-tool to sign and verify with both modules

Using yubico-piv-tool 2.3.0 and Yubicos libykcs11.so as the pkcs library, the example code worked. 
* The first issue pivtool 2.2.0 being broken  does not need fixing (or can be addressed with a note in docs)
* the pkcs module issue maybe should be somehow mitigated: I don't understand who is supposed to choose the PKCS11 module and why?

setup.py
```diff
@@ -100,10 +100,11 @@
     'Topic :: Software Development'
   ],
   install_requires = ['six>=1.11.0', 'subprocess32; python_version < "3"',
-      'python-dateutil>=2.8.0'],
+      'python-dateutil>=2.8.0', 'asn1crypto; python_version > "3"'],
```
jku, line 103: I mentioned this elsewhere but this seems wrong: it should be only needed for "pkcs11" extra, right?

jku (commented):
A few years later, some review comments (I'm mostly looking at this from the POV of the Signer interface and  #447):

I think we should not expose all of this as a new module -- especially things like `get_keys_on_hsm()`, `get_hsms()` seem like things that should not be public API here

`HSMSigner` on the other hand sounds like a reasonable thing: The API would include 
* key and device ids as constructor arguments (optional if PyCKS supports that)
* PKCS module loaded using environment variable (optional if there are reasonable defaults for some systems)
* PIN as constructor variable

This would be compatible with #447 with a URI like `hsm:device/0/slot/c9` or `hsm:?slot=c9&device=0`, and pin getting requested with `secrets_handler()`
 
The remaining functionality would then be `export_pubkey()`. Interestingly this same issue comes up with KMS keys... Maybe _public key export_ could actually be a method on the Signer as well: not all implementations would support it but some could?

Another thing that comes up with KMS keys as well is the need to prehash the payload. We should add that support somewhere -- figuring out the algorithm from the public key is currently very annoying. Can we have `Key.get_payload_hash_algorithm()`?

Left also some code comments -- they may be obsolete so feel free to disregard

lukpueh:
Superseded by #472, which adds the relevant parts of this implementation to the new Signer/Key API.

PR #230, opened 2020-04-08
Remove support for gzipped files in load_json_file

joshuagl:
Compression related functions were removed from securesystemslib as it was
deemed more appropriate for callers to handle compression/decompression at
a different level. However, the automatic decompression functionality in
load_json_file() had remained.

Remove that change and tidy up load_json_file() to only close the
fileobject once. Since Python 2.5 a finally block is always called,
regardless of whether there is an exception or not. Therefore there's no
need to close the fileobject in the else block and the finally block.

Signed-off-by: Joshua Lock <jlock@vmware.com>

that is provided, the better.

requirements**:

### DIFF
--- a/securesystemslib/util.py
+++ b/securesystemslib/util.py
@@ -26,7 +26,6 @@ from __future__ import division
 from __future__ import unicode_literals
 
 import os
-import gzip
 import shutil
 import logging
 import tempfile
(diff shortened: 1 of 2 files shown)


securesystemslib/util.py
```diff
@@ -345,7 +336,6 @@ def load_json_file(filepath):
       ' Python object: ' + repr(filepath))
 
   else:
-    fileobject.close()
```
lukpueh, line 348: 💯 for en passant weeding

lukpueh:
To the interested reader, see [TAP 10](https://github.com/theupdateframework/taps/blob/master/tap10.md) for context. And to @joshuagl, many tanks for removing stale code.

Issue #222, opened 2020-03-24
Create private key files with umask 177


lukpueh:
**Description of issue or feature request**:
Transfer of https://github.com/theupdateframework/tuf/issues/279.

According to common practice (see e.g. ssh utilities) private key files should be created with read and write permissions for the user only (umask 177). Securesystemslib's `interface.generate_and_write_{rsa, ed25519, ecdsa}_keypair` functions should adopt that behavior.

**Current behavior**:
Private key files are created with the OS' default umask.

**Expected behavior**:
Private keys files are created with umask 177.

lukpueh:
I suggest to update `securesystemslib.util.persist_temp_file` to take an optional octal permissions argument and pass `600` in `interface.generate_and_write_{rsa, ed25519, ecdsa}_keypair` functions.

sechkova:
Looking at the issue, I noticed that the intention is 

> private key files should be created with read and write permissions for the user only

which means _file permissions_ 600 and _umask_ 177 (umask unsets the bits).
Can you confirm that my understanding is correct here?

lukpueh:
You are absolutely right, @sechkova! Thanks for pointing this out. :)

joshuagl:
Fixed in #231

PR #231, opened 2020-04-16
Create private key files with read and write permissions for the user only 

sechkova:
**Fixes issue #**: #222 

Private key files should be created with read and write permissions for the user only (`0o600`). 

- `util.persist_temp_file()` takes an optional permissions argument
- add an optional permissions argument to `StorageBackendInterface.put()` and `FilesystemsBackend.put()`
-  `interface.generate_and_write_{rsa, ed25519, ecdsa}_keypair` functions pass a permissions value `0o600` when creating the private key file 

Note: file permissions are calculated as default file mode (`0o777`) - umask (`0o177`)

requirements**:


sechkova:
I came across the 'mode' limitation in the Python os module on Windows platforms coming from the different access control implementations in Unix and Windows. It is documented only for the [os.chmod](https://docs.python.org/3.7/library/os.html#os.chmod) function but it applies also for `os.open` and `os.umask`:

`Note Although Windows supports chmod(), you can only set the file’s read-only flag with it (via the stat.S_IWRITE and stat.S_IREAD constants or a corresponding integer value). All other bits are ignored.`

Meaning that the same concepts for restricting Group/Other permissions do not apply on Windows. 

I added a workaround about it in the test case, however, I will be happy to understand the level of Windows support in securesystemslib.

joshuagl:
Now that we have abstract files and directories (#232) landed, let's figure out how to implement this change in the new world.

I think it probably makes sense to have an optional argument to `securesystemslib.storage. StorageBackendInterface.put()` that modifies the on-disk permissions. Implementers of `StorageBackendInterface` can choose what to do with this optional parameter, for example they might choose to store the permissions in their storage system's metadata and apply them at `get()` time? Or they might choose to ignore them.

What do you think @sechkova ?

sechkova:
An optional argument to` StorageBackendInterface.put()` makes sense to me too + applying the changes from this PR on the `FilesystemBackend` implementation.

sechkova:
Let me know what do you think.
And my Windows concerns from the comment above remain unchanged 🙂

securesystemslib/interface.py
```diff
@@ -69,6 +69,8 @@
 # Supported key types.
 SUPPORTED_KEY_TYPES = ['rsa', 'ed25519']
 
+# Private keys are created with read and write permissions for the user only
+PRIVATE_KEY_MODE = 0o600
```
jku, line 73: I know some people prefer the octal form so I don't make any demands, just note that this could be 
`PRIVATE_KEY_MODE = stat.S_IRUSR | stat.S_IWUSR`

jku (changes requested):
4 months later I get to this :| 

On Windows: I think it's ok to not support Windows ACLs as long as it's documented -- we're doing what we can with the file permissions available.

My only request for change is umask(), assuming it doesn't get much more complex with os.open(). The others are just comments.

securesystemslib/interface.py
joshuagl: Yes, let's use the constants.

sechkova:
A loong time has passed since the opening of this pull request ... let me resolve the conflicts and I will revive the thread.

securesystemslib/storage.py
```diff
@@ -219,12 +223,28 @@ def __exit__(self, exc_type, exc_val, traceback):
...
     # If we are passed an open file, seek to the beginning such that we are
     # copying the entire contents
     if not fileobj.closed:
       fileobj.seek(0)
 
+    # When calculating file permissions, the current OS umask value is first
+    # masked out. If custom permissions are set, we calculate a new umask
+    # followed by a call to open() with default permissions (0o777).
+    # This trick is done since open() does not support mode(permissions) as
+    # input parameter, alternatively os.open() can be used.
```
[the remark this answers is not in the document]
sechkova: I had a look again. If we use `os.open()` its documentation says _"When computing mode, the current umask value is first masked out."_ Meaning if we want to avoid race conditions caused by `os.umask`, we can pass the desired permissions to `os.open` but the resulting file permissions maybe restricted even further due to the applied mask.
 
This is not necessarily a bad thing for this concrete case where we want to restrict keys permissions. We can put the same line as a disclaimer in the docs.

securesystemslib/storage.py
```diff
@@ -233,14 +240,34 @@ def __exit__(self, exc_type, exc_val, traceback):
...
+    except OSError:
+      pass
+
+
+    try:
+      if mode is None:
+        fd = os.open(filepath, os.O_WRONLY|os.O_CREAT)
+      else:
+        fd = os.open(filepath, os.O_WRONLY|os.O_CREAT, mode)
+
```
sechkova, line 269: I don't like these lines, it looks ugly.
My other option was to use `0o777 `as a default value for mode same as` os.open() `but I wanted to avoid setting a concrete value and leave it to the `os.open` implementation.
Let me know if you have different ideas.

securesystemslib/interface.py
sechkova: Used!

sechkova:
I changed a bit the original idea. 
As pointed in https://github.com/secure-systems-lab/securesystemslib/pull/231#discussion_r511364276 `os.unmask `is not recommended for a library so I decided to wrap `os.open`. The desired permissions are set but os.umask is masked out first. 
This makes the user responsible of calling `os.umask` if needed.

securesystemslib/storage.py
jku: seems ok to me.

Question on fd though: does fd get closed by `with os.fdopen()`? It seems like that would make sense but please check

jku (commented):
Left some comments.

Also, can you add a test (or modify an existing one) for the actual beef of the PR: that _private keys_ are written with the correct permissions?

securesystemslib/storage.py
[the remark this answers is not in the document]
sechkova: Yep :facepalm:

securesystemslib/storage.py
```diff
@@ -233,14 +240,34 @@ def __exit__(self, exc_type, exc_val, traceback):
...
+    # input parameter, alternatively os.open() can be used.
+
+
+    # If a file with the same name already exists, the new permissions
+    # may not be applied.
     try:
-      with open(filepath, 'wb') as destination_file:
+      os.remove(filepath)
+    except OSError:
+      pass
```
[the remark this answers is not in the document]
sechkova: The problem would be that `persist_temp_file()` and `storage_backend.put()` are used not only for keys but also for metadata files where I think overwriting is an expected side effect.
Raising an error will change the behaviour slightly or we can add another parameter to opt-out of overwriting?
@joshuagl, as the author of the StorageBackend, any hints?

securesystemslib/storage.py
sechkova: The file descriptor seems to be closed together with the file object, at least my tests confirm it. Couldn't find it documented really.

jku: It makes sense that it would, I was just suspicious :)

securesystemslib/storage.py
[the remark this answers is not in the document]
jku: Got it. In any case you're not changing existing functionality in this -- the overwriting is just more visible now

sechkova:
Pushed a new bunch of commits:
- adding test for creating private keys with restricted permissions
- fixing wrong comments and pointing to `os.open` `mode` documentation in the docstrings

Related to overwriting keys, since this is the current behaviour:
- added a side effect in the docstrings of _generate_keypair wrappers that keys with the same file path will be overwritten
- @joshuagl recommended addressing the problem in a separate issue

sechkova:
Reworded the last commit 01431b5

tests/test_util.py
```diff
@@ -255,8 +256,21 @@ def test_B9_persist_temp_file(self):
     dest_path = os.path.join(dest_temp_dir, self.random_string())
     tmpfile = tempfile.TemporaryFile()
     tmpfile.write(self.random_string().encode('utf-8'))
-    securesystemslib.util.persist_temp_file(tmpfile, dest_path)
+
+    # Write a file with custom permissions
+    # stat.S_IRUSR: user has read permissions
+    # stat.S_IWUSR: user has write permissions
+    mode= stat.S_IRUSR | stat.S_IWUSR
```
jku, line 263:
```suggestion
    mode = stat.S_IRUSR | stat.S_IWUSR
```

jku (commented):
Code looks fine to me. I have two high level comments though (I'm sorry for not coming up with them earlier already).

1. StorageBackendInterface is supposedly abstract (I assume possibly not backed by a filesystem at all) but now we expect it to work with POSIX file permission -- should this have been a more abstract: "private/not private"? Or have a I misunderstood what the ABC is about?

2. The fact that the ABC method signature changes is a bit annoying: I think linters pick it up though so any out-of-tree implementations of the ABC _should_ notice when they update SSLib (can anyone confirm from memory if linters check for the ABC method signature?)

Still, might be worth notifying the known implementors or even asking their opinion on the change. Does warehouse implement their own StorageBackendInterface?

tests/test_util.py
joshuagl: Fixed in [6250c79](https://github.com/secure-systems-lab/securesystemslib/pull/231/commits/6250c790c9b9c317e661b345b345bede42c77cb9)

securesystemslib/interface.py
```diff
@@ -790,6 +804,7 @@ def generate_and_write_ecdsa_keypair(password, filepath=None):
 
   Side Effects:
     Writes key files to disk.
+    Overwrites filepath if already exists.
```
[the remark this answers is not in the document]
joshuagl: Fixed in [6250c79](https://github.com/secure-systems-lab/securesystemslib/pull/231/commits/6250c790c9b9c317e661b345b345bede42c77cb9)

joshuagl:
I just pushed updates to rebase this PR on master and address Jussi's first comment in https://github.com/secure-systems-lab/securesystemslib/pull/231#pullrequestreview-555379734

> StorageBackendInterface is supposedly abstract

I changed the permissions parameter to a boolean `restrict` which lets implementing `StorageBackendInterface`'s determine how best to interpret what restricted means. For `FileSystemBackend` the implementation sets the permissions of the file to 0o600 when `restrict=True`.

securesystemslib/storage.py
```diff
@@ -208,14 +217,31 @@ def get(self, filepath:str) -> Iterator[BinaryIO]:
...
+
+    try:
+      if restrict:
+        # On UNIX-based systems restricted files are created with read and
+        # write permissions for the user only (octal value 0o600).
+        fd = os.open(filepath, os.O_WRONLY|os.O_CREAT,
+          stat.S_IRUSR|stat.S_IWUSR)
+      else:
+        # Use the default value (0o777) of the 'mode' argument of os.open()
```
lukpueh, line 241: Minor comment nits:
Is it worth to mention that passed and default permissions may be silently masked out by umask, or limited to 666 by windows? And is it worth to mention what 666 and 777 means?

tests/test_util.py
```diff
@@ -240,8 +241,18 @@ def test_B9_persist_temp_file(self):
...
+    # Write a file with restricted permissions
+    securesystemslib.util.persist_temp_file(tmpfile, dest_path, restrict=True)
     self.assertTrue(dest_path)
+
+    # Need to set also the stat.S_IFREG bit to match the st_mode output
+    # stat.S_IFREG - Regular file
+    expected_mode = stat.S_IFREG | stat.S_IRUSR | stat.S_IWUSR
+    if os.name != 'posix':
+      # Windows only supports setting the read-only attribute.
+      expected_mode = stat.S_IFREG | 0o666
```
lukpueh, line 254: Minor consistency nit:

```suggestion
      expected_mode = stat.S_IFREG | stat.S_IWUSR | stat.S_IRUSR | stat.S_IWGRP | stat.S_IRGRP | stat.S_IWOTH | stat.S_IROTH
```

tests/test_util.py
```diff
@@ -240,8 +241,18 @@ def test_B9_persist_temp_file(self):
...
+
+    # Write a file with restricted permissions
+    securesystemslib.util.persist_temp_file(tmpfile, dest_path, restrict=True)
     self.assertTrue(dest_path)
+
+    # Need to set also the stat.S_IFREG bit to match the st_mode output
+    # stat.S_IFREG - Regular file
+    expected_mode = stat.S_IFREG | stat.S_IRUSR | stat.S_IWUSR
+    if os.name != 'posix':
+      # Windows only supports setting the read-only attribute.
```
lukpueh, line 253: Is this comment accurate? Doesn't 666 mean rw-rw-rw?

lukpueh (changes requested):
Thanks for the PR, @sechkova and @joshuagl! Code looks good. I have two minor suggestions and one more important question inline...

tests/test_util.py
joshuagl: It's _really_ hard to locate and parse the docs around this. 

Python's [`os.open()`](https://docs.python.org/3/library/os.html#os.open) points to the Windows [docs for `_open()`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/open-wopen?redirectedfrom=MSDN&view=msvc-170) in their C Runtime which say that possible modes on Windows are:


pmode | Meaning
-- | --
_S_IREAD | Only reading permitted.
_S_IWRITE | Writing permitted. (In effect, permits reading and writing.)
_S_IREAD \| _S_IWRITE | Reading and writing permitted.

which I read as being effectively 0o666 on Windows. Some additional confusion creeps in because Python's [`os.chmod()`](https://docs.python.org/3/library/os.html#os.chmod) docs state:

> Note Although Windows supports [chmod()](https://docs.python.org/3/library/os.html#os.chmod), you can only set the file’s read-only flag with it (via the stat.S_IWRITE and stat.S_IREAD constants or a corresponding integer value). All other bits are ignored.

but I assume this means the mode cannot be changed from `stat.S_IREAD` to `stat.S_IWRITE`. However, I don't have access to a Windows box to verify. Nor do we have CI set up for Windows, so perhaps that (having Windows testing) is a blocker for this PR?

securesystemslib/storage.py
joshuagl: I'll take a stab at including that info.

tests/test_util.py
lukpueh: Thanks for the detailed research. Just ran the test on a windows box and it works. Having Windows CI does not seem to be a blocker for this PR.

joshuagl: OK, great. I think I will push an extra commit to change the `os.name != 'posix'` to `os.name == 'nt'` for additional clarity (I don't expect anyone is using sslib on QNX, but 🤷).

tests/test_util.py
```diff
@@ -240,8 +241,18 @@ def test_B9_persist_temp_file(self):
...
-    securesystemslib.util.persist_temp_file(tmpfile, dest_path)
+
+    # Write a file with restricted permissions
+    securesystemslib.util.persist_temp_file(tmpfile, dest_path, restrict=True)
     self.assertTrue(dest_path)
+
+    # Need to set also the stat.S_IFREG bit to match the st_mode output
+    # stat.S_IFREG - Regular file
+    expected_mode = stat.S_IFREG | stat.S_IRUSR | stat.S_IWUSR
+    if os.name != 'posix':
```
joshuagl, line 252:
```suggestion
    if os.name == 'nt':
```

securesystemslib/storage.py
joshuagl, line 241:
```suggestion
        # Non-restricted files use the default 'mode' argument of os.open()
        # granting read, write, and execute for all users (octal mode 0o777).
        # NOTE: mode may be modified by the user's file mode creation mask
        # (umask) or on Windows limited to the smaller set of OS supported
        # permisssions.
```
WDYT @lukpueh ?

tests/test_util.py
joshuagl: Added that change as a suggestion https://github.com/secure-systems-lab/securesystemslib/pull/231/files#r988760854

lukpueh:
@joshuagl, are you going to commit these changes, or are you waiting for me to press the button? :D

joshuagl:
> @joshuagl, are you going to commit these changes, or are you waiting for me to press the button? :D

Oh, thanks for the nudge. Batched and committed.

lukpueh (approved):

PR #232, opened 2020-04-17
Implement filesystem abstraction

joshuagl:
**Note**: this branch of securesystemslib doesn't currently work with theupdateframework/tuf – it's very much WIP.

It feels like the shape these changes are taking introduces a conflict between the desire to support abstract filesystem backends vs. reasonable feature enhancements that have been suggested such as #222. It's not yet clear to me how we will support both abstract filesystem backends _and_ creating private key files with restricted privileges.

**Fixes issue #**: step one, two and four of theupdateframework/tuf#1009

* Add an abstract base class defining an interface for storage
* Add a concrete implementation of the storage interface for accessing the local file system 
* Add abstract storage support to `securesystemslib.hash.digest_filename()`
* Add abstract storage support to `securesystemslib.util.[get_file_details|ensure_parent_dir|load_json_file]()`

requirements**:

### DIFF
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -55,7 +55,7 @@ try:
   from colorama import Fore
   TERM_RED = Fore.RED
   TERM_RESET = Fore.RESET
-except ImportError:
+except ImportError: # pragma: no cover
   logger.debug("Failed to find colorama module, terminal output won't be colored")
   TERM_RED = ''
   TERM_RESET = ''
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -43,6 +43,7 @@ import random
 
 import securesystemslib.formats
 import securesystemslib.settings
+import securesystemslib.storage
 import securesystemslib.util
 import securesystemslib.keys
 
@@ -242,7 +243,8 @@ def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
 
 
 def import_rsa_privatekey_from_file(filepath, password=None,
-    scheme='rsassa-pss-sha256', prompt=False):
+    scheme='rsassa-pss-sha256', prompt=False,
+    storage_backend=None):
   """
   <Purpose>
     Import the PEM file in 'filepath' containing the private key.
@@ -271,6 +273,11 @@ def import_rsa_privatekey_from_file(filepath, password=None,
     prompt:
       If True the user is prompted for a passphrase to decrypt 'filepath'.
       Default is False.
+
+    storage_backend:
+      An object which implements
+      securesystemslib.storage.StorageBackendInterface. When no object is
+      passed a FilesystemBackend will be instantiated and used.
 
   <Exceptions>
     ValueError, if 'password' is passed and 'prompt' is True.
@@ -344,8 +351,11 @@ def import_rsa_privatekey_from_file(filepath, password=None,
     logger.debug('No password was given. Attempting to import an'
         ' unencrypted file.')
 
+  if storage_backend is None:
+    storage_backend = securesystemslib.storage.FilesystemBackend()
+
   # Read the contents of 'filepath' that should be a PEM formatted private key.
-  with open(filepath, 'rb') as file_object:
+  with storage_backend.get(filepath) as file_object:
     pem_key = file_object.read().decode('utf-8')
 
   # Convert 'pem_key' to 'securesystemslib.formats.RSAKEY_SCHEMA' format.
@@ -360,7 +370,8 @@ def import_rsa_privatekey_from_file(filepath, password=None,
 
 
 
-def import_rsa_publickey_from_file(filepath, scheme='rsassa-pss-sha256'):
+def import_rsa_publickey_from_file(filepath, scheme='rsassa-pss-sha256',
+    storage_backend=None):
   """
   <Purpose>
     Import the RSA key stored in 'filepath'.  The key object returned is in the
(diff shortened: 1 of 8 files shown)


securesystemslib/storage.py
```diff
@@ -0,0 +1,219 @@
...
+      A list containing the names of the files in the folder. May be an empty
+      list.
+    """
+    raise NotImplementedError # pragma: no cover
+
+
+
+
+
+class FilesystemBackend(StorageBackendInterface):
```
mnm678, line 155: Should this be FileSystemBackend?

securesystemslib/util.py
```diff
@@ -77,15 +89,15 @@ def get_file_details(filepath, hash_algorithms=['sha256']):
     raise securesystemslib.exceptions.Error('Path ' + repr(filepath) + ' doest'
         ' not exist.')
 
-  filepath = os.path.abspath(filepath)
```
mnm678, line 80: Is this handled by the FilesystemBackend?

joshuagl: It isn't handled by the FilesystemBackend. This would probably be OK to leave in, as `abspath()` doesn't access the underlying files or directories.

We don't do this kind of normalisation everywhere we take paths, thus my thought process when making this change was to try and move towards expecting normalised paths in securesystemslib functions. That's probably better as a separate PR and discussion, though.

I'll force push a change to add this back.

securesystemslib/storage.py
joshuagl: I've seen file system and filesystem used, I don't have a strong preference. I'd be happy to change it.

joshuagl:
Thanks for the review @mnm678

> This looks good to me, but it would be nice to get a review from a non-TUF ssl user as well.

Agreed. @SantiagoTorres, would you be able to take a look?

> Regarding #222, would it be possible to require this as part of the abstract class? There could be a put_secure method that uses the permissions you would want for a key file.

That's an interesting suggestion. @sechkova proposed similar when we discussed #222 offline. 
What I don't know is, if we added a `put_secure()` or similar to `StorageBackendInterface`, would that make sense on systems like S3 and GCS? 
How would it be implemented? Would they store the permission bits as metadata on the file and apply it during `get()`?

Perhaps @woodruffw can help us understand how this might work on S3 and similar storage?

joshuagl:
I just tested this branch with the _develop_ branch of in-toto and the tests run without issue.

securesystemslib/interface.py
```diff
@@ -664,10 +694,11 @@ def import_ed25519_privatekey_from_file(filepath, password=None, prompt=False):
 
   # Finally, regardless of password, try decrypting the key, if necessary.
   # Otherwise, load it straight from the disk.
```
woodruffw, line 696: Nit: `the disk` -> `storage`?

securesystemslib/storage.py
woodruffw: I have a _slight_ preference for `FilesystemBackend` based on "filesystem" being a full noun, but this backend doesn't affect me in either case 😉

securesystemslib/storage.py
```diff
@@ -0,0 +1,219 @@
...
+  """
+  <Purpose>
+    A concrete implementation of StorageBackendInterface which interacts with
+    local filesystems using Python standard library functions.
+  """
+
+  def get(self, filepath):
```
woodruffw, line 162: Proposal: Can we turn this into a context manager? Right now, this requires the following API pattern:

```python
foo = backend.get("foo")
# operations
foo.close()
```

...which makes it easy for a consumer to accidentally leak a file descriptor. With [contextlib](https://docs.python.org/3/library/contextlib.html), we could do something like this:

```python
with backend.get("foo") as foo:
  # operations...
```

joshuagl: Great suggestion, thanks. I want to spend a bit of time with the documentation to figure out a clean way to do this without _requiring_ use of a context manager. I would like to be able to support the current explicit `get()`/`close()` pattern as well as the `with get() as ...` pattern.

Would you expect the interface to indicate/recommend implementing a context manager? When I looked during initial work on this PR I couldn't see an obvious way to do that beyond documentation.

joshuagl: I've converted `get()` to a ContextManager in both the interface (docstrings) and the `FilesystemBackend`. As implemented `get()` can only be used as a context manager. I'd appreciate any thoughts on that.

joshuagl:
I've updated the PR here to note the `get()` function of the interface should be a context manager and updated the _FilesystemBackend_ implementation to return a ContextManager from `get()`. 

I've also addressed other comments from @mnm678 and @woodruffw. I'd very much appreciate some re-review.

securesystemslib/storage.py
```diff
@@ -0,0 +1,248 @@
...
+
+class FilesystemBackend(StorageBackendInterface):
+  """
+  <Purpose>
+    A concrete implementation of StorageBackendInterface which interacts with
+    local filesystems using Python standard library functions.
+  """
+
+  class GetFile(object):
+    # Implementing get() as a function with the @contextmanager decorator
```
woodruffw, line 169: Thanks for this explanation!

securesystemslib/storage.py
```diff
@@ -0,0 +1,248 @@
...
+        raise securesystemslib.exceptions.StorageError(
+            "Can't open %s" % self.filepath)
+
+
+    def __exit__(self, exc_type, exc_val, traceback):
+      self.file_object.close()
+
+
+
+# Map our class ContextManager implementation to the function expected of the
```
woodruffw, line 194: Nit: Mismatched indentation? Ignore me if it's a style choice 🙂

joshuagl: Thanks for catching that. It wasn't intentional. Fixed in a force-push

PR #198, opened 2019-12-18
Prepare Debian downstream release 0.16.0-1

lukpueh:
**Fixes issue #**:
None

**Description of the changes being introduced by the pull request**:
* Override never released 0.13.1-1 changelog entry, including note-worthy upstream changes since 0.11.3.
* Update upstream release signing key
  Joshua Lock (GPG on YubiKey) <jlock@vmware.com>
  08F3409FCF71D87E30FBD3C21671F65CB74832A4
* Bump standards version

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #236, opened 2020-05-12
Prepare next release (v0.15.0)

joshuagl:
* Allow Blake (blake2s and blake2b) hashing algorithms (#218)
* *new features*
  * Add nistp384 signature verification support (#228)
  * Allow callers to provide a default keyid in format_metadata_to_key, rather
    than using the default keyid value of a hash of the canonical JSON of the
	key metadata (#225)
  * Implement files and directories abstraction as an abstract base class;
    StorageBackendInterface, with a concrete implementation for local
	filesystems; FilesystemBackend  (#232). This enables users, such as tuf,
	to support non-local/non-traditional filesystems, so long as they provide
	an object implementing securesystemslib.storage.StorageBackendInterface.
	All functions which take a StorageBackendInterface default to creating a
	FilesystemBackend object for local filesystem interaction when an object
	isn't provided. This means that behaviour remains the same as in prior
	(0.14.x) releases of securesystemslib, only instead of throwing exceptions
	from the Python standard library a custom, generic, error is thrown:
	securesystemslib.exceptions.StorageError
* *removed features*
  * Remove support for gzipped files in load_json_file (#230)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 # Changelog
 
+## securesystemslib v0.15.0
+
+* Allow Blake (blake2s and blake2b) hashing algorithms (#218)
+* *new features*
+  * Add nistp384 signature verification support (#228)
+  * Allow callers to provide a default keyid in format_metadata_to_key, rather
+    than using the default keyid value of a hash of the canonical JSON of the
+	key metadata (#225)
+  * Implement files and directories abstraction as an abstract base class;
+    StorageBackendInterface, with a concrete implementation for local
+	filesystems; FilesystemBackend  (#232). This enables users, such as tuf,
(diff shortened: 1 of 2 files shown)


SantiagoTorres:
so I think we've fell behind a little bit, but I wonder if we want to update the debian metadata in this PR or we'd like to do this on a separate one...

joshuagl:
I'm not familiar with expectations around the Debian metadata. Are we building/publishing Debian packages somewhere? Or trying to get securesystemslib and friends into Debian?

We have an open PR for Debian packaging #198, perhaps we can handle this separately by updating that PR for 0.15.0?

SantiagoTorres:
Yeah, I think we could do a separate PR. Let me go ahead and prepare this release

Issue #241, opened 2020-06-02
create_folder does nothing (and does not throw) if filepath is an empty string


joshuagl:
**Description of issue or feature request**:

In order to support existing behaviour in tuf the `create_folder` method of `FilesystemBackend ` does nothing and does not raise an error if passed an empty string.

This feels like counter-intuitive behaviour that doesn't make sense for the storage backend. Ideally this would be resolved in tuf by removing calls to `create_folder` with an empty path i.e. performing any special casing in the caller.

See #240

**Current behavior**:

When `securesystemslib.storage.FilesystemBackend.create_folder()` is called with an empty string it returns immediately.

**Expected behavior**:

When `securesystemslib.storage.FilesystemBackend.create_folder()` is called with an empty string it raises an error, which better mimics the behaviour of `os.mkdir()`:
```Python
>>> os.mkdir('')
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    os.mkdir('')
FileNotFoundError: [Errno 2] No such file or directory: ''
```

joshuagl:
If memory serves this change will require fixes in tuf to not call `create_folder` with an empty path. We should also update the `TestStorageBackend` implementation so that it does not encode this behaviour.

woodruffw:
Thanks for opening this! I agree that this behavior should be resolved in tuf and that the invariant should be modified (to raising, as expected).

lukpueh:
I second that motion. Let's raise a `StorageError` here.

MVrachev:
I will push a pr to fix this.

In my changes, I will raise 
```
raise securesystemslib.exceptions.StorageError(
            "Can't create a folder with an empty filepath!")
```

exception.

I am not sure how we are going to handle the special case when we call `create_folder` with empty string in tuf.
It sounds to me that it's normal to not catch that exception and receive it. After all, how are we supposed to recover if `create_folder` was called with an empty string in tuf?

woodruffw:
Sounds good to me. It sounded like there were some historical reasons for `""` being silently ignored, but those reasons are opaque to me 🙂

joshuagl:
> I am not sure how we are going to handle the special case when we call `create_folder` with empty string in tuf.
> It sounds to me that it's normal to not catch that exception and receive it. After all, how are we supposed to recover if `create_folder` was called with an empty string in tuf?

[...]

PR #244, opened 2020-06-18
securesystemslib: Add AnyNonemptyString, use for PATH_SCHEMA

woodruffw:
TODO: Needs tests.

**Fixes issue #**: #241.

requirements**:

### DIFF
--- a/securesystemslib/schema.py
+++ b/securesystemslib/schema.py
@@ -206,7 +206,7 @@ class AnyString(Schema):
 
 
 
-class AnyNonemptyString(Schema):
+class AnyNonemptyString(AnyString):
   """
   <Purpose>
     Matches any string with one or more characters.
@@ -235,14 +235,8 @@ class AnyNonemptyString(Schema):
     False
   """
 
-  def __init__(self):
-    pass
-
-
   def check_match(self, object):
-    if not isinstance(object, six.string_types):
-      raise securesystemslib.exceptions.FormatError('Expected a string'
-          ' but got ' + repr(object))
+    super().check_match()
 
     if object == "":
         raise securesystemslib.exceptions.FormatError('Expected a string'
--- a/securesystemslib/schema.py
+++ b/securesystemslib/schema.py
@@ -206,6 +206,8 @@ class AnyString(Schema):
 
 
 
+
+
 class AnyNonemptyString(AnyString):
   """
   <Purpose>
@@ -241,6 +243,9 @@ class AnyNonemptyString(AnyString):
     if object == "":
         raise securesystemslib.exceptions.FormatError('Expected a string'
             ' with at least one character but got ' + repr(object))
+
+
+
 
 
 class AnyBytes(Schema):
--- a/securesystemslib/schema.py
+++ b/securesystemslib/schema.py
@@ -236,7 +236,7 @@ class AnyNonemptyString(AnyString):
   """
 
   def check_match(self, object):
-    super().check_match()
+    super().check_match(object)
 
     if object == "":
         raise securesystemslib.exceptions.FormatError('Expected a string'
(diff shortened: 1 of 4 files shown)


woodruffw:
`test_B3_file_in_confined_directories` currently fails on this:

```python
self.assertTrue(in_confined_directory('a/b/c.txt', ['']))
```

...presumably because that second parameter is now being enforced via `PATH_SCHEMA` as a nonempty string. My intuition is that this should really become `assertFalse`, but that's a behavioral change. Thoughts @lukpueh @joshuagl @mnm678 @trishankatdatadog?

lukpueh:
> ...presumably because that second parameter is now being enforced via PATH_SCHEMA as a non-empty string. My intuition is that this should really become assertFalse, but that's a behavioral change. Thoughts @lukpueh @joshuagl @mnm678 @trishankatdatadog?

I see this function for the first time. According to a comment, an empty string in the second parameter acts as some sort of "arbitrarily chosen" wildcard:

https://github.com/secure-systems-lab/securesystemslib/blob/9b3a78e998412c39418efb29f411591b20cdd236/securesystemslib/util.py#L216-L217

The only place where this function is used is in `tuf.mirrors.get_list_of_mirrors`, which btw. I'm also looking at for the first time. @trishankdatdatadog, do you know more about this? Confined directories are nowhere mentioned in the specification. Do we really need this function as `securesystemslib.util`? To me it looks like something that could be an inline check  in `get_list_of_mirrors`.

@willwoodruff, for the time being I suggest to just patch that function to not validate the second argument as list of paths but rather as list of any strings (including empty string).

securesystemslib/schema.py
```diff
@@ -206,6 +206,47 @@ def check_match(self, object):
...
+  """
+
+  def __init__(self):
+    pass
+
+
+  def check_match(self, object):
+    if not isinstance(object, six.string_types):
+      raise securesystemslib.exceptions.FormatError('Expected a string'
+          ' but got ' + repr(object))
```
lukpueh, line 245: Minor suggestion: We could subclass from AnyString and replace this first check with `super().check_match`.
Subclassing would also make it safe to remove the constructor here. Because the implicitly called `AnyString.__init__(): pass` will then prevent us from calling any further upper-level constructors (... which for some reason seem to be desired, or otherwise I don't understand the many `__init__(): pass` here).

woodruffw:
> for the time being I suggest to just patch that function to not validate the second argument as list of paths but rather as list of any strings (including empty string).

Sounds good to me. JFYI, you tagged another Will Woodruff 😉 

I should also have some bandwidth for unit tests today.

securesystemslib/schema.py
woodruffw: Good call! `super()` is a bit messy since this code (appears to be) Python 2.7 compatible, so I called `AnyString.check_match(self, object)` explicitly. We could use `super` if we turned the top-level `Schema` class into a "new-style" one, but I figured that this was less invasive for a small change.

trishankatdatadog:
> Confined directories are nowhere mentioned in the specification. Do we really need this function as `securesystemslib.util`?

Yeah, this is some non-standard optional thing we should really get rid off. That, and the list of mirrors. Nobody uses them AFAIK. I would wholeheartedly endorse and merge any PR that does this.

lukpueh:
> > Confined directories are nowhere mentioned in the specification. Do we really need this function as `securesystemslib.util`?
> 
> Yeah, this is some non-standard optional thing we should really get rid off. That, and the list of mirrors. Nobody uses them AFAIK. I would wholeheartedly endorse and merge any PR that does this.

Thanks for weighing in, @trishankatdatadog. I opened an issue: https://github.com/secure-systems-lab/securesystemslib/issues/245

PR #248, opened 2020-06-29
Restore PATH_SCHEMA to be AnyString

MVrachev:
**Fixes issue #**: https://github.com/theupdateframework/tuf/issues/1059

In one of the latest commits AnyNonemptyString was introduced
and with it, PATH_SCHEMA was changed to AnyNonemptyString from
AnyString. This led to unit tests failures in TUF.

One could argue what is the right approach: to consider
that PATH_SCHEMA is every non-empty string or every string.
I think this is not of much importance, but for TUF and
possibly for other securesystemslib adopters, this leads to tests
failures. The reason is that there are multiple functions
with arguments which should conformant against the PATH_SCHEMA
but have default values of `''`

See :
- "location_in_repository" argument in create_new_project function
both of which are in tuf/developer_tool.py: https://github.com/theupdateframework/tuf/blob/49031b424b65e852194c911f85bc29c14ad97432/tuf/developer_tool.py#L519
- "prefix" argument in load_repository function:
https://github.com/theupdateframework/tuf/blob/49031b424b65e852194c911f85bc29c14ad97432/tuf/developer_tool.py#L758

requirements**:

### DIFF
--- a/securesystemslib/formats.py
+++ b/securesystemslib/formats.py
@@ -128,7 +128,7 @@ KEYIDS_SCHEMA = SCHEMA.ListOf(KEYID_SCHEMA)
 SCHEME_SCHEMA = SCHEMA.AnyString()
 
 # A path string, whether relative or absolute, e.g. 'metadata/root/'
-PATH_SCHEMA = SCHEMA.AnyNonemptyString()
+PATH_SCHEMA = SCHEMA.AnyString()
 PATHS_SCHEMA = SCHEMA.ListOf(PATH_SCHEMA)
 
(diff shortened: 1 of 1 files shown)


lukpueh:
I'd rather update TUF than restore the change here. We just added this in https://github.com/secure-systems-lab/securesystemslib/pull/244 for good reason, that is to fix https://github.com/secure-systems-lab/securesystemslib/issues/241.

MVrachev:
Yes, you are right.
I will close that pr and submit a pr in tuf.

Issue #183, opened 2019-09-19
Revise schema and formats facility


lukpueh:
**Description of issue or feature request**:
Here are several loosely ordered observations in regards to securesystemslib's [schema](https://github.com/secure-systems-lab/securesystemslib/blob/v0.11.2/securesystemslib/formats.py)/[formats](https://github.com/secure-systems-lab/securesystemslib/blob/v0.11.2/securesystemslib/schema.py) facility:

**Current behavior**:
- Some schemas **sound more specific than they are**, e.g. `RELPATH_SCHEMA == PATH_SCHEMA == AnyString`. This is misleading, when assessing a function's capability of sanitizing inputs.

- Some schemas are an **odd replacement for constants**, e.g. schemas that define hard-coded strings, like `"rsassa-pss-sha256"`, and are then used to check the same strings, which have been hardcoded into function default arguments.

- Schema validation seems generally **overused**.  In TUF nearly every function runs `check_match` for each argument. IMHO argument sanitizing is mostly important in public-facing interfaces. Programming errors, on the other, should be caught through extensive testing and code review.
Using it everywhere makes the code bloated, also because there are a lot of generic comments describing the checks, and the obligatory and also quite generic `FormatError, if <arg> is not properly formatted`- blocks in the `<Exceptions>` section of docstrings. 

- Schema checking sometimes makes **execution branches unreachable**  (also see https://github.com/secure-systems-lab/code-style-guidelines/issues/18):

  ```python
  X_OR_Y_SCHEMA.check_match(arg)

  if arg == X:
    # ...
  elif arg == Y:
    # ...
  else:
    raise WillNeverBeRaisedError()
  ```
  And I've even seen cases where `WillNeverBeRaisedError` is also listed in the docstring as Exception that might be raised (documentation rot)

- The error messages from checking schemas with the `check_match` method are often not helpful, because they usually don't show the value of the checked variable or lack context.

**Expected behavior**:

- Review existing schemas for their validity/strictness, especially when chained.

- Only use in public facing interfaces (open for discussion). 
This would also require a clearer definition of what functions should be public interfaces, which btw. TUF/in-toto integrators would greatly benefit from.

- At least don't blindly  add `<arg>.<SCHEMA>.check_match` to every function. Coordinate with the rest of the function, and its purpose.

- Make sure the error message is helpful, e.g. by:
  - supporting an error message override argument in `check_match`, or by
[...]

PR #250, opened 2020-07-03
include keyid in public ed25519/ecdsa public key files

shibumi:
that is provided, the better.

**Fixes issue**: https://github.com/in-toto/docs/issues/33

In the past we did not include the keyid in public ed25519/ecdsa key files. This commit changes this behavior and introduces a new argument to the `format_keyval_to_metadata` function. The argument is optional,
so this should not have any affect on using `format_keyval_to_metadata` for keyid generation.

requirements**:

Do we need further tests/documentation for this change? The new argument is optional, so it should be API stable, however the generated public key itself may be different from now on. I have not tested this PR with in-toto-keygen yet.

### DIFF
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -516,13 +516,15 @@ def generate_and_write_ed25519_keypair(filepath=None, password=None):
   # to final destination.
   file_object = tempfile.TemporaryFile()
 
-  # Generate the ed25519 public key file contents in metadata format (i.e.,
-  # does not include the keyid portion).
+  # Generate the ed25519 public key file contents in metadata format
+  # We can include the keyid here, because the keyid is calculated
+  # by the key data without the private key.
   keytype = ed25519_key['keytype']
+  keyid = ed25519_key['keyid']
   keyval = ed25519_key['keyval']
   scheme = ed25519_key['scheme']
   ed25519key_metadata_format = securesystemslib.keys.format_keyval_to_metadata(
-      keytype, scheme, keyval, private=False)
+      keytype, scheme, keyval, keyid=keyid, private=False)
 
   file_object.write(json.dumps(ed25519key_metadata_format).encode('utf-8'))
 
@@ -786,8 +788,9 @@ def generate_and_write_ecdsa_keypair(filepath=None, password=None):
   keytype = ecdsa_key['keytype']
   keyval = ecdsa_key['keyval']
   scheme = ecdsa_key['scheme']
+  keyid = ecdsa_key['keyid']
   ecdsakey_metadata_format = securesystemslib.keys.format_keyval_to_metadata(
-      keytype, scheme, keyval, private=False)
+      keytype, scheme, keyval, keyid=keyid, private=False)
 
   file_object.write(json.dumps(ecdsakey_metadata_format).encode('utf-8'))
 
--- a/securesystemslib/keys.py
+++ b/securesystemslib/keys.py
@@ -378,7 +378,7 @@ def generate_ed25519_key(scheme='ed25519'):
 
 
 
-def format_keyval_to_metadata(keytype, scheme, key_value, private=False):
+def format_keyval_to_metadata(keytype, scheme, key_value, keyid=None, private=False):
   """
   <Purpose>
     Return a dictionary conformant to 'securesystemslib.formats.KEY_SCHEMA'.
@@ -464,7 +464,15 @@ def format_keyval_to_metadata(keytype, scheme, key_value, private=False):
 
   else:
     public_key_value = {'public': key_value['public']}
-
+    # If we encounter a keyid, we are dealing with pub key file generation
+    # as in interface.py#L526
+    if keyid is not None:
+      return {'keytype': keytype,
+            'scheme': scheme,
+            'keyid': keyid,
+            'keyid_hash_algorithms': securesystemslib.settings.HASH_ALGORITHMS,
+            'keyval': public_key_value}
+    
     return {'keytype': keytype,
             'scheme': scheme,
             'keyid_hash_algorithms': securesystemslib.settings.HASH_ALGORITHMS,


securesystemslib/keys.py
```diff
@@ -464,7 +464,15 @@ def format_keyval_to_metadata(keytype, scheme, key_value, private=False):
 
   else:
     public_key_value = {'public': key_value['public']}
-
+    # If we encounter a keyid, we are dealing with pub key file generation
+    # as in interface.py#L526
+    if keyid is not None:
```
shibumi, line 469: is this check here is enough? Maybe we want to validate the keyid?

securesystemslib/interface.py
```diff
@@ -516,13 +516,15 @@ def generate_and_write_ed25519_keypair(filepath=None, password=None):
   # to final destination.
   file_object = tempfile.TemporaryFile()
 
-  # Generate the ed25519 public key file contents in metadata format (i.e.,
-  # does not include the keyid portion).
```
shibumi, line 520: Do you know why we never included the keyid here? It doesn't make sense for me, because if we don't provide a filepath we are even naming the keypair files <KEYID>.{pub,key}

shibumi:
mh funny that the coverage increased, I didn't even push tests?!

shibumi:
Ok, I have tested the PR with in-toto with the following requirements.txt file for in-toto:
```
attrs==19.3.0
cffi==1.14.0              # via cryptography, pynacl
colorama==0.4.3           # via securesystemslib
cryptography==2.9.2         # via securesystemslib
enum34==1.1.10             # via cryptography
ipaddress==1.0.23         # via cryptography
iso8601==0.1.12
pathspec==0.8.0
pycparser==2.20           # via cffi
pynacl==1.4.0             # via securesystemslib
python-dateutil==2.8.1
git+git://github.com/shibumi/securesystemslib@shibumi/fix-keyids-in-ed25519-public-keys#egg=securesystemslib
six==1.15.0
subprocess32==3.5.4       # via securesystemslib
```

Then I did:
```
$ python setup.py build
$ python setup.py install
```

Then:
```
(in-toto) chris motoko ~/export/github/in-toto 17:01:51 6cfa36cf shibumi/fix-ed25519-keyid-in-public-key  
❯ cat bob.pub 
{"keytype": "ed25519", "scheme": "ed25519", "keyid": "a1bcfbc3cf506b2a2ac974b679da58a3bde354a4acd23998ba2fdb60c011ba17", "keyid_hash_algorithms": ["sha256", "sha512"], "keyval": {"public": "07dea440b3e151f02a4e80296863f75a0bc0e46f61d7f4e43dd8c441ebb96ce5"}}%                                                             (in-toto) chris motoko ~/export/github/in-toto 17:01:54 6cfa36cf shibumi/fix-ed25519-keyid-in-public-key  
❯ cat bob 
{"keytype": "ed25519", "scheme": "ed25519", "keyid": "a1bcfbc3cf506b2a2ac974b679da58a3bde354a4acd23998ba2fdb60c011ba17", "keyid_hash_algorithms": ["sha256", "sha512"], "keyval": {"public": "07dea440b3e151f02a4e80296863f75a0bc0e46f61d7f4e43dd8c441ebb96ce5", "private": "5aaa95fc5b72bbf1792a8021f7622034d63b8e3099cbe91f577b161fe3042e8c"}}%    
```

As you can see, we do include keyid's now for ed25519. Does `in-toto-keygen` support ecdsa?? Because the help states otherwise..

lukpueh:
Thanks for pointing out this inconsistency and for proposing a fix, @shibumi! Your initiative is highly appreciated, however, I am against changing our custom public key format without a discussion why we came up with this format in the first place, and if the assumptions from back then still hold true.

If we make a potentially breaking change, we might want to consider switching to a standard format altogether (we do use PEM for RSA keys on disk), or at least revise the included fields, e.g. why do ed25519/ecdsa private keys on disk include a keyid field and public keys don't, or why do both not include the `keyid_hash_algorithms` field, etc.... (see https://github.com/theupdateframework/tuf/issues/848 for discussion about the latter field).

Right now, securesystemslib and its dependents do know how to handle their keys. We can load public keys from disk (without keyid), keep them in memory (with keyid) and add them to in-toto layout metadata (with keyid), or in tuf root/targets  metadata (without keyid).

It would be good to make it very clear, when we use which format why. The various `securesystemslib.formats.*KEY_SCHEMA`s, are not particularly helpful for that. Some of them require a keyid and other don't which (unexpectedly) makes it optional (see https://github.com/secure-systems-lab/securesystemslib/issues/183#issuecomment-654154307).

Moreover, the way we currently use the "serialization" function `format_keyval_to_metadata`  looks like a lot of unnecessary work, e.g.:

```python
# simplified and slightly exaggerated for emphasis
a = key["a"]
b = key["b"]
c = key["c"]
metadata = format_keyval_to_metadata(a, b, c)
# metadata is {"a": "...", "b": "...", c: "...", d: "..."}
```

cc @mnm678, @SantiagoTorres, @trishankatdatadog

shibumi:
> Thanks for pointing out this inconsistency and for proposing a fix, @shibumi! Your initiative is highly appreciated, however, I am against changing our custom public key format without a discussion why we came up with this format in the first place, and if the assumptions from back then still hold true.

Don't worry, I am good :D This was more of a quick fix. The discussion alone about this topic is more important. Do we want to keep this discussion in the PR or shall I compile an issue for it?

lukpueh:
An issue would be most welcome!  Feel free to copy-paste whatever I wrote here or on slack. :)

Btw. we already had a related issue, but I think I closed it due to it being partially addressed by our GPG subpackage and RSA key files being standard PEM. Here it is: https://github.com/secure-systems-lab/securesystemslib/issues/55

shibumi:
closing this PR, because discussion should be moved to the related issue.

PR #253, opened 2020-07-09
Moved the metadata into setup.cfg

KOLANICH:
that is provided, the better.

**Fixes issue #**:

1. `setup.cfg` is a declarative format for metadata requiring no remote code execution in order to be parsed.
2. `setuptools_scm` fetches versions from git tags, which decreises new release burden. For non-tag commits it adds pat of a commit hash to a version, which helps bug reporting if anything goes wrong.
3. `pyproject.toml` is a newer declarative format for configuration. In some distant future `setup.cfg` contents will be moved there. Tools recently introduced declarative configuration, like `setuptools_scm`, use it. It also allows to build packages in an unified way: `python3 -m pep517.build -b`

requirements**:


setup.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
```
sechkova, line 1: Unfortunately, I don't think python2 ca be dropped out of the equation yet. I let a maintainer confirm.
Is python2 fully incompatible with this set of changes or the update to py3 comes for a convenience?

setup.py
```diff
@@ -60,56 +60,6 @@
...
 from setuptools import setup
-from setuptools import find_packages
-
-
-with open('README.rst') as file_object:
-  long_description = file_object.read()
-
-setup(
-  name = 'securesystemslib',
-  version = '0.18.0',
```
sechkova, line 71: I tested locally with python 3.7 generating a source dist as it is still listed in the documentation:
` $ python setup.py sdist` 
The resulting archive is missing a version number: securesystemslib-**0.0.0**.  I assume this may be some incompatibility with `setuptools_scm`?

In case I was not supposed to use `setup.py` to generate the sdist, what is the intended way of archive generation after this set of changes? Using the `pep517` module as shown in the PR description did the job but the package [documentation](https://pypi.org/project/pep517/) marks its high-level functionality as deprecated.
Seems like the recommended way is by using [build](https://pypa-build.readthedocs.io/en/latest/).

sechkova (commented):
I support shifting towards newer packaging systems but can you elaborate (maybe update the doc section) what is the intended way of archive generation after this set of changes (see inline comments)?

setup.py
KOLANICH: Try to upgrade setuptools. You need a recent (at least January 2019) setuptools to understand pyproject.toml, and if it understands it, it would upgrade itself to the version having the needed hooks `setuptools_scm` that uses `pyproject.toml` for config depends on.

>Seems like the recommended way is by using `build`.

Yes, it's true. The PR was originally created when there was no `build`.

setup.py
KOLANICH: I don't remember. I remember that 2020 version of setuptools has dropped python 2. I don't use python 2 since 2013.

sechkova: Not everyone is that lucky ...

pyproject.toml
```diff
@@ -0,0 +1,63 @@
...
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Topic :: Security",
+    "Topic :: Software Development",
+]
+requires-python = "~=3.7"
+dependencies = [
+    "six>=1.11.0",
+    'subprocess32; python_version < "3"',
+    "python-dateutil>=2.8.0",
+]
```
lukpueh, line 42:
```suggestion
```
We don't have any of these dependencies anymore.

pyproject.toml
```diff
@@ -0,0 +1,63 @@
...
+
+[project.urls]
+Homepage = "https://github.com/secure-systems-lab/securesystemslib"
+Source = "https://github.com/secure-systems-lab/securesystemslib"
+Issues = "https://github.com/secure-systems-lab/securesystemslib/issues"
+
+[project.optional-dependencies]
+crypto = ["cryptography>=37.0.0"]
+pynacl = ["pynacl>1.2.0"]
+testing = ['mock; python_version < "3.3"']
```
lukpueh, line 53:
```suggestion
```
Don't need this one anymore either.

pyproject.toml
```diff
@@ -0,0 +1,63 @@
+[build-system]
+requires = ["setuptools>=61.2", "setuptools_scm[toml]>=3.4.3"]
+build-backend = "setuptools.build_meta"
```
lukpueh, line 3: I don't think we need the features of `setuptools_scm`, and I'd generally like us to switch to `hatchling` as we did in python-tuf and in-toto (https://github.com/theupdateframework/python-tuf/pull/1896, https://github.com/in-toto/in-toto/pull/490).

pyproject.toml
```diff
@@ -0,0 +1,63 @@
...
+Issues = "https://github.com/secure-systems-lab/securesystemslib/issues"
+
+[project.optional-dependencies]
+crypto = ["cryptography>=37.0.0"]
+pynacl = ["pynacl>1.2.0"]
+testing = ['mock; python_version < "3.3"']
+PySPX = ["PySPX==0.5.0"]
+
+[tool.setuptools]
+include-package-data = false
```
lukpueh, line 57: How does this relate to MANIFEST.in? I think we want to include some non-*.py files in source distributions.

lukpueh (changes requested):
Thanks for the persistence and sorry for the late review, @KOLANICH!

Let me know if you feel like switching to `hatchling` in this PR. I'm also fine if we do that in a follow-up PR. I wouldn't add `setuptools_scm`, though, as we didn't use it before and won't use in the future.

pyproject.toml
lukpueh: Oh right, I forgot that we don't have the version in `__init__.py`. I'd rather add it there and not use git to infer it. But I'm fine with doing that in a follow up PR.

KOLANICH:
>Let me know if you feel like switching to hatchling in this PR. I'm also fine if we do that in a follow-up PR.

I'd prefer it to happen in a separate PR/commit. Switching to `PEP 621` without changing the build backend is one change. Switching to `hatchling` is another one. (Though former is a prereq of the latter).

>I wouldn't add setuptools_scm, though, as we didn't use it before and won't use in the future.

As `jku` has noticed, it is used to eliminate the need to populate the version manually into multiple places (git history, source code file, metadata file). Also it places git commit hash into version number of non-releaseversions, which is helpful for debugging if anything goes wrong. Hatchling has a similar plugin `hatch-vcs` using `setuptools_scm` under the hood.

>I'd rather add it there and not use git to infer it. But I'm fine with doing that in a follow up PR.

It may be possible to make `setuptools` to infer it from the source code. Not sure about `hatchling` though, I have too little experience with it.

>How does this relate to MANIFEST.in? I think we want to include some non-*.py files in source distributions.

When it is `true` it uses it to place the files into the wheel. It is not documented, but [the docs](https://setuptools.pypa.io/en/latest/userguide/datafiles.html) make an impression this setting is not about sdists.

lukpueh:
> > Let me know if you feel like switching to hatchling in this PR. I'm also fine if we do that in a follow-up PR.
> 
> I'd prefer it to happen in a separate PR/commit. Switching to `PEP 621` without changing the build backend is one change. Switching to `hatchling` is another one. (Though former is a prereq of the latter).

Agreed. Let's switch to hatchling in a follow-up PR.

> 
> > I wouldn't add setuptools_scm, though, as we didn't use it before and won't use in the future.
> 
> As `jku` has noticed, it is used to eliminate the need to populate the version manually into multiple places (git history, source code file, metadata file). Also it places git commit hash into version number of non-releaseversions, which is helpful for debugging if anything goes wrong. Hatchling has a similar plugin `hatch-vcs` using `setuptools_scm` under the hood.

Okay, let's use `setuptools_scm` then.

> 
> > I'd rather add it there and not use git to infer it. But I'm fine with doing that in a follow up PR.
> 
> It may be possible to make `setuptools` to infer it from the source code. Not sure about `hatchling` though, I have too little experience with it.

Don't know how to do it in `setuptools`, but you can [tell `hatchling`, where to find the version in source](https://github.com/theupdateframework/python-tuf/blob/a4d5cf49c9eaddf26813d1b89db4a86b90c34a28/pyproject.toml#L57-L58).

> 
> > How does this relate to MANIFEST.in? I think we want to include some non-*.py files in source distributions.
> 
> When it is `true` it uses it to place the files into the wheel. It is not documented, but [the docs](https://setuptools.pypa.io/en/latest/userguide/datafiles.html) make an impression this setting is not about sdists.

Thanks for the docs link. I suggest to either set it to true then or replicate the includes/excludes from MANIFEST.in here and remove MANIFEST.in.

lukpueh:
I see you force-pushed but did not address my review comments. Are you interested in finalizing this PR, @KOLANICH?

lukpueh (approved):
LGTM! I tried this and it works as expected.

Also checked that it includes everything we want it to include in an sdist via MANIFEST.in. Interestingly, it also includes at least one of the files we ignore via the manifest.

No need to fix this here, as we need to change the configuration when switching to hatchling anyways.

Thanks again for the contribution, @KOLANICH!

PR #254, opened 2020-07-09
Replaced absolute paths to the components with relative ones.

KOLANICH:
that is provided, the better.

Fixes: #254

Importing stuff from an own package by its absolute name is harmful. It causes problems with forks using other names. So I have replaced absolute imports with relative ones.

requirements**:

### DIFF
--- a/securesystemslib/ecdsa_keys.py
+++ b/securesystemslib/ecdsa_keys.py
@@ -67,8 +67,8 @@ except ImportError:
   CRYPTO = False
 
 # Perform object format-checking and add ability to handle/raise exceptions.
-import securesystemslib.formats
-import securesystemslib.exceptions
+from . import formats
+from . import exceptions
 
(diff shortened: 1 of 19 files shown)


joshuagl:
Whilst I appreciate the suggestion, I'm not convinced this is a change we want to take in securesystemslib. [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports) recommends against this style of import unless necessary, and I'm not convinced it is necessary. 

Are you aware of a fork for which this import style is causing problems? Can we encourage them to work with us, rather than forking?

KOLANICH:
I don't know any fork it really causes problems to (I haven't searched for any forks), but using absolute imports are making issues from thin air. It not only creates problems for forks, it also creates problems for us. Code using absolute imports is harder to refactor.

joshuagl:
Thank you for the submission! Whilst we are very grateful for the contribution we have decided not to merge this PR. We like following the ecosystem norms as defined in PEP 8.

Issue #246, opened 2020-06-24
Split up util.get_file_details()


joshuagl:
**Description of issue or feature request**:

`util.get_file_details()` returns the length and hashes of the specified file. This makes a lot of sense for generating information about TUF targets, where the length and hashes of each target are required, but is potentially too costly a function when generating information about TUF metadata files for the snapshot role, where lengths and hashes are optional. Snapshot metadata may only want hashes, or only want lengths, and we don't want to force an adopter to implement their own subset of the functionality in `get_file_details()`.

Therefore I suggest we factor out the two features of `get_file_details()` into two new utility functions – one for length and one for hashes – and have the existing `get_file_details()` wrap them.

**Current behavior**:

Callers wanting to compute length or hashes, not both, need to either accept the overhead of unwanted computations in `get_file_details()` (only a significant problem for hashes), or implement their own equivalent functionality.

**Expected behavior**:

A separate API is provided for at least hash computation, ideally hash and length computation, with `get_file_details()` conveniently wrapping both.
`get_file_details(filepath, hash_algorithms=['sha256'], storage_backend=None)`
`get_file_length(file_path, storage_backend=None)`
`get_file_hashes(file_path, hash_algorithms=['sha256'], storage_backend=None)`

joshuagl:
@lukpueh are you aware of `get_file_details()` being used outside of the tuf reference implementation? I don't see it used in a quick `git grep` of in-toto.

Maybe this would be better as two utility functions in securesystemslib and a convenience wrapper in tuf?

lukpueh:
I know that it is not used in in-toto. We [hash files directly with `securesystemslib.hash`](https://github.com/in-toto/in-toto/blob/6f7f5cc8d2955c511f6774df178ed376798815bc/in_toto/runlib.py#L68). A quick grep through sslib's  [dependents](https://github.com/secure-systems-lab/securesystemslib/network/dependents) does not reveal any usage either. I'm fine with moving the wrapper to tuf. :)

joshuagl:
Resolved in #259

PR #259, opened 2020-07-17
Split up util.get_file_details()

MVrachev:
that is provided, the better.

**Fixes issue #**: https://github.com/secure-systems-lab/securesystemslib/issues/246

The `util.get_file_details()` function returns the length and hashes
of the specified file. 
As said on issue #246 on first glance it seems that only TUF uses 
`util.get_file_details()` and if that's the case then for TUF it will
be better if we split that function

In the current format the function `util.get_file_details()` 
makes a lot of sense for generating information about TUF targets, 
where the length and hashes of each target are required but this is 
potentially too costly function when generating information about TUF 
metadata files for the snapshot role, where lengths and hashes are optional.
Snapshot metadata may only want hashes, or only want lengths,
and we don't want to force an adopter to implement their own subset
of the functionality in get_file_details().

requirements**:

### DIFF
--- a/securesystemslib/util.py
+++ b/securesystemslib/util.py
@@ -81,13 +81,53 @@ def get_file_details(filepath, hash_algorithms=['sha256'],
   if storage_backend is None:
     storage_backend = securesystemslib.storage.FilesystemBackend()
 
-  # The returned file hashes of 'filepath'.
+  file_length = get_file_length(filepath, storage_backend)
+  file_hashes = get_file_hashes(filepath, hash_algorithms, storage_backend)
+
+  return file_length, file_hashes
+
+
+def get_file_hashes(filepath, hash_algorithms=['sha256'],
+    storage_backend=None):
+  """
+  <Purpose>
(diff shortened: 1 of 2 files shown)


securesystemslib/util.py
```diff
@@ -81,13 +81,52 @@ def get_file_details(filepath, hash_algorithms=['sha256'],
...
+
+  return file_length, file_hashes
+
+
+def get_file_hashes(filepath, hash_algorithms=['sha256'],
+    storage_backend=None):
+  """
+  <Purpose>
+    To get file's hash/hashes information.  The hash is computed using the
+    sha256 algorithm by default.
```
joshuagl, line 95:
```suggestion
    Compute hash(es) of the file at filepath using each of the specified
    hashing algorithms. If no algorithms are specified, the hash is
    computed using the SHA-256 algorithm.
```

MVrachev:
I agree with all of the docstrings suggestions and updated the commit by addressing them.

Issue #270, opened 2020-09-11
Revise module architecture and define clear API modules and functions


lukpueh:
**Description of issue or feature request**:
Securesystemslib lacks of a clear public API. It should be clear and intuitive for users of secureystemslib, which modules and which functions are public API.

**Current behavior**:
API is scattered across:

- for *general* key operations (generate, import, sign, verify)
  - `keys.py` -- high-level public interface to sign/very (key type independent), and generate and import keys (key type dependent). Calls into low-level non-public `{rsa, ecdsa, ed25519}_keys.py` modules, which are (mostly) separated by key type.
  - `interface.py` -- higher-level public interface (mostly calls into `key.py`) to generate and import keys

- for GPG key operations (import, sign, verify)
  - `gpg/functions.py` -- public interface for gpg subpackage, independent from above key operations.

- for other other non-key related operations
  - `hash.py` -- facade for `hashlib` from stdlib and `cryptography.hazmat.primitives.hashes`
  - `process.py` -- thin subprocess wrapper
  - `storage.py`-- file system abstraction 
  - `util.py` -- mostly I/O related utils
  - `formats.py`-- schema definition constants (likely to be deprecated, see https://github.com/secure-systems-lab/securesystemslib/issues/183), OLPC canonical json implementation 

 
**Expected behavior**:
Revise module architecture to use mnemonic names for (public) modules (not `interface` or `functions`) appropriate for the interface functions they contain. Also see discussion about [import guidelines](https://github.com/secure-systems-lab/code-style-guidelines/issues/20#issuecomment-679225442).

trishankatdatadog:
💯 💯 💯

MVrachev:
Overall the suggestions sound good. 

On the topic of `RSASigner` example (as @lukpueh has already heard my opinion :D) I am not a fan of adding an inheritance just for the sake of simpler implementation without logical connection.
In this example, I don't see a logical relationship between `RSASigner` and `RSAPrivateKey`.
This of course is just my opinion and the maintainers should decide. 

One minor naming suggestion - why don't we use `SSlibSigner` instead of `SslibSigner`?
That's for all naming suggestions, I think double upper case `S` looks better.

lukpueh:
> One minor naming suggestion - why don't we use SSlibSigner instead of SslibSigner?
That's for all naming suggestions, I think double upper case S looks better.

Agreed, the class in securesystemslib is actually called `SSlibSigner`.  I'll update comments above...

[...]

PR #278, opened 2020-09-30
Add interface.export_pubkeys convenience function

lukpueh:
Fixes: # -
Related to in-toto/in-toto#80

Add convenience function to import multiple public keys of different key types from file into an sslib dict format at once, and tests.

Note: Uses the new pseudo-standard docstring style suggested in secure-systems-lab/code-style-guidelines#20. All new interface functions should use that style (existing docstrings will be converted in separate PRs).

Note2: This PR also adds a few package-wide constants for key types, that are still hardcoded in many places.  We should switch to using these and probably other global constants in a separate PR.

### DIFF
--- a/securesystemslib/__init__.py
+++ b/securesystemslib/__init__.py
@@ -9,3 +9,10 @@ import logging
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.WARNING)
 logger.addHandler(logging.StreamHandler())
+
+
+# Global constants
+# TODO: Replace hard-coded key types with these constants (and add more)
+KEY_TYPE_RSA = "rsa"
+KEY_TYPE_ED25519 = "ed25519"
+KEY_TYPE_ECDSA = "ecdsa"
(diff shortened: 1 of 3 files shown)


securesystemslib/__init__.py
```diff
@@ -9,3 +9,10 @@
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.WARNING)
 logger.addHandler(logging.StreamHandler())
+
+
+# Global constants
+# TODO: Replace hard-coded key types with these constants (and add more)
```
joshuagl, line 15: Would these constants be better suited to the keys module? Is it worth documenting how we're choosing where to put constants?

joshuagl (approved):
I have a question about where we are defining constants, but otherwise this looks like a nice addition. Thanks @lukpueh !

securesystemslib/__init__.py
lukpueh: IMO `__init__` is a good default place for global constants. If we want to module-scope them, we should coordinate with #270.

lukpueh:
Thanks for the quick review, @joshuagl!

PR #279, opened 2020-10-01
Refactor interface tests and add more

lukpueh:
Fixes: # -

Note, this PR does not change the interface behavior, it only cleans up the test module, with the goal to make future interface changes more easily comprehensible by looking at changes in the test module. 

The clean up includes:

- Remove repetitive test dir creation in individual test functions.
  Use setUp function instead
- Remove redundant test dir creation in setUpClass function.
- Remove repetitive key generation and key import code used to test
  one or the other "unit".
  Instead aggregate key generation and loading testing in a single
  test function per key type (rsa, ed25519, ecdsa).
- Use context manager for assertRaises to increase readability
- Use test tables (for loops) for more extensive testing of
  different combinations of args and kwargs.

This PR also includes additional testing:
- interactive password prompts tests (rendering coverage exemption
  pragmas obsolete).
- import regression tests for ed25519 and ecdsa keys, using
  pre-generated keys instead of on-the-fly generated keys.
- stronger assertions, e.g. actually check the key length to
  test the custom bits parameter for RSA key creation, or look at
  error message contents
- more error testing, e.g. by trying to load invalid keys, or keys

This PR also identifies and adds tests and FIXME comments for unexpected interface behavior, identified while updating the tests, i.e.
- import_ed25519_publickey_from_file can import private keys
- import_ed25519_privatekey_from_file can import public keys
- import_ecdsa_publickey_from_file can import ed25519 public keys


SantiagoTorres:
Wow, we lost all our coverage. I assume this is a misconfig or a bug...

trishankatdatadog:
Coverage is honestly one of the well-intentioned but mostly-misguided things I've seen in testing...

lukpueh:
> Wow, we lost all our coverage. I assume this is a misconfig or a bug...

I think that was caused by a force-push interrupting a build.

> Coverage is honestly one of the well-intentioned but mostly-misguided things I've seen in testing...

I think it's not that bad. :) But I agree that one should not overrate coverage results, especially if developers are very liberal with coverage exemptions, some of which I'm removing in this PR.

Btw. the build is still failing because of coverage dropping below 99% on Python 3.5. Looking at the results, it seems like it's only a rounding difference, because the absolute coverage is the same as on other versions. Anyways, I think 99% is a rather strict threshold. If you don't mind I'll lower it a bit.

joshuagl (approved):
Looks like a solid set of enhancements to me, thanks Lukas!

Could you file the FIXME comments as issues?

lukpueh:
Thanks for the review, @joshuagl!

> Could you file the FIXME comments as issues?

Done in #280.

Merging...

Issue #282, opened 2020-10-08
blake2b digest length is incompatible with Warehouse


jku:
**I believe the blake2b hashes produced by warehouse and securesystemslib  are incompatible**

I can't actually run the full update cycle with my pip-branch against the Warehouse tuf-branch yet but I've taken another look at the metadata and I'm fairly sure we will be incompatible in the hash verification step.

This is what Warehouse target metadata currently looks like: 
```
"0b/de/ae38ae748ae74093a0c8abad03d83e179870639d2d1b092ba4b3e82164df/litefeel_pycommon-0.1.1-py3-none-any.whl": {
    "custom": {
     "backsigned": true
    },
    "hashes": {
     "blake2b": "0bdeae38ae748ae74093a0c8abad03d83e179870639d2d1b092ba4b3e82164df"
    },
    "length": 3077
   }
```

A TUF client compares that "blake2b" hash to one that securesystemslib produces for the given data, something like this simplified example:
```
>>> dobj = securesystemslib.hash.digest("blake2b")
>>> dobj.update(b'')
>>> dobj.hexdigest()
'786a02f742015903c6c6fd852552d272912f4740e15847618a86e217f71f5419d25e1031afee585313896444934eb04b903a685b1448b755d56f701afe9be2ce'
```

The digests have different lengths. Warehouse uses blake2b-256 and TUF uses blake2b-512 (which is the maximum).

So:
* First, is the warehouse choice set in stone? I can see how it kind of might be as the shorter hash is already used in many places (ping @woodruffw)
* If it is, we probably want to support at least a new "blake2b-256" hash name (and add the little bit of new code needed in `hash.py::digest()`)

I can do the PR if there's consensus.

lukpueh:
Thanks for catching this. I'm fine either way. In case you haven't looked it up yet, in hashlib we get `"blake2b-256"` with `hashlib.blake2b(digest_size=32)` or  `hashlib.new("blake2b", digest_size=32)`.`pyca/cryptography`'s hash interface doesn't seem to support blake2b-256.

woodruffw:
To answer my part: yes, the current Warehouse usage is (more or less) set in stone: it's part of the DB schema and is already present in a number of other artifacts of the Warehouse packaging pipeline.

PR #283, opened 2020-10-08
Add blake2b-256 algorithm

jku:
Fixes: #282 

Add 'blake2b-256' as a new supported algorithm: it is really the same algorithm as blake2b, just half the digest size.

Refactors hash tests a little to ensure that blake2* gets tested but that they do not get tested on pyca_crypto (which does not support them).

I've manually verified that we now get same results as Warehouse :)

### DIFF
--- a/tests/test_hash.py
+++ b/tests/test_hash.py
@@ -47,231 +47,156 @@ if not 'hashlib' in securesystemslib.hash.SUPPORTED_LIBRARIES:
 class TestHash(unittest.TestCase):
 
   @staticmethod
-  def _get_algorithms(library):
-    algorithms = ['md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512']
-    if library in TestHash._get_blake_libraries():
-      algorithms += ['blake2b', 'blake2b-256', 'blake2s']
-    return algorithms
+  def _is_supported_combination(library, algorithm):
+    blake_algos = ['blake2b', 'blake2b-256', 'blake2s']
+
+    # pyca does not support blake2*
+    if algorithm in blake_algos:
+      if library == 'pyca_crypto':
+        return False
+      # hashlib does not support blake2* if < 3.6
+      elif library == 'hashlib' and sys.version_info[:2] < (3, 6):
+        return False
+    return True
 
 
-  @staticmethod
-  def _run_with_hash_libraries(test_func, libraries=None):
-    if libraries is None:
-      libraries = securesystemslib.hash.SUPPORTED_LIBRARIES
-    for lib in libraries:
-      test_func(lib)
+  def _run_with_all_algos_and_libs(self, test_func):
+    algorithms = [
+      'md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512',
+      'blake2b-256', 'blake2b', 'blake2s',
+    ]
+    for algorithm in algorithms:
+      self._run_with_all_hash_libraries(test_func, algorithm)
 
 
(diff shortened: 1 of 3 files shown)


tests/test_hash.py
```diff
@@ -45,13 +45,89 @@
 
 class TestHash(unittest.TestCase):
 
-  def _run_with_all_hash_libraries(self, test_func):
-    for lib in securesystemslib.hash.SUPPORTED_LIBRARIES:
+  @staticmethod
+  def _get_algorithms(library):
+    algorithms = ['md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512']
+    if library != 'pyca_crypto':
```
trishankatdatadog, line 51: I'm not sure if this should be so dynamic between every function call. Also, it assumes that the library has been/not been loaded/unloaded.

trishankatdatadog (approved):
LGTM, I can't think of weaknesses. Thanks!

jku:
Huh, blake2* is only supported in hashlib from python3.6 apparently

jku:
I think test rig looks ok now: it should test every algorithm if the library and python version allow.
What I don't like is that it's not very easy to see that he correct algorithms did get tested -- but I think it's better than before.

joshuagl (commented):
Should we be throwing some kind of error if a caller is trying to generate blake2* digests with Python < 3.6 or pyca?

jku:
It does throw an error... and yes: if the tests checked for that, they would be much easier to reason about

jku:
I plan to squash the commits into two: but did not do it yet in case Joshua wants to see diff vs the reviewed version.

Notes for reviewers:
* ended up refactoring the `_do_{algorithm}_update()` functions into a single function
* the supported algorithms for a test are still not listed in every test but now it's pretty easy to reason and verify what is expected to fail and what is expected to succeed: `_run_with_all_hash_libraries()` decides that and can easily be made to print out those decisions if they need to be checked
* found an inconsistency in my new code: see the TypeError handling in hash.py (yay for testing)

joshuagl (approved):
This looks much better, thanks @jku! 

I wonder if it would be clearer or not if we were using the [`unittest.skip*` decorators](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) to explicitly skip test cases where we don't expect them to work. That might lead to fewer unexpected errors than only accepting failures if it's an `UnsupportedAlgorithmError`, though does require explicit tests for the `UnsupportedAlgorithmError` on configurations which don't support it.

Let me know what you think.

jku:
> I wonder if it would be clearer or not if we were using the [`unittest.skip*` decorators](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) to explicitly skip test cases where we don't expect them to work.

I like the fact that we now test every combination and actually verify the correct Error so skip doesn't sound right. A `@ExpectedFailureIf(condition, error, reason)` might be useful but that does not exist out of the box.

Even with that, I don't think we can make it work since the tests are not "unique": a single test case tests all libraries so we can't condition the skip/expectedfailure it on the library argument. Parameterizing 'library' (and maybe 'algorithm') would enable that but that means an additional test dependency on something like https://github.com/wolever/parameterized -- I think that's too much.

TL;DR: we should keep parameterizing in mind as an option in future, but for now the decorations are IMO not helpful

joshuagl:
Thanks for taking the time to reason through that and discuss with me! Merging as is with a plan to look at parameterising tests in future.

Issue #251, opened 2020-07-06
key format inconsistency


shibumi:
**Description of issue or feature request**:

Based on this PR (https://github.com/secure-systems-lab/securesystemslib/pull/250) we had a discussion around a default or custom key format. The PR showed up a few inconsistencies regarding public key generation in `in-toto-keygen`. We generate a key id for private keys, but no key id will be written for public keys. This lead to the question if we want to support key ids in our key format.

**Current behavior**:

Right now TUF and in-toto behave differently regarding key formats:

- Neither [in-toto](https://github.com/in-toto/docs/blame/92483b5b31e34d059445ba132d76d476e6879a11/in-toto-spec.md#L497-L502) nor [TUF](https://github.com/theupdateframework/specification/blame/9794b922b5f6ac615854b308862327d589935479/tuf-spec.md#L508-L513) specs mention the keyid field when defining key object formats.
- Neither [in-toto](https://github.com/in-toto/docs/blame/92483b5b31e34d059445ba132d76d476e6879a11/in-toto-spec.md#L583) nor [TUF](https://github.com/theupdateframework/specification/blame/9794b922b5f6ac615854b308862327d589935479/tuf-spec.md#L659-L665) specs show the keyid field in metadata examples.
- The TUF reference implementation does not include the keyid field in public key objects in [root](https://github.com/theupdateframework/tuf/blob/de8649e12a850fccbc745f0129d43956ef7da33b/tests/repository_data/repository/metadata/root.json#L13-L23) or [targets](https://github.com/theupdateframework/tuf/blob/de8649e12a850fccbc745f0129d43956ef7da33b/tests/repository_data/repository/metadata/targets.json#L12) metadata on disk.
- The in-toto reference implementation does include the keyid field in public key objects in [layout](https://github.com/in-toto/in-toto/blob/6cfa36cfc2c4b864936560294c032d3bfa869064/tests/demo_files/demo.layout.template#L37-L47) metadata on disk.
- The in-toto reference implementation uses schema checks on key public key objects (held in memory) that require the keyid, and checks that allow keyid:
[GPG_PUBKEY_SCHEMA](https://github.com/secure-systems-lab/securesystemslib/blob/22a3c4e252c60e911355feaa204f70c2c191aae2/securesystemslib/formats.py#L384-L385) -- requires keyid
[...]

PR #288, opened 2020-10-16
Consolidate key interface password handling and general overhaul

lukpueh:
Fixes: #122
Follows up on: #124 (see for interface discussion) and #148 
Partially fixes in-toto/in-toto#80

- Adds and adopts helper functions to aggregate repetitive code that handles encryption- (in key pair generation), and decryption- (in private key import) passwords, in order to make the behavior consistent for all key types.
- Cleans up the code, comments and docstrings and use Google style formatting for the latter to nicely render on readthedocs. A demo is available at: https://in-toto-lukpueh.readthedocs.io/en/latest/api.html#utilities
- Adds generic private key import interface functions, to import any of the supported private key types from file (rsa, ecdsa, ed25519).
- Changes the behavior of  `generate_and_write_*_keypair` and adds two additional functions (see below).

**Behavior changes**:
- Adds support to optionally not encrypt (on generation) and decrypt (on import) ecdsa private keys to be consistent with rsa and ecdsa key interface functions.
- ~Adds optional `prompt` kwarg to all key generation functions.~
- Adds optional `prompt` kwarg to ecdsa private key import function (rsa and ed25519 already had it).
- Changes handling of `None` or `""` passwords passed to key generation or private key import functions.
  - ~**MOST IMPORTANTLY:** generation function no longer open a prompt if no password is passed as argument, i.e. keys are written to disk **unencrypted by default**. This may be revisited in the future (see https://github.com/secure-systems-lab/securesystemslib/pull/288#discussion_r511420534 for rationale/discussion).~ 
  - `generate_and_write_*_keypair` functions now require a positional `password` argument as **first** argument, which must be a non-empty string. The functions no longer open a prompt or write unencryted passwords. For that two new interface functions are added per key type,  `generate_and_write_*_keypair_with_prompt` and `generate_and_write_unencrypted_*_keypair`. This follows the principles of secure defaults and least surprise.

**Pain points**:
- Rsa functions use standardized PEM, and ed25519 and ecdsa functions use a custom json format for keys on disk, as a consequence the signing scheme for rsa keys must be assigned on import, and for ed25519 and ecdsa keys on generation.
- Ed25519 and ecdsa key generation  functions in the interface don't take a scheme parameter, as a consequence only the the default scheme can be used.
- The newly added generic key import function does not take a scheme parameter and thus only the default signing scheme can be used. This is a deliberate decision in this PR due to the existing scheme assignment inconsistency described above.
- The newly added generic key import function requires a `key_type` parameter to dispatch to the appropriate function. It would be nice if the key type could be assessed based on the on-disk format, but this does not seem feasible given the different formats as described above.
- Minor: The used sphinx napoleon plugin to render docstrings as html does not handle the custom "Side Effects" section well (see demo above and https://github.com/sphinx-contrib/napoleon/issues/2).

**Note to reviewer**
I advise to review commit by commit, or at least to read the commit messages for details

### DIFF
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -662,7 +662,8 @@ def import_ed25519_privatekey_from_file(filepath, password=None, prompt=False,
 
 
 
-def generate_and_write_ecdsa_keypair(filepath=None, password=None):
+def generate_and_write_ecdsa_keypair(filepath=None, password=None,
+    prompt=False):
   """
   <Purpose>
     Generate an ECDSA keypair, where the encrypted key (using 'password' as the
@@ -701,6 +702,8 @@ def generate_and_write_ecdsa_keypair(filepath=None, password=None):
     The 'filepath' of the written key.
   """
 
+  password = _get_key_file_encryption_password(password, prompt, filepath)
+
   # Generate a new ECDSA key object.  The 'cryptography' library is currently
   # supported and performs the actual cryptographic operations.
   ecdsa_key = securesystemslib.keys.generate_ecdsa_key()
@@ -715,23 +718,6 @@ def generate_and_write_ecdsa_keypair(filepath=None, password=None):
   # Does 'filepath' have the correct format?
   # Raise 'securesystemslib.exceptions.FormatError' if there is a mismatch.
   securesystemslib.formats.PATH_SCHEMA.check_match(filepath)
-
-  # If the caller does not provide a password argument, prompt for one.
-  if password is None:
-
-    # It is safe to specify the full path of 'filepath' in the prompt and not
-    # worry about leaking sensitive information about the key's location.
-    # However, care should be taken when including the full path in exceptions
-    # and log files.
-    password = get_password('Enter a password for the ECDSA'
-        ' key (' + TERM_RED + filepath + TERM_RESET + '): ',
-        confirm=True)
-
-  else:
-    logger.debug('The password has been specified.  Not prompting for one')
-
-  # Does 'password' have the correct format?
-  securesystemslib.formats.PASSWORD_SCHEMA.check_match(password)
 
   # If the parent directory of filepath does not exist,
   # create it (and all its parent directories, if necessary).
@@ -759,10 +745,16 @@ def generate_and_write_ecdsa_keypair(filepath=None, password=None):
   # Write the encrypted key string, conformant to
   # 'securesystemslib.formats.ENCRYPTEDKEY_SCHEMA', to '<filepath>'.
   file_object = tempfile.TemporaryFile()
+
+  if password is not None:
+    ecdsa_key = securesystemslib.keys.encrypt_key(ecdsa_key, password)
+
+  else:
+    ecdsa_key = json.dumps(ecdsa_key)
+
   # Raise 'securesystemslib.exceptions.CryptoError' if 'ecdsa_key' cannot be
   # encrypted.
-  encrypted_key = securesystemslib.keys.encrypt_key(ecdsa_key, password)
-  file_object.write(encrypted_key.encode('utf-8'))
+  file_object.write(ecdsa_key.encode('utf-8'))
   securesystemslib.util.persist_temp_file(file_object, filepath)
 
   return filepath
@@ -814,7 +806,7 @@ def import_ecdsa_publickey_from_file(filepath):
 
 
 
-def import_ecdsa_privatekey_from_file(filepath, password=None,
+def import_ecdsa_privatekey_from_file(filepath, password=None, prompt=False,
     storage_backend=None):
   """
   <Purpose>
@@ -858,37 +850,25 @@ def import_ecdsa_privatekey_from_file(filepath, password=None,
   # Raise 'securesystemslib.exceptions.FormatError' if there is a mismatch.
   securesystemslib.formats.PATH_SCHEMA.check_match(filepath)
 
-  # If the caller does not provide a password argument, prompt for one.
-  # Password confirmation disabled here, which should ideally happen only
-  # when creating encrypted key files (i.e., improve usability).
-  if password is None:
-
-    # It is safe to specify the full path of 'filepath' in the prompt and not
-    # worry about leaking sensitive information about the key's location.
-    # However, care should be taken when including the full path in exceptions
-    # and log files.
-    password = get_password('Enter a password for the encrypted ECDSA'
-        ' key (' + TERM_RED + filepath + TERM_RESET + '): ',
-        confirm=False)
-
-  # Does 'password' have the correct format?
-  securesystemslib.formats.PASSWORD_SCHEMA.check_match(password)
+  password = _get_key_file_decryption_password(password, prompt, filepath)
 
   if storage_backend is None:
     storage_backend = securesystemslib.storage.FilesystemBackend()
 
   # Store the encrypted contents of 'filepath' prior to calling the decryption
   # routine.
-  encrypted_key = None
-
   with storage_backend.get(filepath) as file_object:
-    encrypted_key = file_object.read()
+    key_data = file_object.read().decode('utf-8')
 
   # Decrypt the loaded key file, calling the 'cryptography' library to generate
   # the derived encryption key from 'password'.  Raise
   # 'securesystemslib.exceptions.CryptoError' if the decryption fails.
-  key_object = securesystemslib.keys.decrypt_key(encrypted_key.decode('utf-8'),
-      password)
+  if password is not None:
+    key_object = securesystemslib.keys.decrypt_key(key_data, password)
+
+  else:
+    key_object = securesystemslib.util.load_json_string(key_data)
+
 
   # Raise an exception if an unexpected key type is imported.
   # NOTE: we support keytype's of ecdsa-sha2-nistp256 and ecdsa-sha2-nistp384
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -120,6 +120,41 @@ def get_password(prompt='Password: ', confirm=False):
 
 
 
+def _get_key_file_encryption_password(password, prompt, path):
+  """Encryption password helper.
+
+  - Fail if 'password' is passed and 'prompt' is True (precedence unclear)
+  - Fail if empty 'password' arg is passed (encryption desire unclear)
+  - Return None on empty pw on prompt (suggests desire to not encrypt)
+
+  """
+  securesystemslib.formats.BOOLEAN_SCHEMA.check_match(prompt)
+
+  # We don't want to decide which takes precedence so we fail
+  if password is not None and prompt:
+    raise ValueError("passing 'password' and 'prompt=True' is not allowed")
+
+  # Prompt user for password and confirmation
+  if prompt:
+    password = get_password("enter password to encrypt private key file "
+        "'" + TERM_RED + str(path) + TERM_RESET + "' (leave empty if key "
+        "should not be encrypted): '", confirm=True)
+
+    # Treat empty password as no password. A user on the prompt can only
+    # indicate the desire to not encrypt by entering no password.
+    if not len(password):
+      return None
+
+  if password is not None:
+    securesystemslib.formats.PASSWORD_SCHEMA.check_match(password)
+
+    # Fail on empty passed password. A caller should pass None to indicate the
+    # desire to not encrypt.
+    if not len(password):
+      raise ValueError("encryption password must be 1 or more characters long")
+
+  return password
+
 
 
 def _get_key_file_decryption_password(password, prompt, path):
@@ -155,7 +190,7 @@ def _get_key_file_decryption_password(password, prompt, path):
 
 
 def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
-    password=None):
+    password=None, prompt=False):
   """
   <Purpose>
     Generate an RSA key pair.  The public portion of the generated RSA key is
@@ -198,6 +233,7 @@ def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
   # Does 'bits' have the correct format?
   # Raise 'securesystemslib.exceptions.FormatError' if there is a mismatch.
   securesystemslib.formats.RSAKEYBITS_SCHEMA.check_match(bits)
+  password = _get_key_file_encryption_password(password, prompt, filepath)
 
   # Generate the public and private RSA keys.
   rsa_key = securesystemslib.keys.generate_rsa_key(bits)
@@ -214,25 +250,9 @@ def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
   # Does 'filepath' have the correct format?
   securesystemslib.formats.PATH_SCHEMA.check_match(filepath)
 
-  # If the caller does not provide a password argument, prompt for one.
-  if password is None:
-
-    # It is safe to specify the full path of 'filepath' in the prompt and not
-    # worry about leaking sensitive information about the key's location.
-    # However, care should be taken when including the full path in exceptions
-    # and log files.
-    password = get_password('Enter a password for the encrypted RSA'
-        ' key (' + TERM_RED + filepath + TERM_RESET + '): ',
-        confirm=True)
-
-  else:
-    logger.debug('The password has been specified.  Not prompting for one')
-
-  # Does 'password' have the correct format?
-  securesystemslib.formats.PASSWORD_SCHEMA.check_match(password)
 
   # Encrypt the private key if 'password' is set.
-  if len(password):
+  if password is not None:
     private = securesystemslib.keys.create_rsa_encrypted_pem(private, password)
 
   else:
@@ -421,7 +441,8 @@ def import_rsa_publickey_from_file(filepath, scheme='rsassa-pss-sha256',
 
 
 
-def generate_and_write_ed25519_keypair(filepath=None, password=None):
+def generate_and_write_ed25519_keypair(filepath=None, password=None,
+    prompt=False):
   """
   <Purpose>
     Generate an Ed25519 keypair, where the encrypted key (using 'password' as
@@ -460,7 +481,8 @@ def generate_and_write_ed25519_keypair(filepath=None, password=None):
     The 'filepath' of the written key.
   """
 
-  # Generate a new Ed25519 key object.
+  password = _get_key_file_encryption_password(password, prompt, filepath)
+
   ed25519_key = securesystemslib.keys.generate_ed25519_key()
 
   if not filepath:
@@ -475,23 +497,6 @@ def generate_and_write_ed25519_keypair(filepath=None, password=None):
   # types, and that all dict keys are properly named.
   # Raise 'securesystemslib.exceptions.FormatError' if there is a mismatch.
   securesystemslib.formats.PATH_SCHEMA.check_match(filepath)
-
-  # If the caller does not provide a password argument, prompt for one.
-  if password is None:
-
-    # It is safe to specify the full path of 'filepath' in the prompt and not
-    # worry about leaking sensitive information about the key's location.
-    # However, care should be taken when including the full path in exceptions
-    # and log files.
-    password = get_password('Enter a password for the Ed25519'
-        ' key (' + TERM_RED + filepath + TERM_RESET + '): ',
-        confirm=True)
-
-  else:
-    logger.debug('The password has been specified. Not prompting for one.')
-
-  # Does 'password' have the correct format?
-  securesystemslib.formats.PASSWORD_SCHEMA.check_match(password)
 
   # If the parent directory of filepath does not exist,
   # create it (and all its parent directories, if necessary).
@@ -522,7 +527,7 @@ def generate_and_write_ed25519_keypair(filepath=None, password=None):
   file_object = tempfile.TemporaryFile()
 
   # Encrypt the private key if 'password' is set.
-  if len(password):
+  if password is not None:
     ed25519_key = securesystemslib.keys.encrypt_key(ed25519_key, password)
 
   else:
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -173,7 +173,6 @@ def _get_key_file_decryption_password(password, prompt, path):
 
 
 
-
 def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
     password=None, prompt=False):
   """Generates RSA key pair and writes PEM-encoded keys to disk.
@@ -205,7 +204,8 @@ def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
     StorageError: Key files cannot be written.
 
   Side Effects:
-      Writes key files to disk.
+    Prompts user for a password if 'prompt' is True.
+    Writes key files to disk.
 
   Returns:
     The private key filepath.
@@ -244,7 +244,6 @@ def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
   securesystemslib.util.persist_temp_file(file_object, filepath)
 
   return filepath
-
 
 
 
@@ -297,8 +296,6 @@ def import_rsa_privatekey_from_file(filepath, password=None,
 
 
 
-
-
 def import_rsa_publickey_from_file(filepath, scheme='rsassa-pss-sha256',
     storage_backend=None):
   """Imports PEM-encoded RSA public key from file storage.
@@ -344,8 +341,6 @@ def import_rsa_publickey_from_file(filepath, scheme='rsassa-pss-sha256',
 
 
 
-
-
 def generate_and_write_ed25519_keypair(filepath=None, password=None,
     prompt=False):
   """Generates ed25519 key pair and writes custom JSON-formatted keys to disk.
@@ -373,6 +368,7 @@ def generate_and_write_ed25519_keypair(filepath=None, password=None,
     StorageError: Key files cannot be written.
 
   Side Effects:
+    Prompts user for a password if 'prompt' is True.
     Writes key files to disk.
 
   Returns:
@@ -419,7 +415,6 @@ def generate_and_write_ed25519_keypair(filepath=None, password=None,
 
 
 
-
 def import_ed25519_publickey_from_file(filepath):
   """Imports custom JSON-formatted ed25519 public key from disk.
 
@@ -451,9 +446,6 @@ def import_ed25519_publickey_from_file(filepath):
     raise securesystemslib.exceptions.FormatError(message)
 
   return ed25519_key
-
-
-
 
 
 
@@ -504,8 +496,6 @@ def import_ed25519_privatekey_from_file(filepath, password=None, prompt=False,
 
 
 
-
-
 def generate_and_write_ecdsa_keypair(filepath=None, password=None,
     prompt=False):
   """Generates ecdsa key pair and writes custom JSON-formatted keys to disk.
@@ -533,6 +523,7 @@ def generate_and_write_ecdsa_keypair(filepath=None, password=None,
     StorageError: Key files cannot be written.
 
   Side Effects:
+    Prompts user for a password if 'prompt' is True.
     Writes key files to disk.
 
   Returns:
@@ -579,7 +570,6 @@ def generate_and_write_ecdsa_keypair(filepath=None, password=None,
 
 
 
-
 def import_ecdsa_publickey_from_file(filepath):
   """Imports custom JSON-formatted ecdsa public key from disk.
 
@@ -606,8 +596,6 @@ def import_ecdsa_publickey_from_file(filepath):
       ecdsa_key_metadata)
 
   return ecdsa_key
-
-
 
 
 
@@ -785,6 +773,7 @@ def import_privatekey_from_file(filepath, key_type=None, password=None,
     raise securesystemslib.exceptions.FormatError(
         "Unsupported key type '{}'. Must be '{}', '{}' or '{}'.".format(
         key_type, KEY_TYPE_RSA, KEY_TYPE_ED25519, KEY_TYPE_ECDSA))
+
 
 
 if __name__ == '__main__':
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -105,12 +105,18 @@ def get_password(prompt='Password: ', confirm=False):
 
 
 def _get_key_file_encryption_password(password, prompt, path):
-  """Encryption password helper.
+  """Encryption password helper for `_generate_and_write_*_keypair` functions.
 
-  - Fail if 'password' is passed and 'prompt' is True (precedence unclear)
-  - Fail if empty 'password' arg is passed (encryption desire unclear)
-  - Return None on empty pw on prompt (suggests desire to not encrypt)
-
+  Combinations of 'password' and 'prompt' -> result (explanation)
+  ----------------------------------------------------------------
+  None                  False -> return None (clear non-encryption desire)
+  "<non-empty string>"  False -> return password (clear encryption desire)
+  <anything else>       False -> raise (bad pw type, unclear encryption desire)
+  <not None>            True  -> raise (unclear password/prompt precedence)
+  None                  True  -> prompt and return password if entered and None
+                                 otherwise (users on the prompt can only
+                                 indicate desire to not encrypt by entering no
+                                 password)
   """
   securesystemslib.formats.BOOLEAN_SCHEMA.check_match(prompt)
 
@@ -142,10 +148,17 @@ def _get_key_file_encryption_password(password, prompt, path):
 
 
 def _get_key_file_decryption_password(password, prompt, path):
-  """Decryption password helper.
+  """Decryption password helper for `import_*_privatekey_from_file` functions.
 
-  - Fail if 'password' is passed and 'prompt' is True (precedence unclear)
-  - Return None on empty pw on prompt (suggests desire to not decrypt)
+  Combinations of 'password' and 'prompt' -> result (explanation)
+  ----------------------------------------------------------------
+  None             False -> return None (clear non-decryption desire)
+  "<any string>"   False -> return password (clear decryption desire)
+  <anything else>  False -> raise (bad pw type, unclear decryption desire)
+  <not None>       True  -> raise (unclear password/prompt precedence)
+  None             True  -> prompt and return password if entered and None
+                            otherwise (users on the prompt can only indicate
+                            desire to not decrypt by entering no password)
 
   """
   securesystemslib.formats.BOOLEAN_SCHEMA.check_match(prompt)
@@ -179,7 +192,7 @@ def _generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
 
   If a password is passed or entered on the prompt, the private key is
   encrypted. According to the documentation of the used pyca/cryptography
-  library encryption is performed "using the best available encryption for a
+  library, encryption is performed "using the best available encryption for a
   given key's backend", which "is a curated encryption choice and the algorithm
   may change over time."  The private key is written in PKCS#1 and the public
   key in X.509 SubjectPublicKeyInfo format.
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -173,7 +173,7 @@ def _get_key_file_decryption_password(password, prompt, path):
 
 
 
-def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
+def _generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
     password=None, prompt=False):
   """Generates RSA key pair and writes PEM-encoded keys to disk.
 
@@ -341,7 +341,7 @@ def import_rsa_publickey_from_file(filepath, scheme='rsassa-pss-sha256',
 
 
 
-def generate_and_write_ed25519_keypair(filepath=None, password=None,
+def _generate_and_write_ed25519_keypair(filepath=None, password=None,
     prompt=False):
   """Generates ed25519 key pair and writes custom JSON-formatted keys to disk.
 
@@ -496,7 +496,7 @@ def import_ed25519_privatekey_from_file(filepath, password=None, prompt=False,
 
 
 
-def generate_and_write_ecdsa_keypair(filepath=None, password=None,
+def _generate_and_write_ecdsa_keypair(filepath=None, password=None,
     prompt=False):
   """Generates ecdsa key pair and writes custom JSON-formatted keys to disk.
 
--- a/securesystemslib/interface.py
+++ b/securesystemslib/interface.py
@@ -122,7 +122,7 @@ def _get_key_file_encryption_password(password, prompt, path):
   if prompt:
     password = get_password("enter password to encrypt private key file "
         "'" + TERM_RED + str(path) + TERM_RESET + "' (leave empty if key "
-        "should not be encrypted): '", confirm=True)
+        "should not be encrypted): ", confirm=True)
 
     # Treat empty password as no password. A user on the prompt can only
     # indicate the desire to not encrypt by entering no password.
@@ -158,7 +158,7 @@ def _get_key_file_decryption_password(password, prompt, path):
   if prompt:
     password = get_password("enter password to decrypt private key file "
         "'" + TERM_RED + str(path) + TERM_RESET + "' "
-        "(leave empty if key not encrypted): '", confirm=False)
+        "(leave empty if key not encrypted): ", confirm=False)
 
     # Treat empty password as no password. A user on the prompt can only
     # indicate the desire to not decrypt by entering no password.
@@ -213,8 +213,6 @@ def _generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
   """
   securesystemslib.formats.RSAKEYBITS_SCHEMA.check_match(bits)
 
-  password = _get_key_file_encryption_password(password, prompt, filepath)
-
   # Generate private RSA key and extract public and private both in PEM
   rsa_key = securesystemslib.keys.generate_rsa_key(bits)
   public = rsa_key['keyval']['public']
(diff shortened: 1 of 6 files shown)


securesystemslib/interface.py
```diff
@@ -569,23 +511,15 @@ def import_ed25519_publickey_from_file(filepath):
...
 
-  # ED25519 key objects are saved in json and metadata format.  Return the
-  # loaded key object in securesystemslib.formats.ED25519KEY_SCHEMA' format that
-  # also includes the keyid.
+  # Load custom on-disk JSON formatted key and convert to its custom in-memory
+  # dict key representation
   ed25519_key_metadata = securesystemslib.util.load_json_file(filepath)
-  ed25519_key, junk = \
-    securesystemslib.keys.format_metadata_to_key(ed25519_key_metadata)
+  ed25519_key, junk = securesystemslib.keys.format_metadata_to_key(
```
joshuagl, line 445: Nit:  I dislike the use of the `junk` variable name here. I think it's confusing when the accepted Python idiom is to use an `_` variable name for unused return values. By having used a name, I would expect to see the variable used somewhere.

lukpueh:
> - Minor: The used sphinx napoleon plugin to render docstrings as html does not handle the custom "Side Effects" section well

Just ticketized this in https://github.com/in-toto/in-toto/issues/401, because we currently only generate sphinx docs in in-toto. But the plan is to fix it upstream, so that we can use the custom section in TUF/in-toto/sslib.

joshuagl (approved):
This is a lovely set of cleanups, thanks a lot @lukpueh! That we have an overall LOC reduction when adding a new function and tests is incredible. I am very pleased with the new docstring style. It uses much less vertical space. I might even be able to read the method being documented at the same time as the docstring. 😆 

My only slight concern is that this PR introduces several behavioural changes to the interface module which arguably break backwards compatibility. I don't think we've made any API stability promises as of yet, but I wanted to at least raise the issue for discussion here.

trishankatdatadog (approved):

securesystemslib/interface.py
```diff
@@ -120,237 +104,194 @@ def get_password(prompt='Password: ', confirm=False):
...
-    Import the PEM file in 'filepath' containing the private key.
 
-    If password is passed use passed password for decryption.
-    If prompt is True use entered password for decryption.
-    If no password is passed and either prompt is False or if the password
-    entered at the prompt is an empty string, omit decryption, treating the
-    key as if it is not encrypted.
-    If password is passed and prompt is True, an error is raised. (See below.)
+def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
+    password=None, prompt=False):
```
jku, line 178: general comment on `generate_*()` functions: If I've understood correctly calling them with default values now leads to creating unencrypted keys. It's not horrible but I think the expected default would be prompting...

Instead of `prompt` these functions could maybe have `encrypted=True`? By default this would lead to a prompt if password was not given, but you could set `encrypted=False` to create unencrypted keys.

jku (commented):
Looks good, the cleanups make the code much more readable.

securesystemslib/interface.py
lukpueh: Agreed. Will remove.

securesystemslib/interface.py
```diff
@@ -991,6 +732,61 @@ def import_publickeys_from_file(filepaths, key_types=None):
...
+
+def import_privatekey_from_file(filepath, key_type=None, password=None,
+    prompt=False):
+  """Imports private key from file.
+
+  If a password is passed or entered on the prompt, the private key is
+  decrypted, otherwise it is treated as unencrypted.
+
+  NOTE: The default signing scheme 'rsassa-pss-sha256' is assigned to RSA keys.
+  Use 'import_rsa_publickey_from_file' to specify any other than the default
```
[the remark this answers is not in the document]
lukpueh: Good eyes! Thanks for catching. :)

tests/test_interface.py
```diff
@@ -529,11 +530,26 @@ def test_ecdsa(self):
 
 
     # TEST: Generation errors
+    for idx, (kwargs, err_msg) in enumerate([
+        # Error on empty password
+        ({"password": ""},
+          "encryption password must be 1 or more characters long"),
+        # Error on 'password' and 'prompt=True'
+        ({"password": pw, "prompt": True},
+          "passing 'password' and 'prompt=True' is not allowed")]):
```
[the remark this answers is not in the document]
lukpueh: I agree that it's annoying and prone to break. As a matter of fact I'm pretty sure that it will break in the next build... I just skimmed through the latest `cryptography` release and saw that they changed an error message that we check for here.

That said, I think there is merit in more granular error test assertions that go beyond the type. I've seen tests in the past, which meant to check for one error and passed although a different error was raised, because they had the same type.

So I think it's a fair trade-off to update the expected error strings in tests every now and then, if it gives us stronger tests. And once we have a more solid exception taxonomy (see #191), I'm happy to remove the message assertions. What do you think?

securesystemslib/interface.py
lukpueh: Good points, @jku! 

We discussed using an `encrypt` instead of a `prompt` parameter in https://github.com/secure-systems-lab/securesystemslib/pull/124#issuecomment-371621875 pp. but opted for `prompt` to be more explicit about whether user interaction is required. However, I do see how one could equally well argue for being more explicit about encryption. 

Regarding default, I agree that from a security point of view `prompt=True` makes more sense (i.e. secure by default). From a usability point of view I don't like that an API function opens a prompt by default. But the former is probably more important. What do others think? (cc @joshuagl, @mnm678, @adityasaky, @SantiagoTorres,  @trishankatdatadog).

lukpueh:
> My only slight concern is that this PR introduces several behavioural changes to the interface module which arguably break backwards compatibility. I don't think we've made any API stability promises as of yet, but I wanted to at least raise the issue for discussion here.

@joshuagl, I'm not so worried about that. `securesystemslib` is mostly for internal (TUF/in-toto) use and we are still in pre-1.X.X-phase. What's IMO more important is that we get the interface "right" (see [comment/discussion about defaults above](https://github.com/secure-systems-lab/securesystemslib/pull/288#discussion_r511834954)).

securesystemslib/interface.py
jku: > I don't like that an API function opens a prompt by default

Fair point, understood.

I guess you could decide that user has to choose something: leaving both `password` and `prompt` unset would be an error. Then calling the function with default argument values would fail with ValueError instead of prompting unexpectedly.

To create unencrypted keys you would have to `generate_and_write_rsa_keypair(filepath, prompt=False)`: it's not ideal because nothing in that explicitly shows that the result is going to be unencrypted... so it's still not great.

trishankatdatadog: Yes, we should encrypt by default, no question about it, and principle of least surprise. Could the API be designed better? Sure, but let's leave that for the future.

mnm678: I think we should only create unencrypted keys if the user explicitly sets this, for example `encrypted=False`, or at the very least we can set `prompt=True` by default.

joshuagl: An API function that opens a prompt by default does seem suboptimal, but we can mitigate that somewhat with documentation making clear that this default is only to ensure security by default. I agree with the sentiment that secure by default would be preferable and less surprising.

lukpueh: > I guess you could decide that user has to choose something: leaving both password and prompt unset would be an error. Then calling the function with default argument values would fail with ValueError instead of prompting unexpectedly.

Do you mean?:
```
(password=None, prompt=None)  --> error (default)
(password=None, prompt=True)  --> prompt
(password=None, prompt=False) --> plain
(password="..", prompt=None)  --> encrypt
(password="..", prompt=True)  --> error
(password="..", prompt=False) --> encrypt
```
I think we can do that, but I'm not a big fan of raising an exception per default, nor of using 3 values for a boolean.

> To create unencrypted keys you would have to generate_and_write_rsa_keypair(filepath, prompt=False): it's not ideal because nothing in that explicitly shows that the result is going to be unencrypted... so it's still not great.

If we go with above suggestion, I think it makes more sense to use `encrypt` instead of `prompt`. 

> I guess a combo is possible as well: user needs to choose exactly one of password=<value>, prompt=True or encrypted=False -- then each call would be explicit about what it wants. In another language this would be a SecurityType enum with three possible choices but that seems a bit clunky in Python

That  -- i.e.  pass a subclass of `KeySerializationEncryption`, e.g. `BestAvailableEncryption(password)` or `NoEncryption` --  is what [`pyca/cryptography` does](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization.html#serialization-encryption-types). I wouldn't say that it is generally clunky in Python, but it doesn't really fit in with the general style of our interface.

lukpueh: I think we have to choose between:
1. default encryption, default prompt
1. no default encryption, no default prompt
1. more sophisticated parameters

jku: Yes, you understood my ideas as I meant them. I agree with your comments (raising an exception as default is not ideal, enum style doesn't fit API well)

tests/test_interface.py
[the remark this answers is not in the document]
lukpueh: FYI: Looks like I was wrong about the [`cryptography` update](https://github.com/pyca/cryptography/commit/58494b41d6ecb0f56b7c5f05d5f5e3ca0320d494#diff-97598491194e5ffd150e82dea4ddb874833c27432477fada5ef1bc9990c4b70fL138-R133) breaking our build. That particular error message is raised by ourselves.

https://github.com/secure-systems-lab/securesystemslib/blob/d05dc6a42908b9ef3d86e8074df6cf0b8e8280b3/securesystemslib/rsa_keys.py#L1062

So no need for action for now... :)

securesystemslib/interface.py
lukpueh: @trishankatdatadog and @joshuagl, you both approved the PR (with no default encryption) but also signaled that you would like to see default encryption. I am unsure how to proceed.

My preferred course of action would be to merge as is, i.e. without default encryption, for the the following (partially interdependent) reasons:
- I'd like to land this asap, because it blocks in-toto 1.0.0.
- I have already adopted the changes from this PR on in-toto (https://github.com/in-toto/in-toto/pull/402) and tuf (will PR later) sides.
- I am against default prompts (see UX).
- I am against more sophisticated parameters (see UX/consistency and *asap*-argument).
- We can reconsider this for securesystemslib v1.0.0. Who knows, we might even remove key generation functionality and limit `securesystemslib` to parsing standard, out-of-band generated key formats, translating them into tuf/in-toto formats, and using them to sign/verify.
- I agree that default encryption is desirable but does not outweigh all these arguments


Can I get 👍 /👎  if others are okay with this?

joshuagl: I'm OK with merging as-is and planning to improve in the future by either: 1) removing key generation functionality 2) implementing encryption by default.

Should we file an issue in the 1.0 milestone to capture the discussion here?

SantiagoTorres: I wanted to add some comments I shared privately. It would worry me that an unintended side-effect is for the function to block (waiting for stdin), as I assume this would catch many people by surprise, may be difficult to debug (e.g., a CI build hanging, or tox randomy hanging with no stdout/in redirected anywhere).

I'm not entirely sure what's the user story we have in mind, but for a library that is generating keys in a program, it also feels like this is a feature that many people may not want. This is reminiscent of that time we added colorama, and all the consequences that ensured (i.e., very little benefit, and a lot of maintenance burden).

Although I agree that secure by default is a thing, I also wonder if this is adding meaningful security for most of the usecases. I'd expect the following:

- The API function to be noninteractive (hell, we don't even know if this is going to be attached to a terminal)
- A CLI tool that uses this function to set up a prompt (or use this batteries-included setting). 

I don't think at the time of writing this function we have enough context to know how users will use this regularly, so I err on the side of non-interactivity by default.

securesystemslib/interface.py
```diff
@@ -120,237 +104,194 @@ def get_password(prompt='Password: ', confirm=False):
...
     pem_key = file_object.read().decode('utf-8')
 
-  # Convert 'pem_key' to 'securesystemslib.formats.RSAKEY_SCHEMA' format.
-  # Raise 'securesystemslib.exceptions.CryptoError' if 'pem_key' is invalid.
-  # If 'password' is None decryption will be omitted.
-  rsa_key = securesystemslib.keys.import_rsakey_from_private_pem(pem_key,
-      scheme, password)
+  # Optionally decrypt and convert PEM-encoded key to 'RSAKEY_SCHEMA' format
+  rsa_key = securesystemslib.keys.import_rsakey_from_private_pem(
+      pem_key, scheme, password)
```
SantiagoTorres, line 294: I have a small gripe with the way we set up these two functions. It feels to me that the pattern I most of the time do is:

```
generate a key
use the key right away
```

It is rather confusing to me that the filepath is returned, rather than the generated key itself, specially given that there is no default filepath (so we don't have to provide users any disambiguation). So with this in mind, what's the added value of returning the filename rather than the key itself?

SantiagoTorres (approved):
I have a small comment, yet I don't know if it's a blocker (I may have well missed the original rationale).

securesystemslib/interface.py
trishankatdatadog: > @trishankatdatadog and @joshuagl, you both approved the PR (with no default encryption) but also signaled that you would like to see default encryption. I am unsure how to proceed.

My bad. I would rather that we encrypt keys on disk rather than our users shoot themselves in the foot. Having said that, TUF/in-toto can take care of that rather than SSLlib. Does this answer your question?

trishankatdatadog (changes requested):
I would rather that we encrypt keys on disk rather than our users shoot themselves in the foot. Having said that, TUF/in-toto can take care of that rather than SSLlib. Does this work?

securesystemslib/interface.py
lukpueh: Agreed. This was just not a focus of this PR.

lukpueh:
> Having said that, TUF/in-toto can take care of that rather than SSLlib. Does this work?

I'd rather try to get this right right away, because I want to expose this interface directly to in-toto users and not create an additional layer that I need to then maintain in in-toto (and TUF).

lukpueh:
Let me explore alternatives...

trishankatdatadog:
> Let me explore alternatives...

What about https://github.com/theupdateframework/tuf/pull/1191#issuecomment-718700406?

securesystemslib/interface.py
lukpueh: Also, it is already possible to do that one level deeper in the abstraction hierarchy (see `securesystemslib.keys.generate_*_key`).

SantiagoTorres: Fair, I still wonder if we'd like to change the return value for the key instead of the filepath?

SantiagoTorres:
> I would rather that we encrypt keys on disk rather than our users shoot themselves in the foot. Having said that, TUF/in-toto can take care of that rather than SSLlib. Does this work?

I think that'd be ideal in my opinion. In the same way the openssl library doesn't encrypt by default on keygen, but the openssl binary requires you to pass -nopasswd (or passwd '' I forget) to avoid encryption.

lukpueh:
Let me recap...

**Original behavior**:

Before this PR the `interface.generate_and_write_*_keypair` had only an optional `password` argument to control encryption (and no `prompt` argument) and behaved like so:

- `password=None` (default): prompt for password
  - if empty: no encryption
  - else: encryption
- `password="<non-empty-string>"`: encryption
- `password=""`: no prompt and no encryption

(NOTE: `generate_and_write_ecdsa_keypair` behaved slightly different, i.e. it would encrypt anyway, even with an empty passed or entered password)

**New behavior**:

All `interface.generate_and_write_*_keypair` now have an optional `prompt` argument and consistently behave like so:

- `password=None, prompt=False` (default): no encryption
- `password="<non-empty-string>", prompt=False`: encryption
- `password="<any-string>", prompt=True`: error
- `password="", prompt=False`: error
- `password=None, prompt=True`: prompt for password
  - if empty: no encryption
  - else: encryption


**Motivation:**

My main motivation was to make encryption control and prompt behavior more explicit and surprise-free, so that I could remove the [`in_toto.util`](https://github.com/in-toto/in-toto/blob/v0.5.0/in_toto/util.py)-facade, which had thin wrappers around generate functions, i.e.
- `in_toto.util.generate_*_keypair(.., password=""`): no auto-prompt, but also no encryption by default
- `in_toto.util.prompt_generate_*_keypair()`: always prompts for a password

I admit that I did not prioritize the security by default principle.

I also tried to withstand the temptation to address other related key interface issues, like
  - module architecture / choice of abstraction layers  (see https://github.com/secure-systems-lab/securesystemslib/issues/270 and https://github.com/secure-systems-lab/securesystemslib/pull/288#discussion_r513570842)
  - non-standard on-disk key formats for ed25519, ecdsa and other key format issues ... (see https://github.com/secure-systems-lab/securesystemslib/issues/251)


**Alternatives to promote default encryption**


1. Revert to original behavior described above
  Cons:
    - Non-surprise-free encryption control
    - default prompt if no password is passed
1. Set default `prompt=True`
  Cons:
    - Default prompt if no password is passed
    - Unclear behavior if password is passed
      (error, or prioritize one of prompt=True or password)
1. Set default `prompt=None` and require either a password or explicitly setting prompt to True or False
   - Default error
   - Three cases for a boolean argument (None, True, False)

2. Use a complex mandatory argument, e.g. `encryption_method`, that allows to pass exactly one of *password*, *prompt=True*, or *no-encrypt* (akin to what `pyca/securesystemslib` does)
   - Unexpected API style give the reset of `securesystemslib`

1. Provide different functions to create unencrypted and encrypted keys (@trishankatdatadog, would you mind describing the exact behavior you have in mind?)
  Cons:
    - Might require more complicated case handling in calling code

1. Remove key generation functionality
  *I think this is what we should aim for in the long run, but it requires us  to support common standard on-disk formats for all our key types.*

SantiagoTorres:
Thank you for the detailed recap lukas! I also see how valuable is the prompt/password split and the more standard behavior between the different keytype variants of the generate* family:

My vote is on:
- the default is no encryption, no prompt.
- once we use the newer functions in in-toto/TUF, we override this default to prompt, encryption, with the option for the user to say *no*

I agree with point 7 as well...

trishankatdatadog:
> 5\. Provide different functions to create unencrypted and encrypted keys (@trishankatdatadog, would you mind describing the exact behavior you have in mind?)

We split the `interface.generate_and_write_*_keypair` functions into something like `interface.generate_and_write_*_encrypted_keypair` and `interface.generate_and_write_*_unencrypted_keypair`. Verbose, but gets rids of prompts. Slightly more work, but I think this is a better long-term solution with not too much short-term work.

lukpueh:
Thanks for elaborating, @trishankatdatadog! What parameters should `generate_and_write_*_encrypted_keypair` receive and how should they behave?

trishankatdatadog:
> Thanks for elaborating, @trishankatdatadog! What parameters should `generate_and_write_*_encrypted_keypair` receive and how should they behave?

I suppose `encrypted()` would take a required passphrase, so no prompt whatsoever (we can finally remove the colorama drama). Otherwise, the two functions should look alike. Do we need anything else?

P.S. Sorry, I'm a bit distracted right now, leaving on a plane in a few hours. I trust you will all make the right decision in case I don't reply in time.

lukpueh:
> once we use the newer functions in in-toto/TUF, we override this default to prompt, encryption, with the option for the user 

You're right, @SantiagoTorres, so we could replace the [`--prompt` flag in `in-toto-keygen`](https://github.com/in-toto/in-toto/blob/a1c5ecdef54c3d17da9a5d904d37c6b46ba4f82d/in_toto/in_toto_keygen.py#L80-L82) with something like `--no-encrypt` and default-prompt if  `--no-encrypt` is not passed.
And we could do the same in TUF's `repo.py` CLI tool, which also provides a keygen feature.

The reason why I didn't do that is because I'd like to not spend any more time on `in-toto-keygen` (there's no real advantage to using the API and it was only created for a [currently hibernating side-project](https://github.com/in-toto/layout-web-tool)), and as said above I'd love it if people could use `ssh-keygen` or `openssl` to generate their TUF/in-toto keys in the future. Same goes for `repo.py` which has many documented issues and an uncertain future (https://github.com/theupdateframework/tuf/issues/881).

lukpueh:
Speaking of [secure defaults and `repo.py`](https://github.com/theupdateframework/tuf/blob/develop/docs/CLI.md#create-a-repository):

>The default top-level key files created with `--init` are saved to disk
encrypted, with a default password of 'pw'.

In [its defense](https://github.com/theupdateframework/tuf/blob/develop/docs/QUICKSTART.md#quickstart)...
> [u]nlike the underlying TUF modules that the CLI uses, the CLI itself is a bit bare-bones. Using the CLI is the easiest way to familiarize yourself with how TUF works, however. It will serve as a very basic update system.

lukpueh:
> ... encrypted() would take a required passphrase, so no prompt whatsoever (we can finally remove the colorama drama)

That would actually be nice. :)

Overall I am not so convinced if providing two functions is such a big win for security. It still takes the cognitive effort to read the function signatures and pick one or the other. And I think the function signatures (especially with updated docs) as proposed in this PR are pretty clear too, but I'm obviously biased. 

Here once more for others to compare, slightly simplified for emphasis:

```
generate_unencrypted()
generate_encrypted(password)
generate(password=None, prompt=False)
```

I'll let this PR rest until tomorrow. Maybe others still want to weigh in.

lukpueh:
Good. So I got convinced that we do want default encryption... and we still agree that the original behavior (with default prompts) is not desirable. Lets explore [above alternatives](https://github.com/secure-systems-lab/securesystemslib/pull/288#issuecomment-718831595)  in more detail so that we can have the cake and eat it too.



complex mandatory argument... (4)
```python
PROMPT = 0x01
NO_ENCRYPT = 0x02

def gen(password):
  ...

# Usage:
gen("<non-empty-string>") # encrypt (default)
gen(PROMPT)               # prompt
gen(NO_ENCRYPT)           # store plain
gen(<anything-else>)      # error

```


different functions to create unencrypted and encrypted keys (5)
```python
def gen_unencrypted():
  ...
def gen_encrypted(password, prompt=False):
  ...

# Usage:
gen_encrypted("<non-empty-string>")    # encrypt (default)
gen_encrypted("<anything-else>")       # error
gen_encrypted(None, prompt=True)       # prompt
gen_encrypted(<not-None>, prompt=True) # error

```



And yet another alternative
```python
def gen(password, prompt=False, plain=False):
  ...

gen("<non-empty-string>")                # encrypt (default)
gen(<anything-else>)                     # error
gen("<not-None>", plain=True)            # error
gen("<not-None>", prompt=True)           # error
gen(<anything>, prompt=True, plain=True) # error
gen(None, prompt=True)                   # prompt
gen(None, plain=True)                    # store plain

```

jku:
I'd remove the "_encrypted" postfix in case 5:
_I also combined the two error cases in usage section_
```python
def gen_unencrypted():
  ...
def gen(password, prompt=False):
  ...
# Usage:
gen("<non-empty-string>")  # encrypt (default)
gen(None, prompt=True)     # prompt
gen(<anything else>)       # error
```
I guess the question is: if password is already an Optional argument (in the `typing` sense: it can be None), do we need prompt at all or can `password=None` now imply that: it would not be the default so it should not surprise anyone?

--- 

Maybe there is also case 7:
```python
def gen_unencrypted():
  ...
def gen_prompt():
  ...
def gen(password):
  ...
```
Sure it's a lot of functions but now suddenly the usage section isn't needed at all so it seems understandable.

trishankatdatadog:
I think I like case 7 the best. A single, public function with multiple, potentially conflicting parameters is not only confusing for the developer, but also the end-user. Internally, the multiple public functions can share many private functions.

mnm678:
I like either 



> ```python
> def gen(password, prompt=False, plain=False):
> ```

or


> ```python
> def gen_unencrypted():
>   ...
> def gen_prompt():
>   ...
> def gen(password):
>   ...
> ```

I think both of these make the options clear to the user, while encouraging encryption.

SantiagoTorres:
I'm a little wary of the argument that encryption is somehow "better" by virtue of existing. We've seen that (e.g., in repo.py) encryption with default passphrases takes place more often than not, and it does give this false sense of security. Next time I run into TUF repository keys, I'll try the password 'pw', as it's a very likely case.

I really really want us to consider that we need to take the usage of the library into account, and not just assume "encryption == better".  A password like "pw" for an encrypted key (as was the case for repo.py) can be broken in less than a minute on a regular laptop, and without using a GPU. 

Defaulting to encryption will implicitly require people of figuring out exactly how to derive/receive/handle a password, and all of its user-facing consequences. This is a general use library, and as such it shouldn't make assumptions on what downstream users are trying to do. I still bin this into the issue that sslib really has no user story; back then, it was supposed to be a multi-crypto-provider abstraction layer, but I can't really say what is it supposed to be today.

To be clear,  I'm not strongly opposed to encryption by default, but rather the argument that it's somehow better just because it's there.

trishankatdatadog:
@SantiagoTorres I understand your concerns, so perhaps we can do this. `repo.py` may use unencrypted keys for demo purposes, and warn that this is the case. Everyone else, including serious integrators, can use encrypted keys. Would this hold up anything for anyone, or is it not possible for some reason right now?

lukpueh:
Thanks for your suggestions, @jku!

> do we need prompt at all or can password=None now imply that: it would not be the default so it should not surprise anyone?

I guess that would work, although it is not the kind of explicitness about prompt I had hoped for.  But I'm willing to compromise. :) 


> Maybe there is also case 7:

That's similar to what I had in [`in_toto.util`](https://github.com/in-toto/in-toto/blob/c775e3b260a50ec5fb673a9b725e2d2eacedfb6b/in_toto/util.py#L262-L273) and hoped to get rid of in order to [spare me some `elif`s](https://github.com/uptane/uptane/blob/13e8b8ca4a990c6b94eb39ce0812e59b128f2f75/dev-requirements.txt#L7) in in-toto and interface code in general.

----

But let me quickly pick up @SantiagoTorres' question about the user story behind sslib or at least it's key gen functions. IMO we have the following use cases:
- **tuf/in-toto adopters generate production keys**
  This one makes the strongest case for a foolproof secure by default interface, and probably justifies the extra amount of interface code. As said before, I think we should push for better interoperability with standard keygen tools in the future, so that we don't have to provide this interface anymore. But until then and maybe even afterwards we do want a nice design for it.

- **python-{tuf/in-toto} command line tool development**
   I think here developer convenience/flexibility is more important than the foolproofness. Ideally I just want to throw the cli args as I get them into a single function call of securesystemslib (see comment about `elif` above).
   Besides, the same argument about the necessity of custom keygen tools applies here. If users can generate their keys with `openssl`, `ssh-keygen` or `gpg`, why should we develop our own.
   And the current cli tools we have are really mainly for educational purposes (see discussion about `in-toto-keygen` and `repo.py` above).

- **python-{tuf/in-toto} test code**
   Encryption is usually not needed, except for testing decryption of keys. So convenience definitely beats security here. 

- **tuf/in-toto demo code / reference**
   Security and foolproofness does not seem to be a priority here, but it's definitely a good idea to exemplify the best practice of key encryption.

lukpueh:
What about the following compromise?: 

For use cases 2 and 3 (flexibility/convenience) we keep the functions [as proposed in this PR](https://github.com/secure-systems-lab/securesystemslib/pull/288#issuecomment-718831595). 
And for use cases 1 and 4 (foolproof/secure by default) we add wrappers per key type with the following behavior:

```python
gen("<non-empty-string>") # encrypt
gen(None):                # don't encrypt
gen(<anything else>)      # error 
```
Note that for use cases 1 and 4 prompt is not really needed.

If we can agree on that (?), we only need to come up with good names for the two different functions. Any ideas?

joshuagl:
I am happy with the proposed compromise. I would even go so far as to suggest that `gen(None)` should result in a warning being logged.

> If we can agree on that (?), we only need to come up with good names for the two different functions. Any ideas?

If we want to have the new functions be logically tied to the existing `interface.generate_and_write_*_keypair()` functions, we could go with: `interface.generate_and_write_*_encrypted_keypair(password)` ?

If instead we want the signature to be distinct enough that it makes users stop and think, we could perhaps use: `interface.generate_*_keypair(password)`?

trishankatdatadog:
> Note that for use cases 1 and 4 prompt is not really needed.

Sorry, disagree. We certainly use the prompt. What would be the proposed functions for these use cases {1,4} then, please?

lukpueh:
> We certainly use the prompt.

Okay, but you also suggested to get rid of prompts in https://github.com/secure-systems-lab/securesystemslib/pull/288#issuecomment-718854074, @trishankatdatadog.

> We split the `generate_and_write_*_keypair` functions into something like `generate_and_write_*_encrypted_keypair` and `generate_and_write_*_unencrypted_keypair`. Verbose, but gets rids of prompts.

and https://github.com/secure-systems-lab/securesystemslib/pull/288#issuecomment-718868569

> `encrypted()` would take a required passphrase, so no prompt whatsoever

lukpueh:
> What would be the proposed functions for these use cases {1,4} then, please?

I described them in https://github.com/secure-systems-lab/securesystemslib/pull/288#issuecomment-721011782 and @joshuagl suggested names for them in https://github.com/secure-systems-lab/securesystemslib/pull/288#issuecomment-721219492.

If use cases 1 and 4 do need prompts, the fully flexible not default encrypting `generate_and_write_*_keypair` as proposed in this PR remain available too.

lukpueh:
But let's just go for @jku's *"suggestion 7"*, which had the best acceptance. Writing a bunch of extra wrappers + docs + tests definitely seems less cumbersome than further discussions :P 

Fully fleshed out this could look like so:

```python
def _generate_and_write_*_keypair(filepath=None, password=None, prompt=False):
    # We make the fully flexible but not secure by default  `generate_and_write_*_keypair`
    # as currently proposed "protected" (leading underscore). Users who know what they do
    # can still  use it internally, i.e. below but also in TUF/in-toto code.

def generate_and_write_*_keypair(password, filepath=None)
    # For full default security the default generation function will only generate encrypted keys
    # NOTE: Given that password is now a positional argument this changes the order  
    # (before the 'filepath' kwarg used to be the first argument)
    PASSWORD_SCHEMA.check_match(password) 
    _generate_and_write_*_keypair(fileapath=filepath, password=password, prompt=False)

def generate_and_write_*_keypair_with_prompt(filepath=None)
    # To encrypt keys with a prompt use this separate function
    # NOTE: An empty string on the prompt will not encrypt the key
    # TODO: We should signal this on the prompt. 
    _generate_and_write_*_keypair(fileapath=filepath, password=None, prompt=True)

def generate_and_write_unencrypted_*_keypair(filepath=None)
    # We also expose a function that unambiguously creates non-encrypted keys.
    _generate_and_write_*_keypair(fileapath=filepath, password=None, prompt=False)

```

trishankatdatadog:
> Okay, but you also suggested to get rid of prompts in [#288 (comment)](https://github.com/secure-systems-lab/securesystemslib/pull/288#issuecomment-718854074), @trishankatdatadog.

My bad. This is what happens when you comment while traveling and jet-lagged during pandemic.

> But let's just go for @jku's _"suggestion 7"_, which had the best acceptance. Writing a bunch of extra wrappers + docs + tests definitely seems less cumbersome than further discussions :P

Makes sense to me.

trishankatdatadog:
> My bad. This is what happens when you comment while traveling and jet-lagged during pandemic.

Now I remember what I was thinking. I wanted to remove the complicatedness of handling prompts in securesystemslib, and push it to integrators such as myself. Still willing to do this. So go ahead and remove the prompts for now if it makes your life simpler.

tests/test_interface.py
```diff
@@ -100,13 +110,12 @@ def tearDown(self):
 
 
   def test_rsa(self):
-    """Test RSA key generation and import interface functions. """
+    """Test RSA key _generation and import interface functions. """
```
joshuagl, line 113:
```suggestion
    """Test RSA key generation and import interface functions. """
```

tests/test_interface.py
```diff
@@ -258,16 +282,19 @@ def test_rsa(self):
...
 
+    # bad prompt
+    with self.assertRaises(FormatError):
+      import_rsa_privatekey_from_file(fn_default, prompt="not-a-bool")
+
 
 
   def test_ed25519(self):
-    """Test ed25519 key generation and import interface functions. """
+    """Test ed25519 key _generation and import interface functions. """
```
joshuagl, line 292:
```suggestion
    """Test ed25519 key generation and import interface functions. """
```

tests/test_interface.py
```diff
@@ -420,23 +462,20 @@ def test_ed25519(self):
...
 
 
   def test_ecdsa(self):
-    """Test ecdsa key generation and import interface functions. """
-    # NOTE: Unlike rsa and ed25519, the ecdsa (key creation and private key
-    # import) interface only supports encrypted keys, even if the passed or
-    # prompted password is an empty string.
-
-    # TEST: Generate pw encrypted keys and import
+    """Test ecdsa key _generation and import interface functions. """
```
joshuagl, line 471:
```suggestion
    """Test ecdsa key generation and import interface functions. """
```

joshuagl (commented):
Apart from some over-eager find/replace the additional changes look reasonable to me.

I'd like to hear from others who contributed to the discussion before we merge them, though. cc @jku @trishankatdatadog @SantiagoTorres @mnm678

lukpueh:
Many thanks for the quick review, @joshuagl! 

> Apart from some over-eager find/replace the additional changes look reasonable to me.

I added these on purpose because `test_{rsa, ed25519, ecdsa}` do only test the `_generate_*` and not the `generate_*` functions. The latter are tested in the newly added `test_generate_keypair_wrappers` function.

lukpueh:
> I'd like to hear from others who contributed to the discussion before we merge them, though. cc @jku @trishankatdatadog @SantiagoTorres @mnm678

So would I!

FYI I implemented @jku's approach as sketched in https://github.com/secure-systems-lab/securesystemslib/pull/288#issuecomment-721638144. 

See the [most recent three commits](https://github.com/lukpueh/securesystemslib/compare/1db7f213d8a5b207b4ff4a3e1d41e459760690d7...misc-interface-enhancements) for these changes, and  https://in-toto-lukpueh.readthedocs.io/en/latest/api.html#generate-key-pairs, for what the docs look like on RTD.

lukpueh:
If we decide to merge this, I'll update the PR description.

jku (approved):
LGTM :+1:

securesystemslib/interface.py
```diff
@@ -120,107 +104,141 @@ def get_password(prompt='Password: ', confirm=False):
...
-  else:
-    logger.debug('The password has been specified.  Not prompting for one')
-
-  # Does 'password' have the correct format?
-  securesystemslib.formats.PASSWORD_SCHEMA.check_match(password)
-
-  # Encrypt the private key if 'password' is set.
-  if len(password):
+  # Encrypt the private key if a 'password' was passed or entered on the prompt
+  if password is not None:
```
trishankatdatadog, line 230: Maybe we can do the password schema check here instead of forcing it on the direct/indirect caller?

securesystemslib/interface.py
```diff
@@ -229,187 +247,191 @@ def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
...
+  Returns:
+    The private key filepath.
 
-def import_rsa_privatekey_from_file(filepath, password=None,
-    scheme='rsassa-pss-sha256', prompt=False,
-    storage_backend=None):
   """
-  <Purpose>
-    Import the PEM file in 'filepath' containing the private key.
+  securesystemslib.formats.PASSWORD_SCHEMA.check_match(password)
```
trishankatdatadog, line 280: See my comment above about moving the password schema check to `_get_key_file_encryption_password()`.

securesystemslib/interface.py
```diff
@@ -229,187 +247,191 @@ def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
...
+    An RSA private key object conformant with 'RSAKEY_SCHEMA'.
+
+  """
+  securesystemslib.formats.PATH_SCHEMA.check_match(filepath)
+  securesystemslib.formats.RSA_SCHEME_SCHEMA.check_match(scheme)
+
+  password = _get_key_file_decryption_password(password, prompt, filepath)
 
   if storage_backend is None:
     storage_backend = securesystemslib.storage.FilesystemBackend()
```
trishankatdatadog, line 389: Off-topic, but: now that I see how we keep using this, we should make a singleton out of `FilesystemBackend()` rather than instantiating it over and over again wherever we need to use it. @joshuagl

securesystemslib/interface.py
```diff
@@ -229,187 +247,191 @@ def generate_and_write_rsa_keypair(filepath=None, bits=DEFAULT_RSA_KEY_BITS,
...
-    logger.debug('No password was given. Attempting to import an'
-        ' unencrypted file.')
+  Returns:
+    An RSA private key object conformant with 'RSAKEY_SCHEMA'.
+
+  """
+  securesystemslib.formats.PATH_SCHEMA.check_match(filepath)
+  securesystemslib.formats.RSA_SCHEME_SCHEMA.check_match(scheme)
+
+  password = _get_key_file_decryption_password(password, prompt, filepath)
```
trishankatdatadog, line 386: This call would fail if the user does not set either the password or prompt, correct?

tests/test_interface.py
```diff
@@ -100,13 +110,12 @@ def tearDown(self):
...
   def test_rsa(self):
-    """Test RSA key generation and import interface functions. """
+    """Test RSA key _generation and import interface functions. """
 
     # TEST: Generate default keys and import
     # Assert location and format
     fn_default = "default"
-    fn_default_ret = generate_and_write_rsa_keypair(
-        filepath=fn_default, password="")
+    fn_default_ret = _generate_and_write_rsa_keypair(filepath=fn_default)
```
trishankatdatadog, line 118: Unencrypted, correct?

trishankatdatadog (commented):
Looks great, fantastic job, thanks for seeing this through, Lukas! Just a few comments...

tests/check_public_interfaces.py
```diff
@@ -64,7 +68,26 @@ def test_interface(self):
 
     with self.assertRaises(
           securesystemslib.exceptions.UnsupportedLibraryError):
```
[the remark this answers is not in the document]
lukpueh: To check that all interface functions return the expected Exception, if the optional dependencies `pyca/cryptography`, `pyca/pynacl` and `gpg` are not available. See `tox.ini` and `travis.yml` how these tests are called with required dependencies only.

securesystemslib/interface.py
```diff
@@ -120,107 +104,141 @@ def get_password(prompt='Password: ', confirm=False):
...
+        "(leave empty if key not encrypted): ", confirm=False)
+
+    # Treat empty password as no password. A user on the prompt can only
+    # indicate the desire to not decrypt by entering no password.
+    if not len(password):
+      return None
+
+  if password is not None:
+    securesystemslib.formats.PASSWORD_SCHEMA.check_match(password)
+    # No additional vetting needed. Decryption will show if it was correct.
```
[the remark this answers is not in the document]
lukpueh: Exactly.

securesystemslib/interface.py
```diff
@@ -120,107 +104,141 @@ def get_password(prompt='Password: ', confirm=False):
...
+  key in X.509 SubjectPublicKeyInfo format.
+
+  NOTE: A signing scheme can be assigned on key import (see import functions).
+
+  Arguments:
+    filepath (optional): The path to write the private key to. If not passed,
+        the key is written to CWD using the keyid as filename. The public key
+        is written to the same path as the private key using the suffix '.pub'.
+    bits (optional): The number of bits of the generated RSA key.
+    password (optional): An encryption password.
```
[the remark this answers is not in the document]
lukpueh: My goal was to remove most redundancy in the docstrings to keep them as dense as possible. The information you suggest is already available in the `Raises` section, i.e. `ValueError: An empty string is passed as 'password'...`

securesystemslib/interface.py
lukpueh: We check in the `_get_key_file_encryption_password` helper, which we call right before.

securesystemslib/interface.py
lukpueh: That's already the case. However, it only checks if the password is not None. That's because `_generate_and_write_*_keypair` (which uses `_get_key_file_encryption_password`) accept None for a password so that callers can indicate the desire to not encrypt. (see docstrings)

`generate_and_write_*_keypair` OTOH do not accept None as password, that's why we unconditionally check the schema here right away.

securesystemslib/interface.py
lukpueh: Good idea! Mind submitting a feature request?

tests/test_interface.py
lukpueh: Correct.

securesystemslib/interface.py
lukpueh: No, it will return None. Because callers may use `import_*_privatekey_from_file` with neither password nor `prompt=True` to import an unencrypted private key.

See `import_*_privatekey_from_file` docstrings:
*"If a password is passed or entered on the prompt, the private key is decrypted, otherwise it is treated as unencrypted."*

tests/test_interface.py
lukpueh: But note that this is the protected internal `_generate_and_write_rsa_keypair`  (see `_`-prefix)  function for flexible but hazardous use. The public `generate_and_write_rsa_keypair` does not accept an empty password, and thus can not write unencrypted keys.

lukpueh:
Thanks for the comments, @trishankatdatadog. I integrated one of your suggestions and tried clarifying your other questions, both here inline and also by updating the helper function docstrings.

trishankatdatadog (approved):

lukpueh:
Thanks for all the reviews and comments! The required adoptions for tuf and in-toto are pending in https://github.com/theupdateframework/tuf/pull/1191 and https://github.com/in-toto/in-toto/pull/408. And I updated the PR description here to reflect the recent changes. Merging ...

SantiagoTorres:
Hi, this seems like it's been approved enough and it also LGTM. I'm going to merge it.

lukpueh:
Thanks for merging, @SantiagoTorres! Just realized that I forgot to update the snippets in README.rst a second time after the most recent commits. 🤦 Will follow up with a fix.

Issue #290, opened 2020-10-23
invalid escape sequences in docstrings


jku:
This comes up in pip tests (the latter linenumber is really 109 in securesystemslib master):
```
pip0/pip/src/pip/_vendor/securesystemslib/_vendor/ed25519/ed25519.py:77: DeprecationWarning: invalid escape sequence \m
    """$= z^{-1} \mod q$, for z != 0"""

pip0/pip/src/pip/_vendor/securesystemslib/ed25519_keys.py:114: DeprecationWarning: invalid escape sequence \.
    """
```

I think docstrings can be raw strings as well so using `r"""..."""` probably fixes this. 

I'm not quite sure how to reproduce this outside of pip test suite... This does not give a warning (the `\.` is printed as `\\.`):
```
from securesystemslib import ed25519_keys
ed25519_keys.generate_public_and_private.__doc__
```
Copy-pasting the string to the interpreter does show the issue
```
>>> """'\xa2F\x99\xe0\x86\x80%\xc8\xee\x11\xb95T\xd9\...'"""
<stdin>:1: DeprecationWarning: invalid escape sequence \.
"'¢F\x99à\x86\x80%Èî\x11¹5TÙ\\...'"
```

lukpueh:
I think we can definitely just remove the byte string in `ed25519_keys.generate_public_private` as it's not particularly useful. 

In `ed25519` we can probably too, or we convert the latex(?) syntax of the formula to something that doesn't contain invalid escape sequences.

jku:
Sounds reasonable... the reason I haven't done a PR is that I'm torn on modifying `_vendor/ed25519`: it's currently a bit-by-bit "vendored" copy of the upstream  https://github.com/pyca/ed25519. Would be nice to preserve that situation.

Maybe I'll modify upstream instead and update our copy after that.

The next question of course is, how do we make sure upstream ed25519 doesn't change without us noticing -- it's not very likely as the project has not had a commit in 7 years but still... The vendoring tool does not support git and we really don't want anything complicated like git submodules. Since we don't expect changes maybe just adding this somewhere in CI would be acceptable:
```
expected_pyca_ed25519="18d0b5515c8bda6bab0148f1946845bf3027bc4f"
pyca_ed25519_master_head=$(git ls-remote https://github.com/pyca/ed25519.git master | cut -f1)
if [ "$pyca_ed25519_master_head" != "$expected_pyca_ed25519" ]; then 
[...]

PR #296, opened 2020-10-30
Remove invalid escape sequence from docstring

jku:
generate_public_and_private() docstring contains an invalid escape sequence which can lead to an annoying
    `DeprecationWarning: invalid escape sequence`
Remove the whole string as it does not seem very useful.

--- 

The related issue is #290: this fixes half of it. Updating the vendored ed25519 will fix the other instance.

### DIFF
--- a/securesystemslib/ed25519_keys.py
+++ b/securesystemslib/ed25519_keys.py
@@ -110,11 +110,8 @@ def generate_public_and_private():
   <Purpose>
     Generate a pair of ed25519 public and private keys with PyNaCl.  The public
     and private keys returned conform to
-    'securesystemslib.formats.ED25519PULIC_SCHEMA' and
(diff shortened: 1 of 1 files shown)


securesystemslib/ed25519_keys.py
```diff
@@ -111,10 +111,7 @@ def generate_public_and_private():
     Generate a pair of ed25519 public and private keys with PyNaCl.  The public
     and private keys returned conform to
     'securesystemslib.formats.ED25519PULIC_SCHEMA' and
```
lukpueh, line 113: Would you mind?
```suggestion
    'securesystemslib.formats.ED25519PUBLIC_SCHEMA' and
```

lukpueh (approved):

lukpueh:
🙇

PR #299, opened 2020-11-02
ed25519: Update to upstream master HEAD

### LINKED ISSUE #298: ed25519 upstream has new commits
It seems https://github.com/pyca/ed25519.git has new commits. Please review them, update the vendored copy if needed, and then update the expected hash in _vendor/test-ed25519-upstream.sh

jku:
Update ed2519 from 18d0b5515 to c13748e1d2.

This includes three commits that
* fixes invalid escape sequence in docstring (our issue #290)
* updates CI, lint & tox configuration (not included in our copy)
* whitespace fixes to satisfy the new linter configuration

So no functional changes but the first item should silence a
DeprecationWarning in some cases.

Fixes #298


joshuagl:
Both Python 2.7 builds failed on Travis. Did upstream make a change which isn't supported by Python 2.7? 😕

jku:
> Both Python 2.7 builds failed on Travis. Did upstream make a change which isn't supported by Python 2.7?

That seems to be my fault: The old version had an empty `__init__.py` (apparently added in sslib) and I missed that it disappeared when I wiped the contents before checking out the new version... The implicit namespacing in Python>=3.3 means that the missing file was not a problem in those versions.

I'll fix this but first I'll think if there is maybe a simple way to sure I can't mess up next time...

securesystemslib/_vendor/README.md
```diff
@@ -0,0 +1,17 @@
...
+cd securesystemslib/_vendor
+rm -rf ed25519/
+git clone git@github.com:pyca/ed25519.git
+touch ed25519/__init__.py # needed by python<3.3
+git clean -f ed25519/
+git commit -a
+```
+
+Note that this does not commit any new files (our copy does not include
+all of the upstream files). Remember to update the expected upstream
```
joshuagl, line 15:
```suggestion
all of the upstream files), any new implementation files will need to be
`git add`ed before commit. Remember to update the expected upstream
```
Is it worth making this explicit?

joshuagl (approved):
Apart from a minor question on _vendor/README.md this looks good to me, thanks!

lukpueh (approved):
Thanks for the impeccable work, @jku!

PR #304, opened 2020-11-19
Fix empty list as a default value for function arg

MVrachev:
This quote from the Google Python style guide made me realize
why empty list as a default value for an argument could be
dangerous:

"Default arguments are evaluated once at module load time.
This may cause problems if the argument is a mutable object
such as a list or a dictionary. If the function modifies the object
(e.g., by appending an item to a list), the default value is modified."

Read more here:
https://google.github.io/styleguide/pyguide.html#2123-cons

### DIFF
--- a/securesystemslib/schema.py
+++ b/securesystemslib/schema.py
@@ -878,7 +878,7 @@ class Struct(Schema):
     False
   """
 
-  def __init__(self, sub_schemas, optional_schemas=[], allow_more=False,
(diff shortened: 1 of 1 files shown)


trishankatdatadog:
It's an old Python surprise. Thanks for the fix!

trishankatdatadog (approved):

lukpueh (approved):
Good find! Thanks for the patch, @MVrachev! And hooray for the new style guide recommendation bearing fruit. :)

PR #313, opened 2020-12-10
Prepare Debian downstream release 0.18.0-1

lukpueh:
* Add changelog entry
* Remove obsolete build dep in d/control
* Bump standard version and compat level in d/control
* Update d/upstream/signing-key.asc
```
  gpg --armor --export --export-options export-minimal,export-clean\
     8BA69B87D43BE294F23E812089A2AD3C07D962E8
```


debian/changelog
```diff
@@ -1,3 +1,17 @@
...
+
+  * New upstream release includes among other things:
+    - an enhanced key generation, export and import interface,
+    - updated documentation of interface functions,
+    - and more and better tests.
+
+  * Update d/control to:
+    - remove obsolete python-dateutil dependency,
+    - bump standards version to 4.5.1,
+    - and compat to 13.
```
joshuagl, line 11: Nit:
```suggestion
    - increase compat to 13.
```
or bump?

joshuagl (approved):
FWIW this looks good to me.

debian/changelog
lukpueh: Oh, I only saw this just now. I had conceived this as a sentence, that is ...
```
 Update d/control to remove obsolete python-dateutil dependency, bump standards version to 4.5.1, and compat to 13.
```
But then I'd maybe have to change the punctuation? 

Anyways, given that 0.18.0-1 is already released, I I'll make use of my non-native-speaker bonus and leave this as is. :P

But thanks for the thorough review, @joshuagl!

PR #314, opened 2020-12-10
Drop support for Python 3.5 and add 3.9

lukpueh:
Fixes: # None

Python 3.5 has now reached its end-of-life and has been retired: https://www.python.org/dev/peps/pep-0478/

The optional (but highly recommended) 'cryptography' dependency has also just dropped support for 3.5. Continuing support for 3.5 does not seem worth the effort.  Instead we add support for the new stable Python 3.9.

This PR also updates the test configuration accordingly and recomputes the pinned dependencies.

Note, securesystemslib's dependent TUF is also dropping 3.5 support (https://github.com/theupdateframework/tuf/pull/1238) and in-toto will follow shortly. 

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -91,10 +91,10 @@ setup(
     'Programming Language :: Python :: 2',
     'Programming Language :: Python :: 2.7',
     'Programming Language :: Python :: 3',
-    'Programming Language :: Python :: 3.5',
     'Programming Language :: Python :: 3.6',
     'Programming Language :: Python :: 3.7',
     'Programming Language :: Python :: 3.8',
+    'Programming Language :: Python :: 3.9',
     'Programming Language :: Python :: Implementation :: CPython',
     'Topic :: Security',
     'Topic :: Software Development'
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -103,6 +103,7 @@
     'Source': 'https://github.com/secure-systems-lab/securesystemslib',
     'Issues': 'https://github.com/secure-systems-lab/securesystemslib/issues',
   },
+  python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4",
```
jku, line 106: So this means that any users of python 3.5 will now be offered securesystemslib 0.18 by the pypi clients even if new securesystemslib versions are released, correct?

I guess this is correct and reasonable (since cryptography does exactly that), just wanted it to be explicit.

lukpueh: > So this means that any users of python 3.5 will now be offered securesystemslib 0.18 by the pypi clients even if new securesystemslib versions are released, correct?

Good question. The way I read [the docs](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) this makes sure that pip on e.g. 3.5 will not install the next version of securesystemslib, i.e. > 0.18. Not sure if this also means that it will fall back to the latest version, i.e. 0.18, or just refuse to install. I could live with both.

Btw. I didn't add this line because I dropped 3.5, but because I saw that it was missing while dropping 3.5, in case this makes a difference. :)

joshuagl (approved):
LGTM, thanks @lukpueh !

PR #315, opened 2021-01-07
Allow to provide GPG client through env

fepitre:
When `gpg2` or `gpg` is not used at all but an alternative client/wrapper, it allows to generate `in-toto` metadata with such client. For example, in case of Qubes OS we use `qubes-gpg-client` (see for the reference: https://www.qubes-os.org/doc/split-gpg/)

### DIFF
--- a/securesystemslib/gpg/constants.py
+++ b/securesystemslib/gpg/constants.py
@@ -26,31 +26,39 @@ import securesystemslib.process as process
 
 log = logging.getLogger(__name__)
 
+
+def is_available_gnupg(gnupg):
+  gpg_version_cmd = gnupg + " --version"
+  try:
+    process.run(gpg_version_cmd, stdout=process.PIPE, stderr=process.PIPE)
+    return True
+  except OSError:
+    return False
+
+
 # By default, we allow providing GPG client through the environment
 # assuming gpg2 as default value and test if exists. Otherwise, we assume gpg
 # exists.
-GPG_COMMAND = os.environ.get('GNUPG', "gpg2")
+HAVE_GPG = True
+
+GPG_ENV_COMMAND = os.environ.get('GNUPG')
+GPG2_COMMAND = "gpg2"
+GPG1_COMMAND = "gpg"
+
+if GPG_ENV_COMMAND and is_available_gnupg(GPG_ENV_COMMAND):
+  GPG_COMMAND = GPG_ENV_COMMAND
+elif is_available_gnupg(GPG2_COMMAND):
+  GPG_COMMAND = GPG2_COMMAND
+elif is_available_gnupg(GPG1_COMMAND):
+  GPG_COMMAND = GPG1_COMMAND
+else:
+  GPG_COMMAND = ""
+  HAVE_GPG = False
+
 GPG_VERSION_COMMAND = GPG_COMMAND + " --version"
 FULLY_SUPPORTED_MIN_VERSION = "2.1.0"
-
-HAVE_GPG = True
-NO_GPG_MSG = "GPG support requires a GPG command, {} version {} or newer is" \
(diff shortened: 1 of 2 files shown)


jku:
I don't know if this is a good idea or not, just noting potential issues.

If you set `GNUPG=foo` with this patch, the code will fall back to using normal gpg if 'foo' does not work. This seems wrong since the value was explicitly set.

Also the NO_GPG_MSG could be improved... It would now say: `GPG support requires a GPG command, foo version 2.1.0 or newer is fully supported.` (if both `foo` and `gpg` did not work)

fepitre:
@jku We can improve the double check on fallback to `gpg2` then `gpg`. I don't see why someone would set `GNUPG=somethingwhichisnotsupposedtowork` because it's the purpose of setting this particular value to override current hardcoded choices.

fepitre:
I'm even thinking that the code should not try to fallback to another gpg client if the explicit provided set with GNUPG fails. Any thought?

jku:
> I'm even thinking that the code should not try to fallback to another gpg client if the explicit provided set with GNUPG fails. Any thought?

Yes, this is what I meant.

securesystemslib/gpg/constants.py
```diff
@@ -25,30 +26,41 @@
...
+GPG_ENV_COMMAND = os.environ.get('GNUPG')
+GPG2_COMMAND = "gpg2"
+GPG1_COMMAND = "gpg"
 
-  except OSError:
-    HAVE_GPG = False
+# By default, we allow providing GPG client through the environment
+# assuming gpg2 as default value and test if exists. Otherwise, we assume gpg
+# exists.
+if GPG_ENV_COMMAND:
```
fepitre, line 49: @jku what about this version?

lukpueh (approved):
Thanks for the patch, @fepitre! It looks solid to me. @SantiagoTorres, do you want to take another look?

securesystemslib/gpg/constants.py
```diff
@@ -25,30 +26,41 @@
...
+  if is_available_gnupg(GPG_ENV_COMMAND):
+    GPG_COMMAND = GPG_ENV_COMMAND
+elif is_available_gnupg(GPG2_COMMAND):
+  GPG_COMMAND = GPG2_COMMAND
+elif is_available_gnupg(GPG1_COMMAND):
+  GPG_COMMAND = GPG1_COMMAND
+
+if GPG_COMMAND:
+  HAVE_GPG = True
```
trishankatdatadog, line 58: What is this variable used for?

lukpueh: To skip tests and to fail early and gracefully in case someone calls the interface but a gpg command is not available.

I did consider suggesting its removal, now that we can use `if GPG_COMMAND` instead. But I wasn't sure if it's worth the effort. I'm fine either way.

trishankatdatadog: That's fine, but it needs a little documentation at the very least, otherwise someone reading the code will repeat the Debian OpenSSL PRNG mistake.

lukpueh, line 58: Something like this, @trishankatdatadog?
```suggestion
if GPG_COMMAND:
  # Use bool to skip tests or fail early and gracefully if no gpg is available
  HAVE_GPG = True  
```
Mind squashing that into your commits, @fepitre?

fepitre: Sure no problem. I'll do it.

fepitre: @lukpueh @trishankatdatadog I've added it and also a little note in README about GNUPG.

trishankatdatadog: LGTM, thanks both

PR #316, opened 2021-01-14
Vendoring-compatible imports

jku:
This is a sister PR for https://github.com/theupdateframework/tuf/pull/1261: it modifies all imports in securesystemlibs to be compatible with vendoring tool (see TUF PR for explanation).

Notes
* Multiple solutions were used, but usually the module was directly imported into scope (so instead of referring to `securesystemslib.settings` code now refers to `settings`). In some cases this individual items from the module are imported instead
* I've tried very hard to ensure no namespace conflicts happen by grepping for every new top level name
* This is still a draft: I have yet to test that vendoring actually works with this
* The change is surprisingly large, >500 lines

### DIFF
--- a/securesystemslib/formats.py
+++ b/securesystemslib/formats.py
@@ -78,8 +78,8 @@ import datetime
 import time
 import six
 
+from securesystemslib import exceptions
 import securesystemslib.schema as SCHEMA
-import securesystemslib.exceptions
 
 # Note that in the schema definitions below, the 'SCHEMA.Object' types allow
 # additional keys which are not defined. Thus, any additions to them will be
@@ -476,7 +476,7 @@ def datetime_to_unix_timestamp(datetime_object):
   # Raise 'securesystemslib.exceptions.FormatError' if not.
   if not isinstance(datetime_object, datetime.datetime):
     message = repr(datetime_object) + ' is not a datetime.datetime() object.'
-    raise securesystemslib.exceptions.FormatError(message)
+    raise exceptions.FormatError(message)
 
   unix_timestamp = calendar.timegm(datetime_object.timetuple())
 
@@ -554,7 +554,7 @@ def format_base64(data):
     return binascii.b2a_base64(data).decode('utf-8').rstrip('=\n ')
 
   except (TypeError, binascii.Error) as e:
-    raise securesystemslib.exceptions.FormatError('Invalid base64'
+    raise exceptions.FormatError('Invalid base64'
       ' encoding: ' + str(e))
 
 
@@ -583,7 +583,7 @@ def parse_base64(base64_string):
 
   if not isinstance(base64_string, six.string_types):
     message = 'Invalid argument: '+repr(base64_string)
-    raise securesystemslib.exceptions.FormatError(message)
+    raise exceptions.FormatError(message)
 
   extra = len(base64_string) % 4
   if extra:
@@ -594,7 +594,7 @@ def parse_base64(base64_string):
     return binascii.a2b_base64(base64_string.encode('utf-8'))
 
   except (TypeError, binascii.Error) as e:
-    raise securesystemslib.exceptions.FormatError('Invalid base64'
+    raise exceptions.FormatError('Invalid base64'
       ' encoding: ' + str(e))
 
 
@@ -661,7 +661,7 @@ def _encode_canonical(object, output_function):
       _encode_canonical(value, output_function)
     output_function("}")
   else:
-    raise securesystemslib.exceptions.FormatError('I cannot encode '+repr(object))
+    raise exceptions.FormatError('I cannot encode '+repr(object))
 
 
 def encode_canonical(object, output_function=None):
@@ -724,9 +724,9 @@ def encode_canonical(object, output_function=None):
   try:
     _encode_canonical(object, output_function)
 
-  except (TypeError, securesystemslib.exceptions.FormatError) as e:
+  except (TypeError, exceptions.FormatError) as e:
     message = 'Could not encode ' + repr(object) + ': ' + str(e)
-    raise securesystemslib.exceptions.FormatError(message)
+    raise exceptions.FormatError(message)
 
   # Return the encoded 'object' as a string.
   # Note: Implies 'output_function' is None,
--- a/securesystemslib/formats.py
+++ b/securesystemslib/formats.py
@@ -79,7 +79,7 @@ import time
 import six
 
 from securesystemslib import exceptions
-import securesystemslib.schema as SCHEMA
+from securesystemslib import schema as SCHEMA
 
 # Note that in the schema definitions below, the 'SCHEMA.Object' types allow
 # additional keys which are not defined. Thus, any additions to them will be
--- a/securesystemslib/gpg/util.py
+++ b/securesystemslib/gpg/util.py
@@ -30,8 +30,8 @@ except ImportError:
   CRYPTO = False
 
 from securesystemslib import exceptions
+from securesystemslib import process
 import securesystemslib.gpg.exceptions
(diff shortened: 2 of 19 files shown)


jku:
This is ready for review: I've tested this version with pip (it now works without patching securesystemslib) and I've read through the PR again... and decided not to modify it further (without review comments anyway).

As a whole PR there is some inconsistency (that I consciously left there but could change):
* I think it would usually make sense to import individual exceptions from `securesystemslib.exceptions` but doing that may make the PR more likely to break things: I'd rather do that in smaller PRs where possible mistakes are not lost in a sea of changes. The practical fear I have is that exceptions are sometimes not tested, so typos and name clashes could slip by without test suite noticing -- maybe linting takes care of all these cases though?
* however, I have imported smaller number of individual exceptions from other sources (from cryptography or securesystemslib.gpg) where it made sense -- this leads to some exceptions being namespaced and some not
* importing individual schemas from formats seems likewise a reasonable choice: I did not do that based on Lukas comment in the TUF bug but either way works for me

All of the changes are essentially modifying the import style, the only real changes are
* resolved a cyclical import in gpg modules
* renamed process variables to gpg_process

securesystemslib/gpg/util.py
```diff
@@ -372,9 +368,9 @@ def get_hashing_class(hash_algorithm_id):
     A pyca/cryptography hashing class
 
   """
-  supported_hashing_algorithms = [securesystemslib.gpg.constants.SHA1,
-      securesystemslib.gpg.constants.SHA256,
-      securesystemslib.gpg.constants.SHA512]
+  supported_hashing_algorithms = [constants.SHA1,
+      constants.SHA256,
+      constants.SHA512]
```
joshuagl, line 373:
```suggestion
  supported_hashing_algorithms = [constants.SHA1, constants.SHA256,
      constants.SHA512]
```

joshuagl (commented):
I don't think we need to address it in this PR, but I just noticed that the style guide says that we should not import names from a module:
> Use `import` statements for packages and modules only, not for individual classes or functions.

We will want to think about a follow-on PR that addresses the issues arising from this style guide requirement, or documenting why we are not adhering to this aspect of the style guide. Exceptions, formats, and hash are obvious places where this is an issue hash. Perhaps we should file an issue about this?

I noticed that several comments are still using fully-qualified names, especially noticeable where we have comments that state which `Exceptions` a `format` module function will throw. I do not thing this is a blocker for this PR, but we should figure out how to handle this. I think it is probably bad practice in general to state in the code comments which exceptions some code might raise, it feels like it would be an unnecessary maintenance burden.

securesystemslib/hash.py
```diff
@@ -36,7 +36,7 @@
 
 from securesystemslib import exceptions
 from securesystemslib import formats
-import securesystemslib.storage
+from securesystemslib.storage import FilesystemBackend
```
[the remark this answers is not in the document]
jku: I _think_ the common sorting is first by the source module (so `from A import ...` before `from A.B import ...`) and then by the imported module name, but I don't have a reference for that or a strong preference either way

securesystemslib/gpg/util.py
```diff
@@ -317,16 +313,16 @@ def get_version():
...
+        constants.NO_GPG_MSG)
 
-  command = securesystemslib.gpg.constants.GPG_VERSION_COMMAND
-  process = securesystemslib.process.run(command,
-      stdout=securesystemslib.process.PIPE,
-      stderr=securesystemslib.process.PIPE, universal_newlines=True)
+  command = constants.GPG_VERSION_COMMAND
+  gpg_process = process.run(command,
+      stdout=process.PIPE,
+      stderr=process.PIPE, universal_newlines=True)
```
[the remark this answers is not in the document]
jku: These fixes all seem valid: amazing how many I had missed.

I'm going to cleanup with a rebase in the end of review but will try to commit these suggestions as fixup comments for now (let's see how well that works)

jku:
> I don't think we need to address it in this PR, but I just noticed that the style guide says that we should not import names from a module:
> 
> > Use `import` statements for packages and modules only, not for individual classes or functions.
> 
> We will want to think about a follow-on PR that addresses the issues arising from this style guide requirement, or documenting why we are not adhering to this aspect of the style guide. Exceptions, formats, and hash are obvious places where this is an issue. Perhaps we should file an issue about this?

I admit I had either forgotten or misunderstood the google style guide in this. I'm not totally against doing what the style guide says (it would make the PR more consistent and having less exceptions to the style guide is better) but it does feel like a hard requirement -- not having to namespace uniquely named things that are used over and over in the importing module does improve readability. Admittedly there are cases where names aren't quite unique enough and conflicts are more likely.

The other case where importing a module seems counterproductive is when there's only one or two things in the module:
```python
from tuf.client import updater # import the module, not the class
updater = updater.Updater()    # that's not just repetitive, it's also a name clash
```

securesystemslib/hash.py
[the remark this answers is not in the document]
joshuagl: You are correct, I'd somehow missed/ignored that we were importing from a module here. Apologies for the noise.

jku:
Applied all suggestions and rebased as none of the changes were functional, just style.

Lukas, please take a look: The resolved discussion is trivial but do check the last comment for the style guide discussion.

The latest test failure is a little baffling. I _think_ it's unrelated but... have you seen this before?
```
  File "/home/travis/build/secure-systems-lab/securesystemslib/tests/test_process.py", line 121, in test_run_duplicate_streams_timeout
265    securesystemslib.process.run_duplicate_streams("python --version",
266AssertionError: TimeoutExpired not raised
```

lukpueh:
> Lukas, please take a look: The resolved discussion is trivial but do check the last comment for the style guide discussion.

Thanks, will do!

> The latest test failure is a little baffling. I think it's unrelated but... have you seen this before?

No, I have not. And as coincidence would have it, @SolidifiedRay just notified me about a different and very likely also unrelated [issue in the same process module](https://github.com/SolidifiedRay/in-toto/runs/1867612936?check_suite_focus=true#step:7:427).

Let me try if this issue here "goes away" if I restart the build, so that we can merge this PR and troubleshoot the process module separately.

lukpueh:
> Let me try if this issue here "goes away" if I restart the build ...

Looks like it did. But we should still find the cause.

joshuagl:
This PR looks good to me. I believe the places we break the style guide are pragmatic choices to minimise the number of changes and/or the potential for problematic changes (for example, reducing the number of new symbols in a scope).

securesystemslib/formats.py
```diff
@@ -79,7 +79,7 @@
 import six
 
 from securesystemslib import exceptions
-import securesystemslib.schema as SCHEMA
+from securesystemslib import schema as SCHEMA
```
lukpueh, line 82: I guess you kept `as SCHEMA` to make the commit less invasive? This might be something to note in the commit message.

lukpueh (approved):
Thanks for tackling this tedious task, @jku! LGTM. All the decisions you made make sense in their context, even if you deviate from the style guide.

FYI, I also ran pylint with import-related checks only...
```bash
pylint --disable all \
       --enable $(pylint --list-msgs | grep import | grep -Eo "[A-Z][0-9]{4}" | tr "\n" ",") \
       securesystemslib
```

... and didn't notice anything suspicious.

securesystemslib/formats.py
lukpueh: Let me know if you want to reword the commit message. Otherwise, I'll just go ahead and merge. :)

jku: Yes exactly. And I think I can live with that message if you can :)

Issue #308, opened 2020-12-08
Revise public key metadata format


lukpueh:
# Description of issue or feature request:

https://github.com/secure-systems-lab/securesystemslib/issues/251 discusses various consistency concerns related to securesystemslib (TUF/in-toto) cryptographic keys.

This is an attempt to carve (and flesh) out only those concerns related to the format of public keys as they appear in TUF/in-toto metadata, with the goal of a clear specification likely as part (or appendix) of the new [signing-spec](https://github.com/secure-systems-lab/signing-spec) and its reference implementation in `securesystemslib`. 

Below in 'Current behavior' is an exhaustive list of available values for public keys as currently implemented in `securesystemslib`. This includes rsa, ecdsa, ed25519 and spx (WIP #169) keys generated with `securesystemslib`, plus keys exported from a GnuPG keyring using `securesystemslib.gpg` or from a cryptographic token (e.g. HSM) using `securesystemslib.hsm` (WIP #229).

Related concerns and calls for action are listed further below in 'Expected behavior'.

# Current behavior:
## Public key container (mandatory fields)
```
{
  'keyid' <...>, # Optional in the spec, but included in the in-toto reference implementation
  'keytype': <...>,
  'scheme': <...>,
  'keyval': {
    'private': <...>, # Must be omitted or empty for public keys
    'public': <...>
  }
}

```
### 'keyid'
- ascii hex of canonical JSON (OLPC) of parts of the key dictionary
- RFC4880 OpenPGP Issuer Fingerprint (`gpg` keys only)
- custom value (see WIP #169 , [TAP 12](https://github.com/theupdateframework/taps/blob/master/tap12.md) and  https://github.com/theupdateframework/tuf/issues/1084)

*Note that 'keyid' is not part of the public key metadata format according to in-toto/TUF specifications and tuf reference implementation, but it is included in the in-toto reference implementation  (see [#251 PR description](https://github.com/secure-systems-lab/securesystemslib/issues/251#issue-651601785)).*
### 'keytype'
- 'rsa'
- 'ed25519'
- 'ecdsa'
- 'ecdsa-sha2-nistp(256|384)' (legacy, new keys use 'ecdsa', see #239 and #267)
- 'spx' (WIP #169)

### 'scheme'
- 'rsassa-pss-(md5|sha1|sha224|sha256|sha384|sha512)'
- 'rsa-pkcs1v15-(md5|sha1|sha224|sha256|sha384|sha512)'
[...]

Issue #271, opened 2020-09-11
Revise and automate docs with sphinx and host on readthedocs


lukpueh:
**Description of issue or feature request**:
Revise docstrings of API functions regarding contents and format (use Google Style docstring as suggested in https://github.com/secure-systems-lab/code-style-guidelines/issues/20), akin to what in-toto did in https://github.com/in-toto/in-toto/pull/369.

Needs coordination with #270

**Current behavior**:
- No clear API
- Docstrings use custom secure systems lab format

**Expected behavior**:
- Clear API via #270 
- Use Google-style quasi standard docstring format (see [napoleon sphinx extension](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google))
- Revise doc contents (aim for conciseness)

MVrachev:
Please note that when we update the docstrings of API functions we should update the
documentation regarding the `securesystemslib.keys.create_signature()` and the following comment
https://github.com/secure-systems-lab/securesystemslib/blob/a7d36681f32642fce10bc8aa051f394f8e6f0858/securesystemslib/keys.py#L709
because we are not actually supporting `ecdsa-sha2-nistp384`.
If you read `securesystemslib.ecdsa_keys.create_signature()`:
https://github.com/secure-systems-lab/securesystemslib/blob/a7d36681f32642fce10bc8aa051f394f8e6f0858/securesystemslib/ecdsa_keys.py#L173
you will see we support only support `ecdsa-sha2-nistp256`.

Issue #191, opened 2019-10-29
Revise exception taxonomy


lukpueh:
**Description of issue or feature request**:

In accordance with in-toto/in-toto#126 *grep through `raise` statements and re-assess the suitability of each raised error and its message.*

Also see these CR-comments for further motivation:
- https://github.com/in-toto/in-toto/pull/168#discussion_r161009073
- https://github.com/in-toto/in-toto/pull/257#discussion_r261180946

lukpueh:
- https://github.com/secure-systems-lab/securesystemslib/pull/288#discussion_r511396439 and https://github.com/secure-systems-lab/securesystemslib/pull/288#discussion_r511822640

lukpueh:
- https://github.com/theupdateframework/tuf/pull/1261#issuecomment-760945843 (end of comment)

lukpueh:
- https://github.com/secure-systems-lab/securesystemslib/pull/319#discussion_r580995429 -- format errors vary by key type

PR #319, opened 2021-02-03
Add the Signer interface

MVrachev:
Fixes: https://github.com/theupdateframework/tuf/issues/1263?notification_referrer_id=MDE4Ok5vdGlmaWNhdGlvblRocmVhZDE0ODExNjA4Mjg6MTYyNDY3Nzg%3D#issuecomment-772616454

Create a Signer interface with the purpose of supporting multiple signing implementations.
Read more on the issue above.

### DIFF
--- a/securesystemslib/signer.py
+++ b/securesystemslib/signer.py
@@ -1,0 +1,134 @@
+"""Signer interface and example interface implementations.
+
+The goal of this module is to provide a signing interface supporting multiple
+signing implementations and a couple of example implementations.
+
+"""
+
+import abc
+import securesystemslib.keys as sslib_keys
+from typing import Dict
+
+
+class Signature:
+    """A container class containing information about a signature.
+
+    Contains a signature and the keyid uniquely identifying the key used
+    to generate the signature.
+
+    Provides utility methods to easily create an object from a dictionary
+    and return the dictionary representation of the object.
+
+    Attributes:
+        keyid: HEX string used as a unique identifier of the key.
+        signature: HEX string representing the signature.
+
+    """
+    def __init__(self, keyid: str, sig: str):
+        self.keyid = keyid
+        self.signature = sig
+
+
+    @classmethod
+    def from_dict(cls, signature_dict: Dict) -> "Signature":
+        """Creates a Signature object from its JSON/dict representation.
+
+        Arguments:
+            signature_dict:
+                A dict containing a valid keyid and a signature.
+                Note that the fields in it should be named "keyid" and "sig"
+                respectively.
+
+        Raises:
+            KeyError: If any of the "keyid" and "sig" fields are missing from
+                the signature_dict.
+
+        Returns:
+            A "Signature" instance.
+        """
+
+        return cls(signature_dict["keyid"], signature_dict["sig"])
+
+
+    def to_dict(self) -> Dict:
+        """Returns the JSON-serializable dictionary representation of self."""
+
+        return {
+            "keyid": self.keyid,
+            "sig": self.signature
+        }
+
+
+
+class Signer:
+    """Signer interface created to support multiple signing implementations."""
+
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
+    def sign(payload: bytes) -> "Signature":
+        """Signs a given payload by the key assigned to the Signer instance.
+
+        Arguments:
+            payload: The bytes to be signed.
+
+        Returns:
+            Returns a "Signature" class instance.
+        """
+        raise NotImplementedError # pragma: no cover
+
+
+
+class SSlibSigner(Signer):
+    """A securesystemslib signer implementation.
+
+    Provides a sign method to generate a cryptographic signature with a
+    securesystemslib-style rsa, ed25519 or ecdsa private key on the instance.
+    The signature scheme is determined by the key and must be one of:
+
+    - rsa(ssa-pss|pkcs1v15)-(md5|sha1|sha224|sha256|sha384|sha512) (12 schemes)
+    - ed25519
+    - ecdsa-sha2-nistp256
+
+    See "securesystemslib.interface" for functions to generate and load keys.
+
+    Attributes:
+        key_dict:
+            A securesystemslib-style key dictionary, which includes a keyid,
+            key type, signature scheme, and the public and private key values,
+            e.g.::
+
+                {
+                    "keytype": "rsa",
+                    "scheme": "rsassa-pss-sha256",
+                    "keyid": "f30a0870d026980100c0573bd557394f8c1bbd6...",
+                    "keyval": {
+                        "public": "-----BEGIN RSA PUBLIC KEY----- ...",
+                        "private": "-----BEGIN RSA PRIVATE KEY----- ..."
+                    }
+                }
+
+            The public and private keys are strings in PEM format.
+    """
+    def __init__(self, key_dict: Dict):
+        self.key_dict = key_dict
+
+
+    def sign(self, payload: bytes) -> "Signature":
+        """Signs a given payload by the key assigned to the SSlibSigner instance.
+
+        Arguments:
+            payload: The bytes to be signed.
+
+        Raises:
+            securesystemslib.exceptions.FormatError: Key argument is malformed.
+            securesystemslib.exceptions.CryptoError, \
+                securesystemslib.exceptions.UnsupportedAlgorithmError:
+                Signing errors.
+
+        Returns:
+            Returns a "Signature" class instance.
+        """
+
+        sig_dict = sslib_keys.create_signature(self.key_dict, payload)
+        return Signature(**sig_dict)
--- a/tests/test_signer.py
+++ b/tests/test_signer.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python
+
+"""Test cases for "signer.py". """
+
+import unittest
+import securesystemslib.formats
+import securesystemslib.keys as KEYS
+from securesystemslib.signer import Signature, SSlibSigner
+from securesystemslib.exceptions import FormatError, UnsupportedAlgorithmError
+
+class TestSSlibSigner(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.rsakey_dict = KEYS.generate_rsa_key()
+        cls.ed25519key_dict = KEYS.generate_ed25519_key()
+        cls.ecdsakey_dict = KEYS.generate_ecdsa_key()
+        cls.DATA_STR = "SOME DATA REQUIRING AUTHENTICITY."
+        cls.DATA = securesystemslib.formats.encode_canonical(
+                cls.DATA_STR).encode("utf-8")
+
+
+    def test_sslib_sign(self):
+        dicts = [self.rsakey_dict, self.ecdsakey_dict, self.ed25519key_dict]
+        for scheme_dict in dicts:
+            # Test generation of signatures.
+            sslib_signer = SSlibSigner(scheme_dict)
+            sig_obj = sslib_signer.sign(self.DATA)
+
+            # Verify signature
+            verified = KEYS.verify_signature(scheme_dict, sig_obj.to_dict(),
+                    self.DATA)
+            self.assertTrue(verified, "Incorrect signature.")
+
+            # Removing private key from "scheme_dict".
+            private = scheme_dict["keyval"]["private"]
+            scheme_dict["keyval"]["private"] = ""
+            sslib_signer.key_dict = scheme_dict
+
+            with self.assertRaises((ValueError, FormatError)):
+                sslib_signer.sign(self.DATA)
+
+            scheme_dict["keyval"]["private"] = private
+
+            # Test for invalid signature scheme.
+            valid_scheme = scheme_dict["scheme"]
+            scheme_dict["scheme"] = "invalid_scheme"
+            sslib_signer = SSlibSigner(scheme_dict)
+
+            with self.assertRaises((UnsupportedAlgorithmError, FormatError)):
+                sslib_signer.sign(self.DATA)
+
+            scheme_dict["scheme"] = valid_scheme
+
+
+    def test_signature_from_to_dict(self):
+        signature_dict = {
+            "sig": "30460221009342e4566528fcecf6a7a5d53ebacdb1df151e242f55f8775883469cb01dbc6602210086b426cc826709acfa2c3f9214610cb0a832db94bbd266fd7c5939a48064a851",
+            "keyid": "11fa391a0ed7a447cbfeb4b2667e286fc248f64d5e6d0eeed2e5e23f97f9f714"
+        }
+        sig_obj = Signature.from_dict(signature_dict)
+
+        self.assertDictEqual(signature_dict, sig_obj.to_dict())
+
+
+# Run the unit tests.
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/test_signer.py
+++ b/tests/test_signer.py
@@ -2,11 +2,26 @@
 
 """Test cases for "signer.py". """
 
+import sys
+import unittest
+
 import unittest
 import securesystemslib.formats
 import securesystemslib.keys as KEYS
-from securesystemslib.signer import Signature, SSlibSigner
 from securesystemslib.exceptions import FormatError, UnsupportedAlgorithmError
+
+# TODO: Remove case handling when fully dropping support for versions < 3.6
+IS_PY_VERSION_SUPPORTED = sys.version_info >= (3, 6)
+
+# Use setUpModule to tell unittest runner to skip this test module gracefully.
+def setUpModule():
+    if not IS_PY_VERSION_SUPPORTED:
+        raise unittest.SkipTest("requires Python 3.6 or higher")
+
+# Since setUpModule is called after imports we need to import conditionally.
+if IS_PY_VERSION_SUPPORTED:
+    from securesystemslib.signer import Signature, SSlibSigner
+
 
 class TestSSlibSigner(unittest.TestCase):
 


tests/test_signer.py
```diff
@@ -0,0 +1,96 @@
...
+    cls.DATA = securesystemslib.formats.encode_canonical(cls.DATA_STR).encode(
+        'utf-8')
+    cls.FORMAT_ERROR_MSG = 'securesystemslib.exceptions.FormatError was raised!' + \
+        '  Check object\'s format.'
+
+
+  def test_sslib_sign(self):
+    # Test generation of RSA signatures.
+    sslib_signer = SSlibSigner(self.rsakey_dict)
+    rsa_sig_obj = sslib_signer.sign(self.DATA)
```
trishankatdatadog, line 43: I like how simple it looks to use, judging by the code so far! 👏🏽

securesystemslib/signer.py
```diff
@@ -0,0 +1,131 @@
...
+
+    <Arguments>
+      payload: bytes to be signed
+
+    <Returns>
+      Returns a "Signature" class instance containing the signature and the
+      metadata related to the key needed for verification (e.g. keyid).
+    """
+    signiture_dict = sslib_keys.create_signature(self.key_dict, payload)
+    return Signature(signiture_dict['keyid'], signiture_dict['sig'])
```
trishankatdatadog, line 120: Hmm, why this indirection? To keep backwards-compatibility for existing functions?

MVrachev: What do you mean by `indirection`? 

PS: I will push a new version of this with the `**` syntax. It looks a lot cleaner.

MVrachev:
This interface is ready for review @lukpueh!
I will add another pr for GPGSigner.

tests/test_signer.py
```diff
@@ -0,0 +1,102 @@
...
+
+    # Removing private key from 'ecdsakey_dict' - should raise a TypeError.
+    private = self.ecdsakey_dict['keyval']['private']
+    self.ecdsakey_dict['keyval']['private'] = ''
+
+    self.assertRaises(ValueError, sslib_signer.sign, self.DATA)
+
+    # Supplying an incorrect number of arguments.
+    self.assertRaises(TypeError, sslib_signer.sign)
+    self.ecdsakey_dict['keyval']['private'] = private
```
lukpueh, line 97: You could DRY the test e.g. with a for loop over `[self.rsakey_dict, self.ecdsakey_dict]`. Then you could easily throw the currently unused `self.ed25519key_dict` in the mix without extra effort. :)

securesystemslib/signer.py
```diff
@@ -0,0 +1,127 @@
...
+  <Attributes>
+    keyid: HEX string used as a unique identifier of the key.
+    signature: HEX string representing the signature.
+  """
+  def __init__(self, keyid, sig):
+    self.keyid = keyid
+    self.signature = sig
+
+
+  def to_dict(self):
```
lukpueh, line 39: If you provide a `to_dict`, I suggest you also provide a `from_dict` classmethod.

lukpueh (changes requested):
Nice work, @MVrachev. I only pointed out some minor issues, most of which are not really your fault but seem to stem from inaccurate code documentation in this repo.

securesystemslib/signer.py
```diff
@@ -0,0 +1,127 @@
...
+      to the Signer instance.
+
+    <Arguments>
+      payload: bytes to be signed
+
+    <Returns>
+      Returns a "Signature" class instance containing the signature and the
+      the keyid which uniquely identifies the key used for signature generation.
+    """
+    pass
```
lukpueh, line 73: This should raise `NotImplementedError`

tests/test_signer.py
```diff
@@ -0,0 +1,102 @@
...
+    sslib_signer = SSlibSigner(self.rsakey_dict)
+    rsa_sig_obj = sslib_signer.sign(self.DATA)
+
+    # Verify rsa signature
+    verified = KEYS.verify_signature(self.rsakey_dict,
+        rsa_sig_obj.to_dict(), self.DATA)
+
+    # Verify rsa signature
+    verified = KEYS.verify_signature(self.rsakey_dict,
+        rsa_sig_obj.to_dict(), self.DATA)
```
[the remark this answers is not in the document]
MVrachev: A mistake it seems. I will remove it.

tests/test_signer.py
```diff
@@ -0,0 +1,102 @@
...
+    # Creating a signature for 'DATA'.
+    ecdsa_sig = sslib_signer.sign(self.DATA)
+
+    # Verify rsa signature
+    verified = KEYS.verify_signature(self.ecdsakey_dict,
+        ecdsa_sig.to_dict(), self.DATA)
+
+    # Verify rsa signature
+    verified = KEYS.verify_signature(self.ecdsakey_dict,
+        ecdsa_sig.to_dict(), self.DATA)
```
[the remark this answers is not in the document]
MVrachev: A mistake it seems. I will remove it.

tests/test_signer.py
```diff
@@ -0,0 +1,102 @@
...
+    self.assertRaises(securesystemslib.exceptions.UnsupportedAlgorithmError,
+        sslib_signer.sign, self.DATA)
+    self.rsakey_dict['scheme'] = valid_scheme
+
+    # Removing private key from 'rsakey_dict' - should raise a TypeError.
+    private = self.rsakey_dict['keyval']['private']
+    self.rsakey_dict['keyval']['private'] = ''
+    sslib_signer.key_dict = self.rsakey_dict
+
+    self.assertRaises(ValueError, sslib_signer.sign, self.DATA)
```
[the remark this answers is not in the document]
MVrachev: A mistake. It should raise a `ValueError` for the missing value.
Will fix it.

MVrachev:
Updated the pr.
- I added a new commit fixing the documentation for `secureystemslib.keys.create_signature()`. 
- Added a `Signature.from_dict()` and a unit test for it.
- Fixed the wording in the documentation
- Rewrote the test and made it simpler.

@lukpueh sorry for the obvious mistakes in the tests. I copied a lot from `tests/test_keys.py` and it seems I did it badly. :D

securesystemslib/signer.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+  @classmethod
+  def from_dict(cls, signature_dict):
+    """
+    <Purpose>
+      Creates a Signature object from its JSON/dict representation.
+
+    <Arguments>
+      signature_dict: A dict containing that should contain valid keyid
```
lukpueh, line 46:
```suggestion
      signature_dict: A dict that contains a valid keyid
```
?

securesystemslib/signer.py
```diff
@@ -0,0 +1,144 @@
...
+    <Purpose>
+      Abstract function used for signing a given payload by the key assigned
+      to the Signer instance.
+
+    <Arguments>
+      payload: bytes to be signed
+
+    <Returns>
+      Returns a "Signature" class instance containing the signature and the
+      the keyid which uniquely identifies the key used for signature generation.
```
lukpueh, line 86:
```suggestion
     keyid which uniquely identifies the key used for signature generation.
```

securesystemslib/signer.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+
+class SSlibSigner(Signer):
+  """
+    <Purpose>
+      Securesystemslib default implementation of the "Signer" interface.
+      With this implementation the following signature schemes are supported:
+
+      'RSASSA-PSS'
```
lukpueh, line 98: We actually support pkcs1v15 as well. And I think we should use the securesystemslib scheme strings here, as e.g. defined in formats:

```python
 RSA_SCHEME_SCHEMA = SCHEMA.OneOf([
  SCHEMA.RegularExpression(r'rsassa-pss-(md5|sha1|sha224|sha256|sha384|sha512)'),
  SCHEMA.RegularExpression(r'rsa-pkcs1v15-(md5|sha1|sha224|sha256|sha384|sha512)')])
```

tests/test_signer.py
```diff
@@ -0,0 +1,87 @@
...
+    cls.ecdsakey_dict = KEYS.generate_ecdsa_key()
+    cls.DATA_STR = 'SOME DATA REQUIRING AUTHENTICITY.'
+    cls.DATA = securesystemslib.formats.encode_canonical(cls.DATA_STR).encode(
+        'utf-8')
+    cls.FORMAT_ERROR_MSG = 'securesystemslib.exceptions.FormatError was raised!' + \
+        '  Check object\'s format.'
+
+
+  def test_sslib_sign(self):
+    for scheme_dict in [self.rsakey_dict, self.ecdsakey_dict]:
```
lukpueh, line 41: ping `ed25519key_dict`?

lukpueh (changes requested):
Thanks, @MVrachev! This is getting better and better. Here are a few more suggestions.

securesystemslib/keys.py
```diff
@@ -713,9 +716,9 @@ def create_signature(key_dict, data):
     sig, scheme = securesystemslib.ed25519_keys.create_signature(
         public, private, data, scheme)
 
-  # Continue to support keytypes of ecdsa-sha2-nistp256 and ecdsa-sha2-nistp384
+  # Continue to support keytypes of ecdsa-sha2-nistp256
   # for backwards compatibility with older securesystemslib releases
-  elif keytype in ['ecdsa', 'ecdsa-sha2-nistp256', 'ecdsa-sha2-nistp384']:
+  elif keytype in ['ecdsa', 'ecdsa-sha2-nistp256',]:
```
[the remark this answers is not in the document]
MVrachev: If the `keytype` is `ecdsa-sha2-nistp384` the execution will continue with the call of `securesystemslib.ecdsa_keys.create_signature(public, private, data, scheme)`. 
I am wondering what will be the value of the `scheme` on this call?
Because the `scheme` argument is the one used to determine if the `ecdsa-sha2-nistp384` signature generation is actually supported.

If the scheme is again  `ecdsa-sha2-nistp384`, then there is no sense in keeping this comment and check and how to drop support for `ecdsa-sha2-nistp256` and `ecdsa-sha2-nistp384` could be decided independently of this.

securesystemslib/keys.py
```diff
@@ -684,7 +687,7 @@ def create_signature(key_dict, data):
 
   # Signing the 'data' object requires a private key. Signing schemes that are
   # currently supported are: 'ed25519', 'ecdsa-sha2-nistp256',
-  # 'ecdsa-sha2-nistp384' and rsa schemes defined in
+  # and rsa schemes defined in
```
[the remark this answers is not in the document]
MVrachev: I updated it above adding the `ecdsa-sha2-nistp256` support.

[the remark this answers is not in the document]
MVrachev: Still, I added 
`rsassa-pss-(md5|sha1|sha224|sha256|sha384|sha512)`
`rsa-pkcs1v15-(md5|sha1|sha224|sha256|sha384|sha512)`
in the doc.

MVrachev:
I rebased and squashed the commits where I address @lukpueh  comments into the appropriate commits. 
I added the `ed25519key_dict` into the tests and fixed another group of docstring mistakes.

securesystemslib/keys.py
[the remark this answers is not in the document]
MVrachev: I decided to not update the commit now until we decide what to do next.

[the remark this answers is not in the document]
trishankatdatadog: Nitpick: I recommend using a set instead of a list to capture the values to check...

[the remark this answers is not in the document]
lukpueh: > If the scheme is again ecdsa-sha2-nistp384, then there is no sense in keeping this comment and check and how to drop support for ecdsa-sha2-nistp256 and ecdsa-sha2-nistp384 could be decided independently of this.

Not sure I follow your reasoning, but I don't think this particular line of code is worth any further consideration without taking into account the big picture in #308 or the even bigger picture in https://github.com/secure-systems-lab/signing-spec/issues/9.

lukpueh (changes requested):
I think we should use the same strategy to move away from legacy code style as we do in TUF (see [ADR 5](https://github.com/theupdateframework/tuf/blob/develop/docs/adr/0005-use-google-python-style-guide.md)). So given that you are adding completely new modules I suggest you also follow the new [style guideline](https://github.com/secure-systems-lab/code-style-guidelines/blob/master/python.md), especially in regards to code documentation. What do you think?

MVrachev:
> I think we should use the same strategy to move away from legacy code style as we do in TUF (see [ADR 5](https://github.com/theupdateframework/tuf/blob/develop/docs/adr/0005-use-google-python-style-guide.md)). So given that you are adding completely new modules I suggest you also follow the new [style guideline](https://github.com/secure-systems-lab/code-style-guidelines/blob/master/python.md), especially in regards to code documentation. What do you think?

I thought I should follow the style guidelines in the other parts in securesytemslib.
My reasoning was that in tuf where we write the new API from scratch we should use the new style guidelines, but for files that are in the old folders, we should use the older style guidelines. 

It seems I am wrong. I will fix it on Monday.

securesystemslib/keys.py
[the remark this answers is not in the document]
joshuagl: > Long story short, instead of making this unrelated change now, I'd rather think about how to best deprecate both 'ecdsa-sha2-nistp256' and 'ecdsa-sha2-nistp384' as keytypes.

Yes please, let's figure out how to deprecate both 'ecdsa-sha2-nistp256' and 'ecdsa-sha2-nistp384' as keytypes.

> @joshuagl, do you want to weigh in as the author of that last change?

With or without this change we don't support generating signatures with a keytype of ecdsa-sha2-nistp384, so I don't feel too strongly about how it fails. It was a mistake to introduce the change in #228, but let's not worry too much about trying to undo it now and focus on the bigger picture items and deprecating the overloaded keytypes.

lukpueh:
Your reasoning makes perfect sense, @MVrachev, especially if you read ADR 5 together with [ADR 3](https://github.com/theupdateframework/tuf/blob/develop/docs/adr/0003-where-to-develop-TUF-1-0-0.md). However, ADR 5 alone even prescribes the new style for "new files" and not only for "new files in a certain new directory".

But regardless of ADR 5, I think it makes a lot of sense to use the new docstring format everywhere especially for classes / functions that should be public API (I even adopted it in an existing module recently, see #288). 

I am less passionate about most of the other style recommendations, but to make things easier I would just follow the "new style for new files" rule of thumb.

securesystemslib/keys.py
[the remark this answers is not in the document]
MVrachev: Okay, then I will leave this commit for now.

MVrachev:
I updated the pr with the new style guidelines.

MVrachev:
I had to force push to resolve conflicts.
I updated the `Update doc for ecdsa-sha2-nistp384 schema` commit to using a `keys_types_set` as Trishank suggested.

securesystemslib/signer.py
```diff
@@ -0,0 +1,137 @@
...
+            An example RSA key dict has the form:
+
+            {'keytype': 'rsa',
+            'scheme': 'rsassa-pss-sha256',
+            'keyid': 'f30a0870d026980100c0573bd557394f8c1bbd6...',
+            'keyval': {'public': '-----BEGIN RSA PUBLIC KEY----- ...',
+                        'private': '-----BEGIN RSA PRIVATE KEY----- ...'}}
+
+            The public and private keys are strings in PEM format.
+    """
```
lukpueh, line 118: I tried to clarify the parts about key types and schemes a bit and ended up re-writing the entire docstring. What do you think?

```suggestion
    """A securesystemslib signer implementation.

    Provides a sign method to generate a cryptographic signature with a
    securesystemslib-style rsa, ed25519 or ecdsa private key on the instance.
    The signature scheme is determined by the key and must be one of:

    - rsa(ssa-pss|pkcs1v15)-(md5|sha1|sha224|sha256|sha384|sha512) (12 schemes)
    - ed25519
    - ecdsa-sha2-nistp256

    See 'securesystemslib.interface' for functions to generate and load keys.

    Attributes:
        key_dict:
            A securesystemslib-style key dictionary, which includes a keyid,
            key type, signature scheme, and the public and private key values,
            e.g.::

                {
                    "keytype": "rsa",
                    "scheme": "rsassa-pss-sha256",
                    "keyid": "f30a0870d026980100c0573bd557394f8c1bbd6...",
                    "keyval": {
                        "public": "-----BEGIN RSA PUBLIC KEY----- ...",
                        "private": "-----BEGIN RSA PRIVATE KEY----- ..."
                    }
                }

    """
```

Note that the code block now also renders nicely with sphinx.

securesystemslib/keys.py
```diff
@@ -706,9 +711,9 @@ def create_signature(key_dict, data):
     private = binascii.unhexlify(private.encode('utf-8'))
     sig, scheme = ed25519_keys.create_signature(public, private, data, scheme)
 
-  # Continue to support keytypes of ecdsa-sha2-nistp256 and ecdsa-sha2-nistp384
+  # Continue to support keytypes of ecdsa-sha2-nistp256
   # for backwards compatibility with older securesystemslib releases
-  elif keytype in ['ecdsa', 'ecdsa-sha2-nistp256', 'ecdsa-sha2-nistp384']:
+  elif keytype in key_types_set:
```
lukpueh, line 716: I'm not particularly happy about the unrelated changes in this module, at least not as part of this PR.
Even with your fixes the docstring is still inaccurate (e.g. *"cryptography library specified in 'settings'"*) and could be more concise (see my suggestions in the corresponding `SSlibSigner` docstring). And I'd also have a comment or two on the changes in the function body.

Would you mind just removing 67699608ea017c0a2c9c1dca51f03c2b589bbc53 from this PR? You can instead submit a new issue or add a comment e.g. to https://github.com/secure-systems-lab/securesystemslib/issues/271.

lukpueh (changes requested):
Hey Martin, this looks great, I only left a few comments regarding docstrings (format), and I'd like to kindly ask you to remove the changes you made in `securesystemslib/keys.py` from this PR. :)

securesystemslib/keys.py
MVrachev: Okay, I will remove the commit.
Added a comment about this: https://github.com/secure-systems-lab/securesystemslib/issues/271#issuecomment-782144697

securesystemslib/signer.py
```diff
@@ -0,0 +1,137 @@
...
+"""
+
+import abc
+import securesystemslib.keys as sslib_keys
+from typing import Dict
+
+
+class Signature:
+    """Storage class containing information about a signature and the keyid
+        uniquely identifying the key used to generate the signature.
```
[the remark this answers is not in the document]
MVrachev: I will update the class docstring here, but I am not sure there is something else to add to the `SSlibSigner` docstring, so
the class summary is actually the class docstring.
The rest information related to the functionality of the class is into the class method `sign`.

securesystemslib/signer.py
MVrachev: Looks great. I will use your suggestion.

MVrachev:
Updated the pr addressing @lukpueh comments around the docstrings and tried my best to follow the style guidelines. 
I decided to squash the changes into the `Add the Signer interface and the SSlibSigner` commit in order to prepare the pr for merge.

PS: Sorry that I force-pushed so many times. I always forgot to add something.

securesystemslib/signer.py
```diff
@@ -0,0 +1,137 @@
...
+Started:
+    Januart 27, 2021.
+
+Copyright:
+    See LICENSE for licensing information.
+
+Purpose:
+    The goal of this module is to provide signing interface supporting multiple
+    signing implementations.
+"""
```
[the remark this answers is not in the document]
lukpueh: > Same comment applies to the module docstring in test_signer.

ping

MVrachev:
Rebased and update the `test_signature.py` file because of https://github.com/secure-systems-lab/securesystemslib/pull/319#discussion_r580063012.

tests/test_signer.py
```diff
@@ -0,0 +1,88 @@
...
+
+import sys
+import unittest
+
+import unittest
+import securesystemslib.formats
+import securesystemslib.keys as KEYS
+from securesystemslib.exceptions import FormatError, UnsupportedAlgorithmError
+
+# TODO: Remove case handling when fully dropping support for versions >= 3.6
```
lukpueh, line 13: Looks like you copy-pasted a mistake I made in tuf's test_api.py :P
```suggestion
# TODO: Remove case handling when fully dropping support for versions < 3.6
```
```suggestion
# TODO: Remove case handling when fully dropping support for versions >= 3.6
```

tests/test_signer.py
```diff
@@ -0,0 +1,88 @@
...
+            verified = KEYS.verify_signature(scheme_dict, sig_obj.to_dict(),
+                    self.DATA)
+            self.assertTrue(verified, "Incorrect signature.")
+
+            # Removing private key from 'scheme_dict' - should raise a ValueError.
+            private = scheme_dict['keyval']['private']
+            scheme_dict['keyval']['private'] = ''
+            sslib_signer.key_dict = scheme_dict
+
+            with self.assertRaises((ValueError, FormatError)):
```
lukpueh, line 55: Above you say:
> Removing private key from 'scheme_dict' - should raise a ValueError

Why do you assert for both `(ValueError, FormatError)`? Is this different for different keys? 🤦

tests/test_signer.py
```diff
@@ -0,0 +1,88 @@
...
+                sslib_signer.sign(self.DATA)
+
+            scheme_dict['keyval']['private'] = private
+
+            # Test for invalid signature scheme.
+            valid_scheme = scheme_dict['scheme']
+            scheme_dict['scheme'] = 'invalid_scheme'
+            sslib_signer = SSlibSigner(scheme_dict)
+
+            with self.assertRaises((UnsupportedAlgorithmError, FormatError)):
```
lukpueh, line 65: Is this also different for different keys? 🤦

tests/test_signer.py
```diff
@@ -0,0 +1,88 @@
...
+                sslib_signer.sign(self.DATA)
+
+            scheme_dict['scheme'] = valid_scheme
+
+            # Supplying an incorrect number of arguments.
+            with self.assertRaises(TypeError):
+                sslib_signer.sign()
```
lukpueh, line 72: What's the reason for this test on this particular method? It seems like we could do this with any function but we don't. I think we don't need it here either.

tests/test_signer.py
```diff
@@ -0,0 +1,88 @@
...
+        dicts = [self.rsakey_dict, self.ecdsakey_dict, self.ed25519key_dict]
+        for scheme_dict in dicts:
+            # Test generation of signatures.
+            sslib_signer = SSlibSigner(scheme_dict)
+            sig_obj = sslib_signer.sign(self.DATA)
+
+            # Verify signature
+            verified = KEYS.verify_signature(scheme_dict, sig_obj.to_dict(),
+                    self.DATA)
+            self.assertTrue(verified, "Incorrect signature.")
```
lukpueh, line 48: Minor nit: This module has a mix of single and double quotes. Please choose one consistently, [preferably double quotes](https://github.com/secure-systems-lab/code-style-guidelines/blob/master/python.md#quotes).

tests/test_signer.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+    def test_signature_from_to_json(self):
+        signature_dict = {
+            "sig": "30460221009342e4566528fcecf6a7a5d53ebacdb1df151e242f55f8775883469cb01dbc6602210086b426cc826709acfa2c3f9214610cb0a832db94bbd266fd7c5939a48064a851",
+            "keyid": "11fa391a0ed7a447cbfeb4b2667e286fc248f64d5e6d0eeed2e5e23f97f9f714"
+        }
+        sig_obj = Signature.from_dict(signature_dict)
+
+        self.assertEqual(signature_dict, sig_obj.to_dict())
```
lukpueh, line 83: Minor nit: It's better to use `assertDictEqual`, because it has a more informative error message.

lukpueh (approved):
Thanks for fixing up the docstrings, @MVrachev! I have one last set of comments in the new test module. Please fix and we are good to merge. :)

tests/test_signer.py
MVrachev: Yes, sadly.
For `rsa` keys   if there is no private key an `ValueError` is raised, 
for `ecdsa` and `ed25519` is a `FormatError`.

I will fix the comment above which says there should be a `ValueError`.

lukpueh: > I was thinking should I create an issue or pr for this, but given that I am researching how to replaces the securesystemslib schemas probably it doesn't make sense.

I agree. I'll still make a note on #191.

tests/test_signer.py
MVrachev: Yes, for `rsa` we have a check in `securesystemslib.keys.create_signature()` which we don't have in for `ed25519` and `ecdsa`:
image

but when we call `ed25519_keys.create_signature` or `ecdsa_keys.create_signature` there is a check inside those
functions for the `scheme` argument and that's how a FormatError is received. 

Do you think I should fix those problems here in this or?
Or in a different pr or issue?
Or even should I fix it now given that I am investigating how to replace the schemas?

lukpueh: Thanks for the detailed report! Let's just add this to the wishlist in #191.

tests/test_signer.py
MVrachev: Okay, will remove it.

MVrachev:
Updated the pr:
- replaced ' with " in both `signer.py` and `test_signer.py`.
- fixed some comments 
- addressed all other Lukas comments.

I will need time to get used to the new code style guidelines.
Sorry for creating so much work in reviewing for you @lukpueh.

lukpueh:
> I will need time to get used to the new code style guidelines.
Sorry for creating so much work in reviewing for you @lukpueh.

No worries, I need to look things up in the new guidelines too. ;)

Issue #297, opened 2020-10-30
Consider switching to GitHub Actions for automated testing


lukpueh:
**Description of issue or feature request**:
We switched from automated testing on Travis CI (Linux) and Appveyor (Windows) to GitHub Actions in our sister-project `in-toto-golang` (see https://github.com/in-toto/in-toto-golang/pull/72), mostly due to availability concerns about Appveyor and because it seems appealing to configure only one service for all platforms.

Let's consider switching here too and testing on more platforms than we currently do.

**Current behavior**:
Use Travis CI for automated tests on Linux (no tests for Windows and Mac OS)

**Expected behavior**:
Use GitHub Actions instead and also test on Windows and Mac OS.

joshuagl:
Given how slow Travis has been this week and how slow AppVeyor has always been, as well as the inconsistency between CI setup between the two platforms, I suggest we don't even need to consider this. Let's do it!

I think GitHub actions might be a bit easier to replicate/debug locally, instead of having to push a commit and see what errors CI throws. Does that match your experience @jku?

lukpueh:
Related: in-toto/in-toto#407 and https://github.com/theupdateframework/tuf/issues/1195

jku:
> I think GitHub actions might be a bit easier to replicate/debug locally, instead of having to push a commit and see what errors CI throws.

Uh, I was hoping that would be the case but local execution is not something Github explicitly supports and getting it working seemed too much hassle for at least the one workflow (https://github.com/nektos/act). 

This worked reasonably well as a loop though (not quite REPL but...): 
* edit the workflow file directly in the github ui in a testing branch
* trigger the action _from that branch_ manually
* check the results

The only caveat is that the workflow must exist in main branch: otherwise it does not seem to appear in the UI at all.

lukpueh:
Thanks for the report, @jku. `nektos/act` is probably worth exploring at some point (maybe an SSLAB stundet task?) 
In the meanwhile I think the 1-service-for-all-platforms + reliability arguments still make a strong case for the switch.

jku:
oh and a major benefit I didn't mention: testing actions in a fork is simple. In my case I just enabled issues in my fork and could then test issue opening from actions there without creating test issues in the project itself

lukpueh:
Also consider injection vulnerabilities reported in https://crbug.com/project-zero/2070

jku:
Considered and implemented in PR #324 

Related still open issues: #325, #326

PR #322, opened 2021-02-09
Pin cryptography 3.3.x on python2

jku:
cryptography no longer supports python2: Before we drop python2 support let's pin the cryptography version -- but only for python2.

This includes #321.

I have no idea what dependabot really does when it runs so **don't know if this could break dependabot or not** -- the example script in requirements.txt does not seem to quite match what dependabot must be doing as pip-compile now produces multiline annotations which breaks the sorting in the example script...

Is there a way to test dependabot somehow?

### DIFF
--- a/requirements-pinned.txt
+++ b/requirements-pinned.txt
@@ -1,6 +1,6 @@
 cffi==1.14.4              # via cryptography, pynacl
 colorama==0.4.4
-cryptography==3.3.1
+cryptography==3.4.3
 enum34==1.1.10            ; python_version < "3" # via cryptography
 ipaddress==1.0.23         ; python_version < "3" # via cryptography
 pycparser==2.20           # via cffi
--- a/requirements-pinned.txt
+++ b/requirements-pinned.txt
@@ -1,6 +1,7 @@
 cffi==1.14.4              # via cryptography, pynacl
 colorama==0.4.4
-cryptography==3.4.3
+cryptography==3.3.2       ; python_version < "3"
+cryptography==3.4.3       ; python_version >= "3"
 enum34==1.1.10            ; python_version < "3" # via cryptography
 ipaddress==1.0.23         ; python_version < "3" # via cryptography
 pycparser==2.20           # via cffi
--- a/requirements-pinned.txt
+++ b/requirements-pinned.txt
@@ -1,7 +1,7 @@
 cffi==1.14.4              # via cryptography, pynacl
 colorama==0.4.4
-cryptography==3.3.2       ; python_version < "3"
-cryptography==3.4.3       ; python_version >= "3"
+cryptography==3.3.2       ; python_version < "3" # cryptography < 3.4 for python2 compat
+cryptography==3.4.4       ; python_version >= "3"
 enum34==1.1.10            ; python_version < "3" # via cryptography
 ipaddress==1.0.23         ; python_version < "3" # via cryptography
 pycparser==2.20           # via cffi
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,7 +31,8 @@
 # 3. Use this command to remove per-version files
 #    `rm requirements-?.?.txt`
 #
-cryptography
+cryptography; python_version >= '3'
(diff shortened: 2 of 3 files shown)


jku:
The test failure is missing rustc (cryptography 3.4 dependency) which Lukas suggested should be fixed by #407

lukpueh:
> Is there a way to test dependabot somehow?

securesystemslib still uses the old non-GitHub-native "Dependabot preview", where you can trigger a check/update via the https://app.dependabot.com/ web UI.

lukpueh:
Could you also update setup.py?

lukpueh:
And should we maybe just disable Python 3.6 tests until we move to GitHub Actions? #297

jku:
I modified setup.py -- I think that's correct but I am really not a python packager.

setup.py
```diff
@@ -107,7 +107,8 @@
   install_requires = ['six>=1.11.0', 'subprocess32; python_version < "3"'],
   extras_require = {
       'colors': ['colorama>=0.3.9'],
-      'crypto': ['cryptography>=2.6'],
+      'crypto:python_version < "3"': ['cryptography>=2.6,<3.4'],
+      'crypto:python_version >= "3.0"': ['cryptography>=2.6'],
```
joshuagl, line 111: the version specifiers elsewhere are "3", but here it's "3.0" – any significance to that? Or just an extra level of precision (comparing to major.minor, instead of just major)?

jku: oops, that was not supposed to be there -- I had to experiment to find the format that setuptools would accept, that's just a leftover from the testing

requirements.txt
```diff
@@ -31,7 +31,8 @@
 # 3. Use this command to remove per-version files
 #    `rm requirements-?.?.txt`
 #
-cryptography
+cryptography; python_version >= '3'
+cryptography < 3.4; python_version < '3'
```
lukpueh, line 35:
```suggestion
cryptography >=3.3.2; python_version >= '3'
cryptography >=3.3.2, < 3.4; python_version < '3'
```
... for CVE-2020-36242?

jku:
Marking this ready for review as moving CI to Github fixes the testing issue... but my dependabot comment still applies (I don't think I can use https://app.dependabot.com/).

requirements-pinned.txt
```diff
@@ -1,6 +1,7 @@
 cffi==1.14.4              # via cryptography, pynacl
 colorama==0.4.4
-cryptography==3.3.1
+cryptography==3.3.2       ; python_version < "3"
```
lukpueh, line 3: Dependabot will likely soon submit a PR to bump this version, which we will need to ignore. But I don't want to ignore cryptography bumps altogether and I don't think we can configure Dependabot to only bump cryptography for python_version >= "3".

The good news is, once we drop Py2 we don't have to worry about it anymore.

setup.py
```diff
@@ -107,7 +107,8 @@
   install_requires = ['six>=1.11.0', 'subprocess32; python_version < "3"'],
   extras_require = {
       'colors': ['colorama>=0.3.9'],
-      'crypto': ['cryptography>=2.6'],
+      'crypto:python_version < "3"': ['cryptography>=2.6,<3.4'],
+      'crypto:python_version >= "3"': ['cryptography>=2.6'],
```
lukpueh, line 111:
```suggestion
      'crypto:python_version < "3"': ['cryptography>=3.3.2,<3.4'],
      'crypto:python_version >= "3"': ['cryptography>=3.3.2'],
```
... for CVE-2020-36242?

lukpueh (changes requested):

requirements-pinned.txt
lukpueh: Maybe we can add an inline comment here for maintainers to ignore Dependabot bumps in this line?

jku:
Agreed to all points, I've done the suggested changes in a new commit

lukpueh (approved):
Many thanks for fixing this, @jku!

Issue #334, opened 2021-02-15
Flaky GPG EdDSA tests


lukpueh:
**Description of issue or feature request**:

Excerpt from GitHub Action log with sporadically failing gpg eddsa tests observed on different Python versions ~and operating systems~:

```
======================================================================
FAIL: test_gpg_sign_and_verify_object_with_default_key (tests.test_gpg.TestGPGEdDSA)
Create a signature using the default key on the keyring
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/runner/work/securesystemslib/securesystemslib/tests/test_gpg.py", line 763, in test_gpg_sign_and_verify_object_with_default_key
    self.assertTrue(verify_signature(signature, key_data, test_data))
AssertionError: False is not true

======================================================================
FAIL: test_gpg_sign_and_verify_object_with_specific_key (tests.test_gpg.TestGPGEdDSA)
Create a signature using a specific key on the keyring
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/runner/work/securesystemslib/securesystemslib/tests/test_gpg.py", line 777, in test_gpg_sign_and_verify_object_with_specific_key
    self.assertTrue(verify_signature(signature, key_data, test_data))
AssertionError: False is not true
```

See  https://github.com/in-toto/in-toto/issues/171 for a similar issue, i.e. sporadically failing gpg *rsa* tests. The code snippets that were used to reproduce/troubleshoot the issue back then might be helpful for this issue as well.

**Current behavior**:
Sporadically failing tests:
- `tests.test_gpg.TestGPGEdDSA.test_gpg_sign_and_verify_object_with_default_key`
- `tests.test_gpg.TestGPGEdDSA.test_gpg_sign_and_verify_object_with_specific_key`

**Expected behavior**:
No sporadically failing tests.

lukpueh:
Right now we can only say for sure that the issue occurred on macOS runners. This would also explain why we can't recall these failures from before switching to GitHub actions (when we didn't test on macOS).

lukpueh:
This indeed might be a padding issue again. I ran one of our debugging scripts from in-toto/in-toto#171 and it looks like signature verification failure is related to the signature length. 


```python
import securesystemslib.gpg.functions as f
from collections import Counter
counter = Counter()

keyid = "4E630F84838BF6F7447B830B22692F5FEA9E2DD2"
homedir = "tests/gpg_keyrings/eddsa"
[...]

PR #340, opened 2021-02-24
Optionally pad OpenPGP EdDSA signature parameters

lukpueh:
Fixes: #334

Left-zero-pad 'r' and 's' values that are shorter than required by RFC 8032 (5.1.6.), to make up for omitted leading zeros in RFC 4880 (3.2.) MPIs. This is especially important for 's', which is little-endian.

Note that GnuPG seems to correctly parse EdDSA signatures with MPIs that omit leading zeros prior to validation. But in order to validate these signatures outside the context of RFC 4880, e.g. with pyca/cryptography's OpenSSL backend, we need to guarantee the correct length of the parsed signatures ourselves.

This PR also adds a test using a special-crafted short signature.  

See commit messages and #334 for more details.

Kudos to @jku and @SantiagoTorres for helping to troubleshoot.


securesystemslib/gpg/eddsa.py
```diff
@@ -140,6 +142,12 @@ def get_signature_params(data):
   ptr += 2
   s = data[ptr:ptr + s_length]
 
+  # Left-zero-pad 'r' and 's' values that are shorter than required by RFC 8032
+  # (5.1.6.), to make up for omitted leading zeros in RFC 4880 (3.2.) MPIs.
+  # This is especially important for 's', which is little-endian.
+  r = r.rjust(int(ED25519_SIG_LENGTH / 2), b"\x00")
+  s = s.rjust(int(ED25519_SIG_LENGTH / 2), b"\x00")
```
jku, line 149: not a big deal but I think this should work in all pythons:
```suggestion
  r = r.rjust(ED25519_SIG_LENGTH // 2, b"\x00")
  s = s.rjust(ED25519_SIG_LENGTH // 2, b"\x00")
```

tests/test_gpg.py
```diff
@@ -778,5 +779,24 @@ def test_gpg_sign_and_verify_object_with_specific_key(self):
...
+    test_data = b"hello"
+    signature_path = os.path.join(self.gnupg_home, "short.sig")
+
+    # Read special-crafted raw gpg signature that is one byte too short
+    with open(signature_path, "rb") as f:
+      signature_data = f.read()
+
+    # Check that the signature is padded upon parsing
+    signature = parse_signature_packet(signature_data)
+    self.assertTrue(len(signature["signature"]) == (ED25519_SIG_LENGTH * 2))
```
jku, line 794: So this is double length here because signature is already a hex string at this point?

jku (approved):
LGTM

securesystemslib/gpg/eddsa.py
lukpueh: Nice use of the floor division operator. Will update.

tests/test_gpg.py
lukpueh: Exactly. I'll add a comment.

lukpueh:
Addressed your comments and force-pushed. Thanks for the review, @jku!

PR #341, opened 2021-02-25
Add  GPGSigner implementation

MVrachev:
Related to: https://github.com/theupdateframework/tuf/issues/1263

The GPGSigner is an implementation of the new "Signer" interface
that was merged in https://github.com/secure-systems-lab/securesystemslib/pull/319
and could be found in securesystemslib.signer.py

This is the next logical step given that securesystemslib supports GPG
and we want this interface to have implementations for all signature
types which are already supported by securesystemslib.

While implementing the GPGSigner, I wanted to make sure that one can
easily sign a portion of data, receive a Signature object and use the
information stored in that object to verify the signature.

To verifty a GPG signature, one have to use
`securesystemslib/gpg/functions.verifty_signature()`:
https://github.com/secure-systems-lab/securesystemslib/blob/6895306b4aa788077c725425efae7018cdb675a5/securesystemslib/gpg/functions.py#L176
There, the `signature_object` argument should be in the
`securesystemslib.formats.GPG_SIGNATURE_SCHEMA` format:
https://github.com/secure-systems-lab/securesystemslib/blob/6895306b4aa788077c725425efae7018cdb675a5/securesystemslib/formats.py#L388
I searched for a way to easily retrieve the additional fields in the
GPG_SIGNATURE_SCHEMA - `other_headers` and `info` from the keyid stored
in the "Signature" object returned by the "sign" operation.
Unfortunately, right now **there is no function** that I can use for that
purpose.
The only option I was left with, was to create a new class:
`GPGSignature` where we can store those additional fields returned
from `securesystemslib.gpg.functions.create_signature()` which we call
during the "sign" process.

### DIFF
--- a/securesystemslib/signer.py
+++ b/securesystemslib/signer.py
@@ -259,7 +259,7 @@ class GPGSigner(Signer):
 
             ValueError: the gpg command failed to create a valid signature.
             OSError: the gpg command is not present or non-executable.
-            securesystemslib.exceptions.UnsupportedLibraryError: thehe gpg
+            securesystemslib.exceptions.UnsupportedLibraryError: the gpg
                 command is not available, or the cryptography library is
                 not installed.
             securesystemslib.gpg.exceptions.CommandError: the gpg command
--- a/securesystemslib/signer.py
+++ b/securesystemslib/signer.py
@@ -7,6 +7,7 @@ signing implementations and a couple of example implementations.
 
 import abc
 import securesystemslib.keys as sslib_keys
+import securesystemslib.gpg.functions as gpg
 from typing import Any, Dict, Optional, Mapping
 
 
@@ -88,6 +89,60 @@ class Signature:
 
 
 
+class GPGSignature(Signature):
+    """A container class containing information about a gpg signature.
+
+    Besides the signature, it also contains other meta information
+    needed to uniquely identify the key used to generate the signature.
+
+    Attributes:
+        keyid: HEX string used as a unique identifier of the key.
+        signature: HEX string representing the signature.
+        other_headers: HEX representation of additional GPG headers.
+    """
+    def __init__(
+        self,
+        keyid: str,
+        signature: str,
+        other_headers: str,
+    ):
+        super().__init__(keyid, signature)
+        self.other_headers = other_headers
+
+
+    @classmethod
+    def from_dict(cls, signature_dict: Dict) -> "Signature":
+        """Creates a GPGSignature object from its JSON/dict representation.
+
+        Args:
+            signature_dict: Dict containing valid "keyid", "signature" and
+                "other_fields" fields.
+
+        Raises:
+            KeyError: If any of the "keyid", "sig" or "other_headers" fields
+                are missing from the signature_dict.
+
+        Returns:
+            GPGSignature instance.
+        """
+
+        return cls(
+            signature_dict["keyid"],
+            signature_dict["sig"],
+            signature_dict["other_headers"]
+        )
+
+
+    def to_dict(self) -> Dict:
+        """Returns the JSON-serializable dictionary representation of self."""
+        return {
+            "keyid": self.keyid,
+            "signature": self.signature,
+            "other_headers": self.other_headers
+        }
+
+
+
 class Signer:
     """Signer interface created to support multiple signing implementations."""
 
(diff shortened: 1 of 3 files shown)


securesystemslib/signer.py
```diff
@@ -60,6 +60,61 @@ def to_dict(self) -> Dict:
...
+    needed to uniquely identify the key used to generate the signature.
+
+    Attributes:
+        keyid: HEX string used as a unique identifier of the key.
+        signature: HEX string representing the signature.
+        other_headers: HEX representation of additional GPG headers.
+        info: An optional dict containing information about the GPG key
+            used to generate the signature and the signed data.
+    """
+    def __init__(self, keyid: str, signature: str, other_headers: str, info=None):
```
lukpueh, line 76: No type annotation for `info`?

securesystemslib/signer.py
```diff
@@ -132,3 +187,69 @@ def sign(self, payload: bytes) -> "Signature":
...
+
+    """
+    def __init__(
+            self, keyid: Optional[str] = None,
+            homedir: Optional[str] = None):
+        self.keyid = keyid
+        self.homedir = homedir
+
+
+    def sign(self, payload) -> 'GPGSignature':
```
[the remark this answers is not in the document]
lukpueh: What about the type annotation for `payload`?

lukpueh (changes requested):
Cool stuff, @MVrachev! Thanks for adding this wrapper. Except for some minor documentation-related nits, the implementation looks good to me.

Regarding testing, I don't think it is necessary to copy/paste so much  of `test_gpg`. In terms of tested functionality (lines of code) it seems redundant, while adding extra review/maintenance effort. 

IMO it should be enough to add one exemplary test case for one key type only, doing
- signature creation
- maybe some vetting of the signature object ... were the fields assigned as expected? etc
- signature conversion (to dict)
- successful and unsuccessful signature verification

I suppose you could do this twice, one time with the default key and one time with a passed keyid.

MVrachev:
I have addressed all comments made by @lukpueh and fixed all documentation issues he pointed out.

Also, I removed a big chunk of the test files as they were replicating the same tests in `test_gpg.py`. 
I left only GPGRSA tests testing that signing and verification can be done with a default and custom keys.

MVrachev:
@lukpueh do we want to move this forward or it's not as much as important as originally discussed in https://github.com/theupdateframework/python-tuf/issues/1263

securesystemslib/signer.py
```diff
@@ -139,3 +197,62 @@ def sign(self, payload: bytes) -> "Signature":
...
+            payload: The bytes to be signed.
+
+        Raises:
+            securesystemslib.exceptions.FormatError:
+                If the keyid was passed and does not match
+                securesystemslib.formats.KEYID_SCHEMA.
+
+            ValueError: the gpg command failed to create a valid signature.
+            OSError: the gpg command is not present or non-executable.
+            securesystemslib.exceptions.UnsupportedLibraryError: thehe gpg
```
lukpueh, line 244:
```suggestion
            securesystemslib.exceptions.UnsupportedLibraryError: the gpg
```

lukpueh (commented):
@MVrachev, would you mind resolving the conflict here so that we can merge this PR?

@PradyumnaKrishna, pinged me about it, as he might find this useful when implementing DSSE (#370).

MVrachev:
@lukpueh and @PradyumnaKrishna  I am sorry I responded so late...
I rebased on top of master and fixed the docstring.

securesystemslib/signer.py
```diff
@@ -88,6 +89,64 @@ def to_dict(self) -> Dict:
...
+        other_headers: HEX representation of additional GPG headers.
+        info: Dict containing information about the GPG key
+            used to generate the signature and the signed data.
+    """
+    def __init__(
+        self,
+        keyid: str,
+        signature: str,
+        other_headers: str,
+        info: dict = None
```
lukpueh, line 110: I suggest we ignore `info` in this class, it's only needed internally and I don't expect `GPGSignature` to be used internally in `secureystemslib.gpg`.

See 
https://github.com/secure-systems-lab/securesystemslib/blob/a9244ed03cee41c593fb3c2459bf673ca33a8d7d/securesystemslib/gpg/common.py#L689-L692

lukpueh (changes requested):
💯 Thanks for reviving the PR, @MVrachev. I have one minor change request. Otherwise this looks great.

MVrachev:
@lukpueh  removed the `info` field as you have suggested.

lukpueh (approved):
Thanks!

PR #343, opened 2021-02-26
Prepare Debian downstream release 0.20.0-1

lukpueh:
* New upstream release includes among other things:
  - a new signing abstraction to facilitate custom implementations,
  - and a fix for a GnuPG/OpenSSL compatibility issue with OpenPGP
    EdDSA signatures.


lukpueh:
0.20.0-1 has been uploaded. Merging...

Issue #344, opened 2021-03-03
Drop support for Python 2


joshuagl:
**Description of issue or feature request**:

Our two main users, tuf (in releases after [0.17.0](https://github.com/theupdateframework/tuf/releases/tag/v0.17.0)) and in-toto (in releases after [1.0.1](https://github.com/in-toto/in-toto/releases/tag/v1.0.1)), have dropped support for Python 2.7. We should do the same.

**Current behavior**:

We have various code and configuration to support running on Python 2.7 and Python 3.x:
- handling multiple versions of dependencies in setup.py and requirements files
- stated support for Python 2.7 in setup.py
- use of the six module
- use of the future module

**Expected behavior**:

No code or configuration exists solely for Python 2.7 support.

joshuagl:
Before we start merging changes to remove Python 2.7, I think that we should make a final 2.7 supporting release which states clearly our intent to drop support for Python 2.7.

joshuagl:
When we start dropping Python 2 support, let's also update the dependabot config to remove pinned cryptography (https://github.com/secure-systems-lab/securesystemslib/pull/349).

PR #352, opened 2021-05-06
Start to drop Python 2.7

joshuagl:
Fixes: #

This is a first PR towards removing Python 2.7 from the code base, it is not sufficient but it does remove enough pieces that people will not be tempted to think we are still supporting Python 2.7 **and** starts getting us newer versions of cryptography via dependabot.

For more detail on what else needs to be removed, see https://github.com/secure-systems-lab/securesystemslib/issues/344

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,8 +88,6 @@ setup(
     'Operating System :: MacOS :: MacOS X',
     'Operating System :: Microsoft :: Windows',
     'Programming Language :: Python',
-    'Programming Language :: Python :: 2',
-    'Programming Language :: Python :: 2.7',
     'Programming Language :: Python :: 3',
     'Programming Language :: Python :: 3.6',
     'Programming Language :: Python :: 3.7',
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -103,7 +101,7 @@
     'Source': 'https://github.com/secure-systems-lab/securesystemslib',
     'Issues': 'https://github.com/secure-systems-lab/securesystemslib/issues',
   },
-  python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4",
+  python_requires="!=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4",
```
jku, line 104:
```suggestion
  python_requires="~=3.6",
```
I think the "compatible release" clause that TUF uses works here too (that should translate to any 3.X version where X>=6)

joshuagl: Yes, thank you. I even introduced that clause in tuf 🤦

jku (approved):
lgtm: there's still mentions of python2 even in the same files but this seems a fine first step

joshuagl:
> lgtm: there's still mentions of python2 even in the same files but this seems a fine first step

Thanks. I've pushed an extra commit to drop Python2 dependencies and dependency pinning (9b3a513).

Issue #357, opened 2021-05-18
Use mypy to statically check sources


jku:
We should start using mypy to check our sources statically. This implies also adding type annotations to functions. Doing this is useful for securesystemslib but doing this for public api is also useful for TUF (as TUF can then statically check the calls it makes to this API).

Configuring mypy and setting tox lint to run it (see https://github.com/theupdateframework/tuf/pull/1395/ for examples) should be easy, but it requires at least some parts of securesystemslib to be type annotated (#358)

See also [Using mypy with an existing codebase](https://mypy.readthedocs.io/en/stable/existing_code.html)

Issue #358, opened 2021-05-18
Add type annotations


jku:
We should type annotate our functions:
* Most importantly we should type annotate public methods (this should enable TUF to type check its calls)
* internal methods are useful to annotate as well (and may be necessary to enable static type checking in this repo)

This is companion to #357 (using mypy in securesystemslib): I think the process goes like this (although it's partially speculation: I've not tried this yet)
* annotate at least one module (at least the public methods)
* when public methods are annotated, this module can be enabled in TUF mypy config
* when one whole module is annotated, we can do #357 and start type checking here

Repeat steps for each module

[Using mypy with an existing codebase](https://mypy.readthedocs.io/en/stable/existing_code.html) may be useful: we may want to try tools to automatically generate type annotations: they might work well for this sort of code

sechkova:
`mypy` is now included in the CI run with #361.
 Also the `util` and `storage` modules are type annotated with #361 and #362.

jku:
The other modules used in TUF (metadata api) are:
```
exceptions
formats
keys
signer
```
So I guess those should be next in line (but I think we could also enable checks in TUF module-by-module if there are trickier modules that can't be annotated in the next weeks)

jku:
Also, I think we need to add a `securesystemslib/py.typed` marker file in the repo: https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-type-hints-for-third-party-library
Otherwise TUF can't use the SSLib annotations (and we definitely want that because we want to find things like https://github.com/theupdateframework/tuf/pull/1442)

jku:
> Also, I think we need to add a securesystemslib/py.typed

3 years later this is still true :joy:

PR #361, opened 2021-05-31
Start using mypy for static type checking

sechkova:
Fixes: #357
Partially #358

Use `mypy` for static type checking. 
- Start with adding type annotations to `util.py` 
- Add initial configuration file `mypy.ini `(needed because `securesystemslib` does not use `setup.cfg` file)
- Add one additional `tox` environment for `mypy` and the corresponding CI workflow

Inspired by https://github.com/theupdateframework/tuf/pull/1395
Developers can run `mypy` directly from the command line or by using `tox -e mypy`.


securesystemslib/util.py
```diff
@@ -246,7 +253,8 @@ def ensure_parent_dir(filename, storage_backend=None):
     storage_backend.create_folder(directory)
 
 
-def file_in_confined_directories(filepath, confined_directories):
+def file_in_confined_directories(filepath: str,
+    confined_directories: Union[List[str], Tuple[str]]) -> bool:
```
jku, line 257: The tuple case seems weird and completely unused ... but I think your annotation does not work as it defines a tuple of specifically one item.

I think this covers both cases: `confined_directories: Sequence[str]` -- docs say _Tuples implement all of the common sequence operations._ and that should be good enough for us since we only need `x in y`

Could update the method documentation as well.

jku (changes requested):
I think util.py should `import json` like normal people. 

If someone feels strongly about preserving the weird import_json(), the implementation of that then becomes a simple `return json`

securesystemslib/util.py
```diff
@@ -23,13 +23,16 @@
...
+from typing import Any, Dict, IO, List, Optional, Tuple, Union
 
 logger = logging.getLogger(__name__)
 
 
-def get_file_details(filepath, hash_algorithms=['sha256'],
-    storage_backend=None):
+def get_file_details(filepath: str, hash_algorithms: List[str] =['sha256'],
+    storage_backend: Optional[StorageBackendInterface] = None
+    ) -> Tuple[int, Dict[str, str]]:
```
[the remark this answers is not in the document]
jku: maybe even move filepath and hash_algorithms to new line? :shrug:

securesystemslib/util.py
sechkova: Seems like the correct annotation was `Tuple[str, ...]` but I agree that `Sequence` works just as well and makes things look cleaner.

sechkova:
Edited all three commits with the proposed changes:

- reformatted the function arguments in sort-of-black-style
- replaced Union[List, Tuple] with Sequence
- added mypy to the list of tox environments
- Dropped `import_json` and its test case

jku (approved):
thanks!

securesystemslib/util.py
```diff
@@ -17,19 +17,25 @@
...
+from typing import Any, Dict, IO, List, Optional, Sequence, Tuple, Union
 
 logger = logging.getLogger(__name__)
 
 
-def get_file_details(filepath, hash_algorithms=['sha256'],
-    storage_backend=None):
+def get_file_details(
+    filepath: str,
+    hash_algorithms: List[str] =['sha256'],
```
joshuagl, line 36:
```suggestion
    hash_algorithms: List[str] = ['sha256'],
```

securesystemslib/util.py
```diff
@@ -129,7 +138,10 @@ def get_file_hashes(filepath, hash_algorithms=['sha256'],
 
 
 
-def get_file_length(filepath, storage_backend=None):
+def get_file_length(
+    filepath: str,
+    storage_backend: Optional[StorageBackendInterface]=None
```
joshuagl, line 143:
```suggestion
    storage_backend: Optional[StorageBackendInterface] = None
```
?

securesystemslib/util.py
```diff
@@ -73,8 +79,11 @@ def get_file_details(filepath, hash_algorithms=['sha256'],
   return file_length, file_hashes
 
 
-def get_file_hashes(filepath, hash_algorithms=['sha256'],
-    storage_backend=None):
+def get_file_hashes(
+    filepath: str,
+    hash_algorithms: List[str] =['sha256'],
```
joshuagl, line 84:
```suggestion
    hash_algorithms: List[str] = ['sha256'],
```

securesystemslib/util.py
```diff
@@ -160,8 +172,12 @@ def get_file_length(filepath, storage_backend=None):
   return storage_backend.getsize(filepath)
 
 
-def persist_temp_file(temp_file, persist_path, storage_backend=None,
-    should_close=True):
+def persist_temp_file(
+    temp_file: IO,
+    persist_path: str,
+    storage_backend: Optional[StorageBackendInterface]=None,
```
joshuagl, line 178:
```suggestion
    storage_backend: Optional[StorageBackendInterface] = None,
```

securesystemslib/util.py
```diff
@@ -160,8 +172,12 @@ def get_file_length(filepath, storage_backend=None):
...
+    should_close: bool=True
```
joshuagl, line 179:
```suggestion
    should_close: bool = True
```

securesystemslib/util.py
```diff
@@ -200,7 +216,10 @@ def persist_temp_file(temp_file, persist_path, storage_backend=None,
     temp_file.close()
 
 
-def ensure_parent_dir(filename, storage_backend=None):
+def ensure_parent_dir(
+    filename: str,
+    storage_backend: Optional[StorageBackendInterface]=None
```
joshuagl, line 221:
```suggestion
    storage_backend: Optional[StorageBackendInterface] = None
```

securesystemslib/util.py
```diff
@@ -340,8 +361,9 @@ def import_json():
 
 json = import_json()
 
+json_str = Union[str, bytes, bytearray]
```
joshuagl, line 320: I think we just need `Union[str, bytes]` here? https://github.com/python/typeshed/blob/master/stdlib/json/__init__.pyi#L36

That should be short enough that we can include it inline?

securesystemslib/util.py
```diff
@@ -378,7 +356,10 @@ def load_json_string(data):
     return deserialized_object
 
 
-def load_json_file(filepath, storage_backend=None):
+def load_json_file(
+    filepath: str,
+    storage_backend: Optional[StorageBackendInterface]=None
```
joshuagl, line 361:
```suggestion
    storage_backend: Optional[StorageBackendInterface] = None
```

mypy.ini
```diff
@@ -0,0 +1,6 @@
+[mypy]
+warn_unused_configs = True
+files = securesystemslib/util.py
+# Supress error messages until enough modules
+# are type annotated
+follow_imports = silent
```
joshuagl, line 6:
```suggestion
follow_imports = silent

```
Very much a nit, but tools will keep complaining about not having a newline at the end of the file.

joshuagl (commented):
Thanks for this. I have only nit-picking comments to make, except for one on type annotations.

securesystemslib/util.py
sechkova: [json's docs](https://docs.python.org/3/library/json.html#json.loads) said that but I agree on the shorter and practical annotation.

joshuagl (approved):
Thanks @sechkova !

PR #362, opened 2021-05-31
Type annotate storage

sechkova:
Addresses #358 

Adds type annotations to `storage.py.`

In addition changes  `FilesystemBackend.get` to avoid  `mypy` complaining (correctly) about the following assignment:
```
  # Map our class ContextManager implementation to the function expected of the
  # securesystemslib.storage.StorageBackendInterface.get definition
  get = GetFile
```

```
securesystemslib/storage.py:224: error: Incompatible types in assignment (expression has type "Type[GetFile]", base class "StorageBackendInterface" defined the type as "Callable[[StorageBackendInterface, str], Generator[BinaryIO, None, None]]")
```

### DIFF
--- a/securesystemslib/storage.py
+++ b/securesystemslib/storage.py
@@ -200,7 +200,7 @@ class FilesystemBackend(StorageBackendInterface):
     try:
       file_object = open(filepath, 'rb')
       yield file_object
-    except (FileNotFoundError, IOError):
+    except OSError:
       raise exceptions.StorageError(
           "Can't open %s" % filepath)
     finally:
@@ -221,7 +221,7 @@ class FilesystemBackend(StorageBackendInterface):
         # and the operating system's buffers.  os.fsync() should follow flush().
         destination_file.flush()
         os.fsync(destination_file.fileno())
-    except (OSError, IOError):
+    except OSError:
       raise exceptions.StorageError(
           "Can't write file %s" % filepath)
 
--- a/securesystemslib/storage.py
+++ b/securesystemslib/storage.py
@@ -22,7 +22,7 @@ import os
 import shutil
 from contextlib import contextmanager
 from securesystemslib import exceptions
-from typing import BinaryIO, Generator, IO, List
+from typing import BinaryIO, IO, Iterator, List
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


sechkova:
Mentioning @joshuagl in case he can remember something that I have missed in the context manager implementation.

jku (changes requested):
Looks good to me apart from that one fishy looking exception.

securesystemslib/storage.py
```diff
@@ -200,7 +200,7 @@ def get(self, filepath:str) -> Generator[BinaryIO, None, None]:
     try:
       file_object = open(filepath, 'rb')
       yield file_object
-    except (FileNotFoundError, IOError):
+    except FileNotFoundError:
```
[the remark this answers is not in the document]
sechkova: You have a point, updated in 32ab712

jku (approved):

securesystemslib/storage.py
```diff
@@ -37,7 +38,8 @@ class StorageBackendInterface():
 
 
   @abc.abstractmethod
-  def get(self, filepath):
+  @contextmanager
+  def get(self, filepath: str) -> Generator[BinaryIO, None, None]:
```
joshuagl, line 42: I think we can just use `Iterator[None]` for the return type here?
See https://stackoverflow.com/questions/49733699/python-type-hints-and-context-managers and https://github.com/python/typeshed/blob/master/stdlib/contextlib.pyi#L34

securesystemslib/storage.py
```diff
@@ -193,7 +195,7 @@ def __new__(cls, *args, **kwargs):
 
 
   @contextmanager
-  def get(self, filepath):
+  def get(self, filepath:str) -> Generator[BinaryIO, None, None]:
```
joshuagl, line 198: Same comment on return type here.

joshuagl (commented):
> Mentioning @joshuagl in case he can remember something that I have missed in the context manager implementation.

This looks great. It may have been a Python 2.7 limitation or just my not finding the right part of the docs when I implemented it without the decorator.

securesystemslib/storage.py
jku: if that works (and that would make sense) then I suppose it's `Iterator[BinaryIO]`?

joshuagl: That would make more sense, indeed.

sechkova:
Added one commit with the proposal. Checked and `mypy` seems to accept it.

joshuagl:
Just realised that you'll need to add storage.py to the list of files in mypy.ini

sechkova:
> Just realised that you'll need to add storage.py to the list of files in mypy.ini

Rebased on top of the main branch and added.

joshuagl (approved):
Thanks, this is a nice change.

PR #383, opened 2022-01-18
Add Signature.__eq__() method

MVrachev:

Add `Signature.__eq__()` method to support comparison between signature
objects.
The need for this method came when solving issue #1696 in TUF
and more precisely read: https://github.com/theupdateframework/python-tuf/issues/1696#issuecomment-1007448882

Issue #1696 in TUF is about adding validation API after metadata
object initialization and because signatures in TUF are dictionaries
whose values are Signature objects they need for Signature.__eq__()
the method became apparent.

### DIFF
--- a/securesystemslib/signer.py
+++ b/securesystemslib/signer.py
@@ -7,7 +7,7 @@ signing implementations and a couple of example implementations.
 
 import abc
 import securesystemslib.keys as sslib_keys
-from typing import Dict
+from typing import Any, Dict
 
 
 class Signature:
@@ -27,6 +27,13 @@ class Signature:
     def __init__(self, keyid: str, sig: str):
         self.keyid = keyid
         self.signature = sig
+
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Signature):
+            return False
+
+        return self.keyid == other.keyid and self.signature == other.signature
 
 
     @classmethod
--- a/tests/test_signer.py
+++ b/tests/test_signer.py
@@ -2,6 +2,7 @@
 
 """Test cases for "signer.py". """
 
+import copy
 import sys
 import unittest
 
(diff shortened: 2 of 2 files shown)


securesystemslib/signer.py
```diff
@@ -29,6 +29,13 @@ def __init__(self, keyid: str, sig: str):
         self.signature = sig
 
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Signature):
+            return False
+
+        return self.keyid == other.keyid and self.signature == other.signature
```
adityasaky, line 36: This should likely handle the `other_headers` field used in GPG signatures. Also, should we be worried about different methods of key ID calculation? This may be an uneducated question on my part. :smile:

MVrachev: I don't agree it should handle `other_headers` fied used in GPG signatures considering it doesn't have access to that field.
If we want to implement `__eq__` for the `GPGSignature` as well we will have to add it's own separate `__eq__` implementation looking like this:
```python
    def __eq__(self, other: Any) -> bool:
        if not isinstance(other, GPGSignature):
            return False

        return super().__eq__(other) and self.other_headers == other.other_headers

```

MVrachev: The definition of `KEYID` is the following
```
The identifier of the key signing the ROLE object, which is a hexdigest of the
SHA-256 hash of the canonical form of the key. 
```
Meaning `keyid` will be a string in all cases and strings have defined `__eq__` methods.

Source: https://theupdateframework.github.io/specification/latest/#role-keyid

adityasaky: > GPGSignature

I was confused because I didn't see it in the repo yesterday, but it's still an open PR. Makes sense to me, then. :)

> KEYID

Cool! I just wanted to explicitly clarify it, thank you!

adityasaky (approved):

jku:
Adityas comment reminded me: 
* should this Signature support custom fields?
* custom fields should probably affect eq wherever they are supported

MVrachev:
In [ADR-0008](https://github.com/theupdateframework/python-tuf/blob/develop/docs/adr/0008-accept-unrecognised-fields.md) we decided we will accept unrecognized fields almost everywhere.
Wondering is possible to compromise the security by adding unrecognized fields inside the `Signature` object?

jku:
> Wondering if it's possible to compromise the security by adding unrecognized fields inside the Signature object?

It is true that we should be more cautious with this part of the metadata (not because they contain signatures but because they are not part of the signed portion of metadata). But: because of TUF design that requires canonicalization we end up parsing even the signed portion before we can check that is has been signed. 

The reason I'm asking about this is that it's possible we want to extend signatures -- as an example I seem to remember the sigstore fulcio support is going to require an extension.

so questions are:
* does specification demand or prevent that signature "object" is an extensible data structure? (my assumption is that spec does not define this clearly but that it should be extensible). if signature object should **not** be extensible, then I think fulcio support needs a forwards incompatible TUF spec update
* should python-tuf support unrecognised fields in "signature" dict: I think we probably should

MVrachev:
I think the question about if signatures should allow unrecognized fields should be clarified first in the specification (see https://github.com/theupdateframework/specification/issues/203) and in python-tuf (see https://github.com/theupdateframework/python-tuf/issues/1802). 

This issue shouldn't stop us from merging the current `__eq__` implementation and if we decide to allow unrecognized fields in the Signature objects, we can always update it.

MVrachev:
@adityasaky what further steps are required to merge this pr?

adityasaky:
I'm fine with this PR, I wasn't sure if @lukpueh was also going to review it. :)

lukpueh (approved):
Thanks!

PR #387, opened 2022-02-09
Add unrecognized fields support for Signature

MVrachev:

The TUF specification says the following regarding unknown fields:
"All of the formats described below include the ability to add more
attribute-value fields to objects for backward-compatible format
changes. Implementers who encounter undefined attribute-value pairs in
the format must include the data when calculating hashes or verifying
signatures and must preserve the data when re-serializing."
From: https://theupdateframework.github.io/specification/latest/#document-formats

This section is the reason [ADR0008]( https://github.com/theupdateframework/python-tuf/blob/develop/docs/adr/0008-accept-unrecognised-fields.md) was accepted inside `python-tuf`
and we have added support for unrecognized fields for all fields
inside the SIGNED portion of the metadata.
However, this limits what the citation implies or that everywhere there
inside the metadata files including signatures we should accept
unrecognized fields. That's why I made these changes.
These changes have the approval of the community see:
- https://github.com/theupdateframework/specification/issues/203
- https://github.com/theupdateframework/python-tuf/issues/1802

### DIFF
--- a/securesystemslib/signer.py
+++ b/securesystemslib/signer.py
@@ -7,7 +7,7 @@ signing implementations and a couple of example implementations.
 
 import abc
 import securesystemslib.keys as sslib_keys
-from typing import Any, Dict
+from typing import Any, Dict, Optional, Mapping
 
 
 class Signature:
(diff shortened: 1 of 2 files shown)


MVrachev:
I run `python-tuf` locally with the changes included in this commit and they all passed without problems.
I had to install `securesystemslib` by using my local path.
I even debugged the `python-tuf` tests code to make sure that when we create metadata objects with `Metadata.from_dict()` and as a consequence call `Signature.from_dict()` that we are actually calling `Signature.from_dict()` with the latest changes included in this pr.

PR #391, opened 2022-02-16
Improve handling of errors in securesystemslib.keys.vertify_signature()

MVrachev:
Together with @jku we have found two cases when calling `securesystemslib.keys.verify_signatures()` 
with malformed `key_dict` argument results to an exception from a third-party library:
1) when the `key_dict` represents an `ecdsa` key and `key_dict["keyval"]["public"]` is invalid (e.g. key_dict["keyval"]["public"] are raw bytes encoded in HEX instead of PEM format). This results in:
```
ValueError: ('Could not deserialize key data. The data may be in an incorrect format, it may be encrypted with an unsupported algorithm, or it may be an unsupported key type (e.g. EC curves with explicit parameters).', [_OpenSSLErrorWithText(code=151584876, lib=9, reason=108, reason_text=b'error:0909006C:PEM routines:get_name:no start line')])
```
2) when the `key_dict` represents an `ed25519 ` key and `key_dict["keyval"]["public"]`  is invalid (e.g. key_dict["keyval"]["public"] are raw bytes encoded in PEM instead of HEX format). This results in:
```
"binascii.Error: Non-hexadecimal digit found" 
```

This doesn't look good considering securesystemslib actively replaces errors related to malformed arguments with `FormatError`, so in order to be consistent it's good we do the same here.

_This pr is initiated by an issue found by @jku inside `python-tuf`: https://github.com/theupdateframework/python-tuf/issues/1866_

PS: If someone wants to reproduce the issue inside a terminal here are exemplary commands:

```python
import securesystemslib.keys as keys
ecdsa = keys.generate_ecdsa_key()
ecdsa["keyval"]["public"]="abcd" # break key value
sig={"keyid": ecdsa["keyid"], "sig": "bb"}
keys.verify_signature(ecdsa, sig, "data")
```

### DIFF
--- a/securesystemslib/ecdsa_keys.py
+++ b/securesystemslib/ecdsa_keys.py
@@ -311,8 +311,11 @@ def verify_signature(public_key, scheme, signature, data):
   formats.ECDSA_SCHEME_SCHEMA.check_match(scheme)
   formats.ECDSASIGNATURE_SCHEMA.check_match(signature)
 
-  ecdsa_key = load_pem_public_key(public_key.encode('utf-8'),
-      backend=default_backend())
+  try:
+    ecdsa_key = load_pem_public_key(public_key.encode('utf-8'),
+        backend=default_backend())
+  except ValueError as e:
+    raise exceptions.FormatError(f'Failed to load PEM key {public_key}') from e
 
   if not isinstance(ecdsa_key, ec.EllipticCurvePublicKey):
     raise exceptions.FormatError('Invalid ECDSA public'
--- a/securesystemslib/keys.py
+++ b/securesystemslib/keys.py
@@ -838,7 +838,12 @@ def verify_signature(key_dict, signature, data):
 
   elif keytype == 'ed25519':
     if scheme == 'ed25519':
-      public = binascii.unhexlify(public.encode('utf-8'))
+      try:
+        public = binascii.unhexlify(public.encode('utf-8'))
+      except binascii.Error as e:
+        raise exceptions.FormatError(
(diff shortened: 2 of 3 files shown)


jku (commented):
This looks right, thanks (ideally the raised errors probably should be ValueErrors but since sslib already uses FormatError, continuing that pattern seems correct).

As a nitpick, the messages could explain what specific action failed, like `"Failed to parse key '{key}' as hex"` or `"Failed to load PEM key '{key}'"`: this could make debugging a little easier

MVrachev:
I wanted to point out that I tested changing the `keyval` inside an rsa key and it behave as expected it threw a `FormatError`. 
A similar case is tested for rsa here: https://github.com/secure-systems-lab/securesystemslib/blob/934fc1b835dc0c594ccb2d52155d5970792ace7c/tests/test_rsa_keys.py#L117 and I don't see a need to include a test specifically making `keyval` in HEX format.

MVrachev:
> This looks right, thanks (ideally the raised errors probably should be ValueErrors but since sslib already uses FormatError, continuing that pattern seems correct).
> 
> As a nitpick, the messages could explain what specific action failed, like `"Failed to parse key '{keyvalue}' as hex"` or `"Failed to load PEM key '{keyvalue}'"`: this could make debugging a little easier -- but the context is there in the stack trace so it's not a major complaint

Changed the code to use the messages suggested by @jku.

jku (approved):

lukpueh (approved):

PR #400, opened 2022-04-11
Ignore FileNotFoundError on GPG tests teardown

### LINKED ISSUE #397: Possible race condition in gpg test cleanup function
Fwd: [*Bug#1008349: python-securesystemslib: FTBFS: dh_auto_test: error: pybuild --test -i python{version} -p "3.10 3.9" returned exit code 13*
](https://groups.google.com/g/securesystemslib-dev/c/J2PYMN696sc/m/FncOdnEGBAAJ)
The following error occurred on Debian sid:

```
======================================================================
ERROR: tearDownClass (test_gpg.TestGPGDSA)
  File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.9_securesystemslib/build/tests/test_gpg.py", line 667, in tearDownClass
    shutil.rmtree(self.test_dir)
  File "/usr/lib/python3.9/shutil.py", line 732, in rmtree
    _rmtree_safe_fd(fd, path, onerror)
  File "/usr/lib/python3.9/shutil.py", line 665, in _rmtree_safe_fd
    _rmtree_safe_fd(dirfd, fullname, onerror)
  File "/usr/lib/python3.9/shutil.py", line 688, in _rmtree_safe_fd
    onerror(os.unlink, fullname, sys.exc_info())
  File "/usr/lib/python3.9/shutil.py", line 686, in _rmtree_safe_fd
    os.unlink(entry.name, dir_fd=topfd)
FileNotFoundError: [Errno 2] No such file or directory: 'S.gpg-agent.ssh'

```

**Possible explanation**:
I cannot reproduce the issue, but given the internals of the relevant test and the fact that we haven't seen the error before, a race condition seems likely.

I suppose `gpg-agent`, which is started during the tests by calling `gpg`, is closing/removing the previously opened socket 'S.gpg-agent.ssh', while `shutil.rmtree` is removing the temporary test gpg homedir containing the socket file. I.e. after `shutil.rmtree` has identified the file to be removed, but before it had the chance to actually remove it.

**Possible solution**:
An easy solution would be to configure `shutil.rmtree` to [`ignore_errors=True`](https://docs.python.org/3/library/shutil.html#shutil.rmtree), or to at least ignore `FileNotFoundError`s using a custom [`onerror` handler](https://docs.python.org/3/library/shutil.html#shutil.rmtree) (see [default `onerror`](https://github.com/python/cpython/blob/27ee43183437c473725eba00def0ea7647688926/Lib/shutil.py#L697-L702) for inspiration). This has the downside that we might masquerade errors that we don't want to pass on.

An alternative solution would be to configure `gpg-agent` to not open 'S.gpg-agent.ssh', which IIUC we don't use anyway. But then we might fail on other sockets created by the agent that we do use (e.g. 'S.gpg-agent' ?). See [this SO post](https://unix.stackexchange.com/a/605639) for good pointers to the corresponding gpg docs.

elfotografo007:
Fixes #397

Signed-off-by: Andrés Torres <andrest@vmware.com>

There is a suspected race condition when calling shutil.rmtree so we are
ignoring any FileNotFoundError when the test class calls tearDownClass

### DIFF
--- a/tests/test_gpg.py
+++ b/tests/test_gpg.py
@@ -59,6 +59,17 @@ from securesystemslib.formats import (GPG_PUBKEY_SCHEMA,
     ANY_PUBKEY_DICT_SCHEMA)
 
 
+class GPGTestUtils:
+    """GPG Test utility class"""
+
+    @staticmethod
+    def ignore_not_found_error(function, path, exc_info):
+        """Callback that ignores FileNotFoundError"""
+        _, error, _ = exc_info
+        if not isinstance(error, FileNotFoundError):
+            raise error
+
+
 @unittest.skipIf(not HAVE_GPG, "gpg not found")
 class TestUtil(unittest.TestCase):
   """Test util functions. """
@@ -503,7 +514,7 @@ class TestGPGRSA(unittest.TestCase):
   def tearDownClass(self):
     """Change back to initial working dir and remove temp test directory. """
     os.chdir(self.working_dir)
-    shutil.rmtree(self.test_dir)
+    shutil.rmtree(self.test_dir, onerror=GPGTestUtils.ignore_not_found_error)
 
   def test_export_pubkey_error(self):
     """Test correct error is raised if function called incorrectly. """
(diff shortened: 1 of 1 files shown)


adityasaky (approved):

lukpueh (commented):
Thanks, @elfotografo007, this is the fix I had in mind! 

One question, why is `ignore_not_found_error` a static method of a new class? Why not make it a normal module scope function? This would spare a couple of lines of code and make the invocation shorter.

lukpueh (approved):

PR #403, opened 2022-04-26
Use reliable python interpreter path in tests

AMDmi3:
`python` executable is not guaranteed to exist, so use `sys.executable` instead, which always points to correct python interpreter.


adityasaky:
Thanks, @AMDmi3!

Issue #360, opened 2021-05-28
accept Signature in verify_signature()


jku:
A signature object was added into SSLib but verify_signature() wasn't updated to accept it. 
It leads to weird code:

```python
if not sslib_keys.verify_signature(
    sslib_key,
    signature.to_dict(), # <- why am I serializing the signature here ?
    data,
):
    print ("oh no")
```

lukpueh:
I think we can change this into a "refactor internals of new signer API"-issue. cc @jku

jku:
Yes I think I'm fine with closing this... If the signer module provides me with everything I as a user need, I don't really care how the keys module operates.

PR #409, opened 2022-06-02
Add support for Signature class instances

PradyumnaKrishna:
Fixes: #360

Currently, `verify_siganture()` method takes a signature dictionary as parameter and there is no implementation for `Signature` class present in `securesystemslib.signer`. This Pull Request adds support for `Signature` class object in `create_signature()` and `verify_signature()` methods.

### DIFF
--- a/securesystemslib/keys.py
+++ b/securesystemslib/keys.py
@@ -613,17 +613,17 @@ def create_signature(key_dict, data):
     >>> ed25519_key = generate_ed25519_key()
     >>> data = 'The quick brown fox jumps over the lazy dog'
     >>> signature = create_signature(ed25519_key, data)
-    >>> securesystemslib.formats.SIGNATURE_SCHEMA.matches(signature)
+    >>> isinstance(signature, securesystemslib.signer.Signature)
     True
     >>> len(signature['sig'])
     128
     >>> rsa_key = generate_rsa_key(2048)
     >>> signature = create_signature(rsa_key, data)
-    >>> securesystemslib.formats.SIGNATURE_SCHEMA.matches(signature)
+    >>> isinstance(signature, securesystemslib.signer.Signature)
     True
     >>> ecdsa_key = generate_ecdsa_key()
     >>> signature = create_signature(ecdsa_key, data)
-    >>> securesystemslib.formats.SIGNATURE_SCHEMA.matches(signature)
+    >>> isinstance(signature, securesystemslib.signer.Signature)
     True
 
   <Arguments>
@@ -662,8 +662,7 @@ def create_signature(key_dict, data):
     actual signing routine.
 
   <Returns>
-    A signature dictionary conformant to
-    'securesystemslib_format.SIGNATURE_SCHEMA'.
+    A "securesystemslib.singer.Signature" class instance.
   """
 
   # Does 'key_dict' have the correct format?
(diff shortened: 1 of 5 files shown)


jku:
Nice work. I think the high level issue here is that this is public API that is being used. So while I think something like this is indeed a good path forward -- in my opinion public API arguments and return values should be better defined than "it's a dict!" -- securesystemslib might not want to just change the API... But I'm just assuming: hopefully actual maintainers have opinions here.

lukpueh:
Impeccable work, @PradyumnaKrishna. There is nothing for me to comment on in the patch. 💯 

But @jku is right that we need to consider carefully if we want to  merge (now) and thus disrupt in-toto/tuf operations in the next release.

I suggest we hold off with merging until we have a clear plan for #270.

cc @adityasaky, @SantiagoTorres

adityasaky (commented):
I like the changes here, but I'm indeed wary about this breaking in-toto / python-tuf. That said, I suspect (I haven't had a chance to look) modifying both in parallel to switch away from the `dict` shouldn't be too difficult or a breaking change by any means. @PradyumnaKrishna, can I ask you to take a look at what it would mean to make that change? :)

PradyumnaKrishna:
@adityasaky In in-toto, `Metablock` uses these functions in `sign` and `verify_signature` methods. in-toto metadata has a list of signatures, which also consist GPG Signatures in format of `dict`.
Simplest way to achieve this is by converting the Signature object to dict in `sign` function while converting the stored dict into a signature object in `verify` function of `Metablock`.

adityasaky:
@lukpueh @SantiagoTorres thoughts on modifying in-toto simultaneously?

lukpueh:
I gave this more thought and would like to suggest to not touch `keys.{create, verify}_signature` for the time being. Let's instead settle on a good alternative first, start using it in tuf and in-toto, and only then deprecate/refactor the current signing/verifying API.

Good alternatives:
-  [`Signer.sign`](https://github.com/secure-systems-lab/securesystemslib/blob/bd4afbe8eb2ded83fc3b1d7a81618a2a9c7242fa/securesystemslib/signer.py#L97)
- [`Key.verify_signature`](https://github.com/theupdateframework/python-tuf/blob/39f03ac8b161ad3063b995fcbd5c64b506a7dafc/tuf/api/metadata.py#L733) (currently implemented in python-tuf metadata API, but seems like a good fit for `securesystemslib`)

securesystemslib/keys.py
```diff
@@ -803,14 +801,15 @@ def verify_signature(key_dict, signature, data):
   formats.ANYKEY_SCHEMA.check_match(key_dict)
 
   # Does 'signature' have the correct format?
-  formats.SIGNATURE_SCHEMA.check_match(signature)
+  if not isinstance(signature, signer.Signature):
+    raise exceptions.FormatError("Wanted a Signature object.")
```
MVrachev, line 805: Nit, feel free to ignore if you decide:
A comment I once got a while ago: `Don't add dots or exclamation marks at the end of an exception message`.

MVrachev (commented):
I like the changes, good work!

jku:
My late review and links to related work -- I believe this mostly duplicates lukas last comment:

I think modifying keys module  API is not a good idea:
  * it is a (hard) breaking change for all current users
  * keys module is not a great API as a whole, so getting it to a good place with small tweaks... is going to require a lot of small tweaks. So this would not be the last API break
  * I think the Signer + Key approach that was taken in the intoto branch (and in my related PR) are better: the create a better API on top of the current functionality. This allows us to see if the new API is good and if it is, then start deprecating and improving the implementation underneath -- requiring only one deprecation and API break. Whether signature verification is implemented in Key or Signature I don't care so much, but I do think keys module API should not change
  
Links:
* [intoto fork](https://github.com/secure-systems-lab/securesystemslib/compare/master...in-toto:securesystemslib:master): this is a combination of Key changes, signer changes and a new DSSE enveloper implementation
* My [draft PR](https://github.com/secure-systems-lab/securesystemslib/pull/456) with Key + Signer improvements, partly overlapping with the above

I think this PR should maybe  be closed?

PR #420, opened 2022-07-18
Fix tests failing in GitHub Actions

PradyumnaKrishna:
Test `test_run_duplicate_streams_timeout` fails sometimes, and after
some investigations it is found that the process used in the test
exits before considered for a Timeout.

To fix the test, the command is now replaced with another command.

Also, I consider testing this Pull Request multiple time using GitHub Actions
before merging to make sure the problem is now solved.

A failed test: https://github.com/secure-systems-lab/securesystemslib/runs/7280432273

### DIFF
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -118,8 +118,8 @@ class Test_Process(unittest.TestCase):
   def test_run_duplicate_streams_timeout(self):
     """Test raise TimeoutExpired. """
     with self.assertRaises(securesystemslib.process.subprocess.TimeoutExpired):
-      securesystemslib.process.run_duplicate_streams(sys.executable + " --version",
-          timeout=-1)
+      securesystemslib.process.run_duplicate_streams(sys.executable +
+        " -c \"while True: pass\"", timeout=-1)
 
 
   def test__default_timeout(self):


tests/test_process.py
```diff
@@ -118,8 +118,7 @@ def test_run_cmd_arg_return_code(self):
   def test_run_duplicate_streams_timeout(self):
     """Test raise TimeoutExpired. """
     with self.assertRaises(securesystemslib.process.subprocess.TimeoutExpired):
-      securesystemslib.process.run_duplicate_streams(sys.executable + " --version",
-          timeout=-1)
+      securesystemslib.process.run_duplicate_streams("sleep 3", timeout=-1)
```
lukpueh, line 121: Although, we currently don't run tests on Windows, I'd like us to keep the tests platform independent. What speaks against running `python -c 'while True: pass'`?

lukpueh (commented):
Thanks for the fix! What do you think should be the expected behavior for `timeout <= 0`? It seems surprising that a subprocess can successfully run with `timeout=-1`, just because it finishes before we check for timeout. Would you mind taking a look how the subprocess module handles this?

PradyumnaKrishna:
> Thanks for the fix! What do you think should be the expected behavior for `timeout <= 0`?

`timeout` in subprocess do pretty similar stuff, they only checks it once which is after process exits.
We checks timeout if the process is not exited, a mandatory check before while loop would work.

Issue #421, opened 2022-08-04
Cross-platform verification of RSA PSS signatures


trishankatdatadog:
**Description of issue or feature request**:

There are two related issues:
1. The [salt length](https://crypto.stackexchange.com/questions/1217/rsa-pss-salt-size) used when _creating_ RSA PSS signatures.
2. And the salt length used when _verifying_ these signatures.

**Current behavior**:

1. We use a salt length equal to [DIGEST_LENGTH](https://cryptography.io/en/37.0.0/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.padding.PSS.DIGEST_LENGTH) when _creating_ these signatures.
2. Same for _verifying_ them. This is a problem when verifying cross-platform signatures (e.g., from the Golang `crypto/rsa` packages, which uses a different salt length by default as described below; see https://github.com/hashicorp/vault/pull/16549).

**Expected behavior**:
1. Although optional, and not necessary for the cross-platform verification of signatures, we should use a salt length equal to [MAX_LENGTH](https://cryptography.io/en/37.0.0/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.padding.PSS.MAX_LENGTH) for the maximum provable security when _creating_ these signatures.
2. We should use a salt length equal to [AUTO](https://cryptography.io/en/37.0.0/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.padding.PSS.AUTO) so that we can automatically _verify_ these signatures regardless of the input salt length.

trishankatdatadog:
(Related: #262)

PR #422, opened 2022-08-04
Fix RSA PSS salt lengths

trishankatdatadog:
Fixes: #421 

This PR does two things:
1. Use the _maximum_ salt length by default when _creating_ RSA PSS signatures.
1. Use the _automatic_ salt length inferred from RSA PSS signatures when _verifying_ them.

This should be the simplest, most backwards-compatible, and yet most secure way to verify cross-platform RSA PSS signatures.

### Missing tests
- [x] Can automatically verify RSA PSS signatures regardless of the input salt length

### DIFF
--- a/tests/test_rsa_keys.py
+++ b/tests/test_rsa_keys.py
@@ -23,8 +23,11 @@ import securesystemslib.exceptions
 import securesystemslib.formats
 import securesystemslib.keys
 import securesystemslib.rsa_keys
+import securesystemslib.hash
 
-from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.asymmetric import padding
+from cryptography.hazmat.primitives.serialization import load_pem_private_key
 
 public_rsa, private_rsa = securesystemslib.rsa_keys.generate_rsa_public_and_private()
 FORMAT_ERROR_MSG = 'securesystemslib.exceptions.FormatError raised.  Check object\'s format.'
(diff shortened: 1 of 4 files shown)


lukpueh (commented):
Thanks, @trishankatdatadog. Great patch and issue/pr description! 

> Can automatically verify RSA PSS signatures regardless of the input salt length

Testing that would indeed be great. Do you have capacities?

trishankatdatadog:
> Testing that would indeed be great. Do you have capacities?

Done. PTAL?

tests/test_rsa_keys.py
```diff
@@ -154,6 +157,31 @@ def test_verify_rsa_signature(self):
           scheme, public_rsa, data))
 
 
+  def test_verify_rsa_pss_sha256(self):
+    global public_rsa
+    global private_rsa
```
lukpueh, line 162: I suppose you just copy/pasted these, but do we actually need `global` statements here?

tests/test_rsa_keys.py
```diff
@@ -154,6 +157,31 @@ def test_verify_rsa_signature(self):
...
+
+    rsa_scheme = 'rsassa-pss-sha256'
+    data = 'The quick brown fox jumps over the lazy dog'.encode('utf-8')
+
+    private_key_object = load_pem_private_key(private_rsa.encode('utf-8'),
+        password=None, backend=default_backend())
+    digest_obj = securesystemslib.hash.digest_from_rsa_scheme(rsa_scheme,
+        'pyca_crypto')
+    max_salt_length = padding.calculate_max_pss_salt_length(private_key_object, digest_obj)
+
```
lukpueh, line 172: `max_salt_length` should stay the same for this key, right? Does it make sense to assert the value, so that we know how many times `assertEqual` is actually called below?

lukpueh (approved):
Thanks for adding tests, @trishankatdatadog! Left two non-blocking comments inline.

tests/test_rsa_keys.py
trishankatdatadog: I don't think it'll test much other than Python FOR loops work as expected, and we didn't accidentally misuse it.

Please take a look at the new test to see if it makes you feel better?

trishankatdatadog: I might have [more comprehensive tests](https://github.com/hashicorp/vault/pull/16549). Pls let me know how you'd like to proceed. Thx.

tests/test_rsa_keys.py
```diff
@@ -158,28 +159,39 @@ def test_verify_rsa_signature(self):
...
-      valid_signature = \
-        securesystemslib.rsa_keys.verify_rsa_signature(signature,
-          rsa_scheme, public_rsa, data)
-
-      self.assertEqual(True, valid_signature)
+      # Current behaviour: automatic inferring of salt length.
+      self.assertTrue(securesystemslib.rsa_keys.verify_rsa_signature(signature,
+          rsa_scheme, public_rsa, data))
+
+      # Previous behaviour: generally wrong assumption about salt length.
```
lukpueh, line 183: What's the purpose of the two tests below? They seem to test `cryptography` API and not our code. Or is this a regression test that accepts a regression? :D

tests/test_rsa_keys.py
lukpueh: I'm actually fine the way it was before and don't fully understand the ["better test"](https://github.com/secure-systems-lab/securesystemslib/pull/422/commits/3f3ed970a45be47209714f5aa66fc10b4e299421) you added. See previous comment.

tests/test_rsa_keys.py
trishankatdatadog: I think my thinking was to show that previously we would have failed, and now we don't, although that's not clear by looking at this. More importantly, I need to add tests to show that: (1) _previous_ signatures are still automatically verifiable, (2) new signatures are automatically verifiable (I think we have already done that here), and (3) new signatures from _other_ platforms (e.g., Vault) are automatically verifiable.

trishankatdatadog:
@lukpueh OK, done, simplified tests, PTAL, thx!

lukpueh (approved):
💯

trishankatdatadog:
Thanks, Lukas! Could we please get a release?

Cc @yzhan289

Issue #239, opened 2020-05-28
Better distinguish between keytype and scheme for ECDSA keys


adityasaky:
**Description of issue or feature request**:

The key format provided by `securesystemslib` defines a `keytype` field and a `scheme` field. Currently, there is no distinction between these fields for ECDSA keys.

**Current behavior**:

ECDSA keys have the following format:

```
{
    "keytype" : "ecdsa-sha2-nistp256",
    "scheme" : "ecdsa-sha2-nistp256",
    "keyval" : {"public" : PUBLIC}
}
```
As noted in https://github.com/secure-systems-lab/securesystemslib/commit/761aded9968cb380004a86c44eb4796c061bca9c, `ecdsa-sha2-nistp384` is also supported by `securesystemslib`.

**Expected behavior**:

Update `securesystemslib` to use just `ecdsa` as `keytype` while retaining the current scheme formats. This also aligns with the key format for RSA keys where we have:

```
{
    "keytype" : "rsa",
    "scheme" : "rsassa-pss-sha256",
    "keyval" : {"public" : PUBLIC}
}
```

shibumi:
Good work, I just want to mention, that in in-toto-golang we currently follow the expected behavior:

```
{
    "keytype" : "ecdsa",
    "scheme" : "ecdsa-sha2-nistp256",
    "keyval" : {"public" : PUBLIC}
}
```

lukpueh:
Closed by https://github.com/secure-systems-lab/securesystemslib/pull/267

PR #427, opened 2022-09-15
Add Sphincs+ To Signer Interface

rugo:
Fixes: 
Updates: #169

The PRs #160 and #169  tried to include the PQC scheme [SPHINCS+](https://sphincs.org/), which is about to be NIST [standardized](https://csrc.nist.gov/News/2022/pqc-candidates-to-be-standardized-and-round-4).
However, work has stalled as people were busy with other stuff.

After talking to @lukpueh we decided to directly include SPHINCS+ into the "new" Signer API.

I welcome feedback and additional TODOs.

### DIFF
--- a/securesystemslib/formats.py
+++ b/securesystemslib/formats.py
@@ -291,19 +291,16 @@ ECDSAKEY_SCHEMA = SCHEMA.Object(
 # An ED25519 raw public key, which must be 32 bytes.
 ED25519PUBLIC_SCHEMA = SCHEMA.LengthBytes(32)
 
-# An SPHINCS raw public key, which must be 32 bytes.
 SPHINCSPUBLIC_SCHEMA = SCHEMA.LengthBytes(32)
 
 # An ED25519 raw seed key, which must be 32 bytes.
 ED25519SEED_SCHEMA = SCHEMA.LengthBytes(32)
 
-# An SPHINCS raw public key, which must be 32 bytes.
 SPHINCSPRIVATE_SCHEMA = SCHEMA.LengthBytes(64)
 
 # An ED25519 raw signature, which must be 64 bytes.
 ED25519SIGNATURE_SCHEMA = SCHEMA.LengthBytes(64)
 
-# An SPHINCS raw signature, which must be 64 bytes.
 SPHINCSSIGNATURE_SCHEMA = SCHEMA.LengthBytes(7_856)
 
 # An ECDSA signature.
--- a/securesystemslib/formats.py
+++ b/securesystemslib/formats.py
@@ -210,7 +210,7 @@ KEYTYPE_SCHEMA = SCHEMA.OneOf(
         SCHEMA.String("ed25519"),
         SCHEMA.String("ecdsa"),
         SCHEMA.RegularExpression(r"ecdsa-sha2-nistp(256|384)"),
-        SCHEMA.String('sphincs-shake-128s')
+        SCHEMA.String('sphincs')
     ]
 )
 
--- a/securesystemslib/formats.py
+++ b/securesystemslib/formats.py
@@ -210,7 +210,7 @@ KEYTYPE_SCHEMA = SCHEMA.OneOf(
         SCHEMA.String("ed25519"),
         SCHEMA.String("ecdsa"),
         SCHEMA.RegularExpression(r"ecdsa-sha2-nistp(256|384)"),
-        SCHEMA.String('sphincs')
+        SCHEMA.String("sphincs"),
     ]
 )
 
@@ -310,7 +310,7 @@ ECDSASIGNATURE_SCHEMA = SCHEMA.AnyBytes()
 # supported.
 ED25519_SIG_SCHEMA = SCHEMA.OneOf([SCHEMA.String("ed25519")])
 
-SPHINCS_SIG_SCHEMA = SCHEMA.OneOf([SCHEMA.String('sphincs-shake-128s')])
+SPHINCS_SIG_SCHEMA = SCHEMA.OneOf([SCHEMA.String("sphincs-shake-128s")])
 
 # An ed25519 key.
 ED25519KEY_SCHEMA = SCHEMA.Object(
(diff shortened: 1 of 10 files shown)


securesystemslib/formats.py
```diff
@@ -256,19 +256,30 @@
...
 ED25519PUBLIC_SCHEMA = SCHEMA.LengthBytes(32)
 
+# An SPHINCS raw public key, which must be 32 bytes.
+SPHINCSPUBLIC_SCHEMA = SCHEMA.LengthBytes(32)
+
 # An ED25519 raw seed key, which must be 32 bytes.
 ED25519SEED_SCHEMA = SCHEMA.LengthBytes(32)
 
+# An SPHINCS raw public key, which must be 32 bytes.
+SPHINCSPRIVATE_SCHEMA = SCHEMA.LengthBytes(64)
```
jku, line 266: comment seems incorrect

securesystemslib/formats.py
```diff
@@ -256,19 +256,30 @@
...
+
 # An ED25519 raw signature, which must be 64 bytes.
 ED25519SIGNATURE_SCHEMA = SCHEMA.LengthBytes(64)
 
+# An SPHINCS raw signature, which must be 64 bytes.
+SPHINCSSIGNATURE_SCHEMA = SCHEMA.LengthBytes(7_856)
```
jku, line 272: comment byte length mismatch?

jku (commented):
I don't know anything about sphincs so just commenting on cosmetics: there's a couple of code comments that do not seem to match code. I think it would be best to not duplicate information like bytearray length in the comments if the code is clear as it seems to be here

rugo:
> I don't know anything about sphincs so just commenting on cosmetics: there's a couple of code comments that do not seem to match code. I think it would be best to not duplicate information like bytearray length in the comments if the code is clear as it seems to be here

Makes sense. I removed them.

securesystemslib/formats.py
```diff
@@ -193,7 +193,7 @@
 # Supported securesystemslib key types.
 KEYTYPE_SCHEMA = SCHEMA.OneOf(
   [SCHEMA.String('rsa'), SCHEMA.String('ed25519'), SCHEMA.String('ecdsa'),
-   SCHEMA.RegularExpression(r'ecdsa-sha2-nistp(256|384)')])
+   SCHEMA.RegularExpression(r'ecdsa-sha2-nistp(256|384)'), SCHEMA.String('sphincs-shake-128s')])
```
lukpueh, line 196: Does it make sense to call the keytype just `sphincs`? (see discussion about keytype vs. scheme in #239 and #308)

securesystemslib/sphincs_keys.py
```diff
@@ -0,0 +1,88 @@
...
+
+    return shake_128s.verify(data, signature, public_key)
+
+
+if __name__ == '__main__':
+    # The interactive sessions of the documentation strings can
+    # be tested by running 'ed25519_keys.py' as a standalone module.
+    # python -B ed25519_keys.py
+    import doctest
+    doctest.testmod()
```
lukpueh, line 88: No need to add `doctest` boilerplate, there are no doctests in this module, and we wouldn't even run them in CI.

lukpueh (commented):
Thanks for the patch, @rugo! I think what's there is great. There are a few peculiarities that were likely inspired from the existing interfaces (e.g. passing args that are not used, keytype/scheme redundancy), but I think it's okay for the sake of consistency.

Would you mind adding tests to `check_public_interfaces.TestPublicInterfaces.test_keys`, to assert that we fail gracefully if `PySPX` is not installed?

It would also be good to have `interface.generate_and_write_sphincs*`, and `interface.import_sphincs*` functions as we have for other key types. But that can be added in a separate PR.

rugo:
@lukpueh Thanks for the feedback. I added the required tests, changed the `keytype` and removed `doctest`.

Regadring `interface.generate_and_write_sphincs`: I would prefer to address this in a separate PR. This would need more coordination as SPHINCS+ does not have a standardized key OID yet (AFAIK). Decoding a SPHINCS+ key from PEM would therefore require us to create our own non-standard key identifier and format.

rugo:
@lukpueh Could you check the failing test and give me some more insights? I don't really understand what is happening as the error says:

```
subprocess.TimeoutExpired: Command '['gpg', '--detach-sign', '--digest-algo', 'SHA256', '--local-user', 'E8AC80C924116DABB51D4B987CB07D6D2C199C7C', '--homedir', '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmpuwgkkmy9/rsa']' timed out after 3 seconds
```

Which doesn't seem related to what I did. Did the `macos` test case break or am I missing something?

lukpueh:
@rugo: I just merged  #439, which auto-formats the entire code base. The easiest thing might be to re-apply your changes manually on top of the updated master branch, run [auto-formatters commands (w/o `--check --diff` options)](https://github.com/secure-systems-lab/securesystemslib/blob/600aca1091b6c4de048d53ae1d923ca98ff4d773/tox.ini#L47-L49) over them, and push them in a single commit to this branch. I apologize for the inconvenience, and am happy to help resolve any conflicts.

rugo:
Alright. I think I have addressed your points. Additional feedback welcome.

rugo:
Since the pylint test case complained about some minor things I fixed them now.

rugo:
I rebased again to avoid the `requirements.txt` merge conflict.

securesystemslib/keys.py
```diff
@@ -349,6 +350,33 @@ def generate_ed25519_key(scheme="ed25519"):
     return ed25519_key
 
 
+def generate_sphincs_key(scheme="sphincs-shake-128s"):
+    """Generate a SPHINCS+ key pair.
+    Arguments:
+          scheme (str): Name of the scheme as defined in formats.py.
+    Returns:
+        dict: A dictionary containing the ED25519 keys.
```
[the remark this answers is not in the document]
rugo: :see_no_evil: fixed

lukpueh (approved):
💯

PR #429, opened 2022-09-21
Catch timeouts on import of gpg.constants

### LINKED ISSUE #428: Checking for `gpg` availability sometimes times out
When importing `constants.py`, `securesystemslib` checks for GPG by calling `is_available_gnupg()`"

https://github.com/secure-systems-lab/securesystemslib/blob/v0.24.0/securesystemslib/gpg/constants.py#L26-L32

In go-tuf, we've seen this time out a few times. For instance,

https://github.com/theupdateframework/go-tuf/actions/runs/3100521501/jobs/5020899719

You might think: that's ridiculous! How would a call to GPG time out? And the answer is: GitHub's hosted Windows runners experience intermittent extreme slowness. I don't believe GPG is even present on these machines, so it's really Windows taking >3 seconds to tell you that the binary is not found. Regardless, it happens.

My preference would be to avoid doing this on import at all, but I think just catching `subprocess.TimeoutExpired` would fix it.

znewman01:
Sometimes, in some systems, not pointing any fingers but Windows, just finding out that there's no `gpg` available takes >3 seconds.

In that case, we now just assume that there's no GPG because a binary that takes 3 seconds even to start isn't any better than no binary at all.

Fixes #428.

### DIFF
--- a/securesystemslib/gpg/constants.py
+++ b/securesystemslib/gpg/constants.py
@@ -28,7 +28,7 @@ def is_available_gnupg(gnupg):
   try:
     process.run(gpg_version_cmd, stdout=process.PIPE, stderr=process.PIPE)
     return True
-  except OSError:
+  except (OSError, subprocess.TimeoutExpired):
     return False
 
 


joshuagl:
CI failures are because the subprocess module isn't imported in gpg.constants. I don't know why this only fails on macOS 😕

znewman01:
Yeah, that's...bizarre. I don't think we're going to merge this  (see discussion in the linked bug) but that might point to some issue in tests.

CC @adityasaky

znewman01:
Oh! The `except` line doesn't execute if the `try` succeeds. That is:

```python
try:
    pass
except ThisExceptionDoeNotExist
```

will always succeed

joshuagl:
Of course, that does suggest we might need more comprehensive CI.

PR #431, opened 2022-09-22
Create backwards-compatible RSA PSS sigs by default

### LINKED ISSUE #430: Maintain backwards-compatibility by default in RSA PSS signatures

In #421, we started creating RSA PSS signatures only with the maximum possible salt length. Unfortunately, this breaks backwards-compatibility when verifying these new signatures with securesystemslib<[0.24.0](https://github.com/secure-systems-lab/securesystemslib/releases/tag/v0.24.0).

**Current behavior**:

Creating RSA PSS signatures ==0.24.0 breaks verification in <0.24.0.

**Expected behavior**:

Creating RSA PSS signatures in >0.24.0 shouldn't break verification by default in <0.24.0 (and above).

trishankatdatadog:
Fixes: #430 

This PR fixes #430 by creating RSA PSS signatures with a salt length that _defaults_ to size of the output of the hash function used in the signature. End-user libraries/applications such as [python-tuf](https://github.com/theupdateframework/python-tuf) and [python-in-toto](https://github.com/in-toto/in-toto) are able to override this default if so desired.

### DIFF
--- a/tests/test_rsa_keys.py
+++ b/tests/test_rsa_keys.py
@@ -159,34 +159,34 @@ class TestRSA_keys(unittest.TestCase):
     rsa_scheme = 'rsassa-pss-sha256'
     data = 'The ancients say the longer the salt, the more provable the security'.encode('utf-8')
 
-    # Old-style signature: use the hash length as the salt length.
-    old_signature, _ = securesystemslib.rsa_keys.create_rsa_signature(private_rsa, data)
+    # Default signature: use the digest length as the salt length.
+    def_sig, _ = securesystemslib.rsa_keys.create_rsa_signature(private_rsa, data)
 
-    # New-style signature: use the maximum salt length.
-    new_signature, _ = securesystemslib.rsa_keys.create_rsa_signature(private_rsa, data,
-                                                                      salt_length=padding.PSS.MAX_LENGTH)
+    # "Max" signature: use the maximum salt length.
+    max_sig, _ = securesystemslib.rsa_keys.create_rsa_signature(private_rsa, data,
+                                                                salt_length=padding.PSS.MAX_LENGTH)
 
-    # Test that old verifiers can verify the old-style signature.
-    old_sign_old_ver = securesystemslib.rsa_keys.verify_rsa_signature(old_signature, rsa_scheme,
-                                                                      public_rsa, data,
-                                                                      salt_length=padding.PSS.DIGEST_LENGTH)
-    self.assertTrue(old_sign_old_ver)
-
-    # Test that new verifiers can also automatically verify the old-style signature.
-    old_sig_new_ver = securesystemslib.rsa_keys.verify_rsa_signature(old_signature, rsa_scheme,
-                                                                     public_rsa, data)
-    self.assertTrue(old_sig_new_ver)
-
-    # Test that old verifiers cannot automatically verify new-style signatures.
-    new_sig_old_ver = securesystemslib.rsa_keys.verify_rsa_signature(new_signature, rsa_scheme,
+    # Test that *old* verifiers can verify the *default* signature.
+    def_sig_old_ver = securesystemslib.rsa_keys.verify_rsa_signature(def_sig, rsa_scheme,
                                                                      public_rsa, data,
                                                                      salt_length=padding.PSS.DIGEST_LENGTH)
-    self.assertFalse(new_sig_old_ver)
+    self.assertTrue(def_sig_old_ver)
 
-    # Test that new verifiers can verify new-style signatures.
-    new_sig_new_ver = securesystemslib.rsa_keys.verify_rsa_signature(new_signature, rsa_scheme,
+    # Test that *new* verifiers can also automatically verify the *default* signature.
+    def_sig_new_ver = securesystemslib.rsa_keys.verify_rsa_signature(def_sig, rsa_scheme,
                                                                      public_rsa, data)
-    self.assertTrue(new_sig_new_ver)
+    self.assertTrue(def_sig_new_ver)
+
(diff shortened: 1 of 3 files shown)


yzhan289 (approved):
🚀

tests/test_rsa_keys.py
```diff
@@ -161,23 +159,34 @@ def test_verify_rsa_pss_sha256(self):
...
-    for signature in (old_signature, new_signature):
-      verified = securesystemslib.rsa_keys.verify_rsa_signature(signature, rsa_scheme,
-                                                                public_rsa, data)
-      self.assertTrue(verified)
+    # Default signature: use the digest length as the salt length.
+    def_sig, _ = securesystemslib.rsa_keys.create_rsa_signature(private_rsa, data)
+
+    # "Max" signature: use the maximum salt length.
+    max_sig, _ = securesystemslib.rsa_keys.create_rsa_signature(private_rsa, data,
+                                                                salt_length=padding.PSS.MAX_LENGTH)
```
adityasaky, line 167: nit: alignment

adityasaky (approved):

tests/test_rsa_keys.py
trishankatdatadog: What about the alignment? I wasn't sure what was the convention. Really we need to fix linting as part of #432.

trishankatdatadog:
Should we also update `interface.(create|verify)_signature` to pass a salt length?

tests/test_rsa_keys.py
SantiagoTorres: I'm not sure what the problem is. There used to be different expectations on the guidelines but it appears to me that now it just tells "align to open parentheses", @adityasaky am I missing sth?

adityasaky: I was looking at https://github.com/secure-systems-lab/code-style-guidelines/blob/master/python.md and I guess we should use `from...import...` to avoid this in the first place?

SantiagoTorres: I'm a little nervous about that overhaul considering the whole test above uses it  as such. I'd assume we could go a linter pass sometime later?

adityasaky:
> Should we also update interface.(create|verify)_signature to pass a salt length?

So that should go in `securesystemslib.keys` but I'm wondering if that should just stick to the default since this would only apply to RSA keys while the API is for all types of keys. What do you think?

SantiagoTorres:
I wonder if we want to at all touch the saltlength. I'm unsure why the underlying api had to change (not in sslib, but cryptography). Is there any backstory to it?

jku:
> This PR fixes #430 by creating RSA PSS signatures with a salt length that defaults to size of the output of the hash function used in the signature. End-user libraries/applications such as python-tuf and python-in-toto are able to override this default if so desired.

So I'm trying to understand the complete picture:
* originally we created sigs with salt length that matches digest length. Originally we verified sigs assuming salt length of digest length
* then in #422 we started creating sigs with "maximum salt length" and verifying sigs with "automatic" salt length. This was not backwards compatible as old verifier code expects "digest length"
* this PR creates sigs by default with digest length (so reverts to original), and verifies sigs by default with "automatic" (keeps the #422 functionality). In addition it passes the buck to caller by adding salt length as an argument

So end result seems to be
* verifiers can now work with RSA PSS sigs produced with different salt length :heavy_check_mark: 
* theoretically callers can specify the salt length ...  but will they?
  * for verify, shouldn't "automatic" always work? what is the purpose of specifying the length?
  * for sign, the same problem applies to calling code as you encountered: if old code can't handle different salt lengths then python-tuf and intoto can't change the default either, until all old code is dead and buried
  
So I suppose my question is, is the added API surface needed? Could sslib just create sigs like it used to originally before #422 (and file an issue for future maintainers to change it when old code really is dead) and just let verify always use automatic length?

tests/test_rsa_keys.py
adityasaky: sgtm.

trishankatdatadog:
> * for verify, shouldn't "automatic" always work? what is the purpose of specifying the length?

Reasons:
1. To simplify testing.
2. For end-users who are interested in fixing the salt length if they so wish (because I believe the salt length is attacker-controlled since it's part of the signature, although the impact is only a DoS attack unless there is some strange crypto attack which wouldn't surprise me).

> Could sslib just create sigs like it used to originally before #422

That's exactly what the signing does by default now.

I don't see any harm in merging this PR as it is: users who override the defaults presumably know what they are doing. Any objections?

trishankatdatadog:
The biggest Q is whether we are OK with the default salt length (digest length). I think it's since, since nobody [knows](https://crypto.stackexchange.com/a/1222) how to break even empty salt lengths apparently, but I think it's not a bad idea to allow those who know what they are doing to override defaults if they so wish.

OTOH, I'm not against simply for now: (1) fixing signatures to the digest length, and (2) setting salt length automatically during verification. Votes?

lukpueh:
>  (1) fixing signatures to the digest length, and (2) setting salt length automatically during verification. Votes?

+1

trishankatdatadog:
> +1

Ok, so no overriding of these defaults, correct?

lukpueh:
I intend to create a release next week to get out a different feature (#231). Happy to include a fix for this issue here if provided.

trishankatdatadog:
> I intend to create a release next week to get out a different feature (#231). Happy to include a fix for this issue here if provided.

@yzhan289 Could you please help take care of this? Thanks!

lukpueh:
Closing in favor of https://github.com/secure-systems-lab/securesystemslib/pull/436

Issue #247, opened 2020-06-25
Don't use CLI gpg


KOLANICH:
Usage of CLI tools may be dangerous. It is also not a very clean and powerful approach because CLIs are quite limited.

The better approach is to use some libs directly.

gpgme (debian package `python3-gpg` and is imported as `gpg`) is claimed to be LGPL, but it is a bit strange, since gpg itself is GPL.

There is also https://github.com/SecurityInnovation/PGPy , a pure python impl that is pretty abandoned.

Also BouncyCastle can be used via JPype.

joshuagl:
Note that gpgpme *is* a wrapper of the gpg CLI tool, with often [surprising behaviour](https://github.com/justinsteven/advisories/blob/master/2020_fwupd_dangling_s3_bucket_and_CVE-2020-10759_signature_verification_bypass.md#gpgme-gnupg-made-easy). Furthermore, because it's a C library which would need to be installed with the Python bindings it makes our installation story, particularly on non-Linux systems, much more difficult.

lukpueh:
Thanks for submitting this issue, @KOLANICH. We did review alternatives back when we implemented this. Unfortunately, *gpgme* did not work out of the box and, as @joshuagl points out, would have added more native dependencies.

Our use case here is to provide a bridge between securesystemslib infrastructure (data containers, verifications routines) and gpg keys, to developers that already have and use gpg e.g. for mail, git, software signing, etc. ...
Wrapping the cli seemed the most feasible approach and suited our needs. That said, we welcome any feedback, especially if it is related to the security of our project.

lukpueh:
cc @SantiagoTorres

KOLANICH:
>Note that gpgpme is a wrapper of the gpg CLI tool

I doubt it very much. Its source code doesn't look like that. Could you prove it?

>Wrapping the cli seemed the most feasible approach and suited our needs.

There was already a wrapper for cli. A pure python one. Called python-gnupg. With vulnrs. Please never use CLI for passing any args.

SantiagoTorres:
> I doubt it very much. Its source code doesn't look like that. Could you prove it?

Sure [here](https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gpgme.git;a=blob;f=src/engine-gpg.c;h=173e940c205bc81dc0f0e931241d0bc7719c17e2;hb=HEAD#l1643)

SantiagoTorres:
[...]

PR #433, opened 2022-09-30
Replace deprecated distutils.version.StrictVersion.

znewman01:
distutils is deprecated as of Python 3.10. We were only using it for `StrictVersion`, which parses a version string and lets you order the result.

The [official recommendation] from the packaging team is to replace `StrictVersion` with a dependency on the `packaging` package, but:

1. The Python version parser was a little different from GPG versioning anyway because it's made for Python versions (it had support for "prerelease tags").
2. I wanted to avoid adding dependencies if possible.

[official recommendation]: https://github.com/pypa/packaging/issues/520

This PR adds a new `Version` class in `securesystemslib.gpg.util` and uses that instead.

It's a breaking change to `gpg.util.get_version()`, but AFAICT that wasn't ever intended for public consumption.

Fixes: N/A

- [X] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [X] Tests have been added for the bug fix or new feature
- [X] Docs have been added for the bug fix or new feature

### DIFF
--- a/tests/test_gpg.py
+++ b/tests/test_gpg.py
@@ -37,11 +37,12 @@ import cryptography.hazmat.primitives.serialization as serialization
 import cryptography.hazmat.backends as backends
 import cryptography.hazmat.primitives.hashes as hashing
 
+from securesystemslib import exceptions
 from securesystemslib import process
 from securesystemslib.gpg.functions import (create_signature, export_pubkey,
     verify_signature, export_pubkeys)
 from securesystemslib.gpg.util import (get_version, is_version_fully_supported,
-    get_hashing_class, parse_packet_header, parse_subpacket_header)
+    get_hashing_class, parse_packet_header, parse_subpacket_header, Version)
 from securesystemslib.gpg.rsa import create_pubkey as rsa_create_pubkey
 from securesystemslib.gpg.dsa import create_pubkey as dsa_create_pubkey
 from securesystemslib.gpg.eddsa import create_pubkey as eddsa_create_pubkey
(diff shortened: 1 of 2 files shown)


jku:
This may be off topic for your PR but since you're touching this you may have an opinion: I wonder how much work would it be to use an actual API for the ~2 gpg commands we need? Spawning new processes via a command line makes me feel dirty (and is a great source of testing trouble).

joshuagl:
Some related discussion in https://github.com/secure-systems-lab/securesystemslib/issues/428 
tl;dr – we don't want to run gpg commands, but we don't want to add dependencies either. Discussion in #428 strives to find a middle ground.

znewman01:
> Spawning new processes via a command line makes me feel dirty (and is a great source of testing trouble).

I'm generally sympathetic to that point of view. While I do think minimizing dependencies is nice, to my mind, there's not much difference between "depends on GPG" and "depends on GPG+Python bindings to GPG."

That said, my most immediate goal is to fix #428 and reduce flakiness in CI for go-tuf; I think the very minor change in #429 would fix it but we need to couple that with some extra testing (#434).

tests/test_gpg.py
```diff
@@ -809,5 +817,30 @@ def test_verify_short_signature(self):
...
+        ('1.3.22', Version(1, 3, 22)),
+    ]:
+      self.assertEqual(Version.from_string(value), expected)
+      self.assertEqual(str(expected), value)
+
+  def test_version_from_string_invalid(self):
+    """Version.from_string rejects invalid inputs."""
+    for value in [
+        '1.3',
+        '1.33.0',
```
jku, line 837: These are both valid for StrictVersion.

Obviously we don't need to re-implement StrictVersion here (we know this is a hack, it just has to work well enough for this specific case) but at least the latter one seems strange to me -- is this just because gpg historically hasn't used multidigit minor versions?

jku (commented):
Seems reasonable to me, especially if we consider the whole _"library spawns other binaries and parses output"_ as a hack that securesystemslib should probably get rid of.

Left a comment but even that does not seem critical (I don't think it'll break anything in this decade)

lukpueh:
> I wonder how much work would it be to use an actual API like https://pypi.org/project/python-gnupg/ or https://pypi.org/project/gpg/

Too much work for the benefit. See https://github.com/secure-systems-lab/securesystemslib/issues/247 and https://wiki.python.org/moin/GnuPrivacyGuard 

Tl;DR: Both packages you suggest also just spawn a process to run  the `gpg` command, the former via `subprocess`, the latter via GPGME C binding. So far we haven't found a good alternatives to our "hack".

tests/test_gpg.py
lukpueh: Good question. Unfortunately, I can't remember if it was just an assumption based on the existing version progression (e.g. last 1.x  was 1.9), or if there was a stronger indication somewhere. Happy to change, although not necessarily in this PR, which just uses the regex that was already there.

lukpueh (approved):
Great work. Looks more robust, than it was before. Regarding backwards-compatibility, I confirm that `gpg.util.get_version()` was not intended as public API.

tests/test_gpg.py
jku: oh you're right: no new bugs then :)

PR #434, opened 2022-09-30
Double-check that GPG tests run where expected.

znewman01:
The test suite passes even if GPG is not available, because GPG-dependent tests are skipped if GPG is not present.

This change asserts the availability of GPG, so CI environments that expect GPG will notice if it goes away unexpectedly rather than silently skipping the GPG tests.

Addresses (does not fix): #428 

- [X] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [X] Tests have been added for the bug fix or new feature
- [X] Docs have been added for the bug fix or new feature

### DIFF
--- a/tests/check_gpg_available.py
+++ b/tests/check_gpg_available.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python
+
+"""
+<Program Name>
+  check_gpg_available.py
+
+<Author>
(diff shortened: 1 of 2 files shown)


lukpueh (approved):
Great work, thanks!

PR #436, opened 2022-10-13
Revert to digest sized salts in rsassa-pss signing

### LINKED ISSUE #430: Maintain backwards-compatibility by default in RSA PSS signatures

lukpueh:
Fixes #430, Supersedes #431

---

In #422 we changed `create_rsa_signature` to use the maximum salt length available, instead of the digest length, when creating rsassa-pss signatures, and adapted `verify_rsa_signature` to infer the salt length automatically.

This made it impossible for users of the old verify function, which could only handle digest sized salts, to verify signatures created by users of the new signing function (see #430).

Since the advantage of max salt lengths is mostly academic, this patch reverts  `create_rsa_signature` to use digest sized salts.

Note that we now use the `padding.PSS.DIGEST_LENGTH` constant instead of passing the actual digest length, as we did before. Using the constant has the same result, but is recommended by the library documentation.

Also note that the patch does not revert the `verify_rsa_signature` part of #422. This allows verifying signatures created with the securesystemslib release that used max salt lengths, or created outside of securesystemslib.

### DIFF
--- a/securesystemslib/rsa_keys.py
+++ b/securesystemslib/rsa_keys.py
@@ -326,11 +326,10 @@ def create_rsa_signature(private_key, data, scheme='rsassa-pss-sha256'):
     if scheme.startswith('rsassa-pss'):
       # Generate an RSSA-PSS signature.  Raise
       # 'securesystemslib.exceptions.CryptoError' for any of the expected
-      # exceptions raised by pyca/cryptography. 'salt_length' is set to
-      # the maximum length available.
+      # exceptions raised by pyca/cryptography.
       signature = private_key_object.sign(
           data, padding.PSS(mgf=padding.MGF1(digest_obj.algorithm),
-          salt_length=padding.PSS.MAX_LENGTH), digest_obj.algorithm)
+          salt_length=padding.PSS.DIGEST_LENGTH), digest_obj.algorithm)
 
     elif scheme.startswith('rsa-pkcs1v15'):
       # Generate an RSA-PKCS1v15 signature.  Raise
--- a/tests/test_rsa_keys.py
+++ b/tests/test_rsa_keys.py
@@ -27,6 +27,7 @@ import securesystemslib.hash
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.asymmetric import padding
+from cryptography.hazmat.primitives.hashes import SHA256
 from cryptography.hazmat.primitives.serialization import load_pem_private_key
 
 public_rsa, private_rsa = securesystemslib.rsa_keys.generate_rsa_public_and_private()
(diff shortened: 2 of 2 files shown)


lukpueh:
@trishankatdatadog, @yzhan289, does this work for you?

securesystemslib/rsa_keys.py
```diff
@@ -326,11 +326,10 @@ def create_rsa_signature(private_key, data, scheme='rsassa-pss-sha256'):
     if scheme.startswith('rsassa-pss'):
       # Generate an RSSA-PSS signature.  Raise
       # 'securesystemslib.exceptions.CryptoError' for any of the expected
-      # exceptions raised by pyca/cryptography. 'salt_length' is set to
-      # the maximum length available.
+      # exceptions raised by pyca/cryptography.
       signature = private_key_object.sign(
           data, padding.PSS(mgf=padding.MGF1(digest_obj.algorithm),
-          salt_length=padding.PSS.MAX_LENGTH), digest_obj.algorithm)
+          salt_length=padding.PSS.DIGEST_LENGTH), digest_obj.algorithm)
```
yzhan289, line 332: The [original code](https://github.com/secure-systems-lab/securesystemslib/pull/422/files#diff-3d4df320352b04eb09fa4f56e48516c9ae514decd0685775301c02a6f60bdb9dL332) had `digest_obj.algorithm.digest_size` instead, is there a difference here?

lukpueh: See PR description / commit mesesage:
 
> Note that we now use the `padding.PSS.DIGEST_LENGTH` constant instead of passing the actual digest length, as we did before. Using the constant has the same result, but is recommended by the library documentation.

lukpueh: I locally created two signature one with the constant and one with the digest length and can confirm that they do the same thing.

I also checked the code.

lukpueh: https://github.com/pyca/cryptography/blob/25d794e079b44e6fac23791ca51da3a3fb9cb526/src/cryptography/hazmat/backends/openssl/rsa.py#L53-L54

tests/test_rsa_keys.py
```diff
@@ -157,26 +158,30 @@ def test_verify_rsa_signature(self):
           scheme, public_rsa, data))
 
 
-  def test_verify_rsa_pss_sha256(self):
+  def test_verify_rsa_pss_different_salt_lengths(self):
     rsa_scheme = 'rsassa-pss-sha256'
-    data = 'The ancients say the longer the salt, the more provable the security'.encode('utf-8')
+    data = 'The ancients say, salt length does not matter that much'.encode('utf-8')
```
trishankatdatadog, line 163: Mathematically speaking, apparently this is not strictly true, although practically it might be the case 😉

securesystemslib/rsa_keys.py
trishankatdatadog: Right: I think `padding.PSS.DIGEST_LENGTH` is simply an internal `cryptography` constant that does look like it's used to end up using `digest_obj.algorithm.digest_size` in the end.

tests/test_rsa_keys.py
```diff
@@ -27,6 +27,7 @@
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.asymmetric import padding
+from cryptography.hazmat.primitives.hashes import SHA256
```
trishankatdatadog, line 30: Unused constant? We really need a linter.

tests/test_rsa_keys.py
lukpueh: Yeah, I would also not trust every word the ancients say. ;)

tests/test_rsa_keys.py
lukpueh: Good catch

trishankatdatadog: What's a "flag day"?

joshuagl: Ultimately, I think a linter (and using a standard coding style) makes the code easier to work on so I am in favour of the flag day (event where the entire codebase changes to the new style, making a large diff with only mechanical changes which invalidate patches and make revision control "noise").

joshuagl (approved):

trishankatdatadog (approved):

securesystemslib/rsa_keys.py
yzhan289: Thanks! Oof my bad I didn't read the description first 🤦

trishankatdatadog:
Thanks very much, @lukpueh and friends! No rush, but in which release could we expect this, please?

PR #442, opened 2022-10-26
Add Google Cloud KMS signing capability

jku:
Add support for Google Cloud KMS signing. The use case for this is TUF repositories with online keys. Removing the private key material from the repository software is a good thing for security: a compromise of the repository software will then only allow signing during the compromise but won't allow stealing the private keys.

The included Google implementation depends on implicit authentication: the runtime environment should contain authentication credentials that google.cloud.kms library is able to find.

Example of a test run: https://github.com/jku/securesystemslib/actions/runs/3384917757/jobs/5622619919
This uses `google-github-actions/auth` to authenticate to Google Cloud (this is the implicit authentication part): the github project token is configured to have _Signer/Verifier_ permissions in Google KMS.

The usability should improve if #447 or something like it gets implemented: 
* the private key URI for this would be something like 
`gcpkms://projects/python-tuf-kms/locations/global/keyRings/securesystemslib-tests/cryptoKeys/ecdsa-sha2-nistp256/cryptoKeyVersions/1`
* a function like `Key.get_payload_hash_algorithm()` would be helpful at that point (needed by signer but it doesn't have to be public)

### DIFF
--- a/.github/workflows/test-kms.yml
+++ b/.github/workflows/test-kms.yml
@@ -16,11 +16,12 @@ jobs:
 
     steps:
       - name: Checkout securesystemslib
-        uses: actions/checkout@v2
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
 
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@13ae5bb136fac2878aff31522b9efb785519f984
         with:
+          python-version: '3.x'
           cache: 'pip'
           cache-dependency-path: 'requirements*.txt'
 
--- a/.github/workflows/test-kms.yml
+++ b/.github/workflows/test-kms.yml
@@ -1,0 +1,38 @@
+name: Run KMS tests
+
+on:
+  push:
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  test-kms:
+    runs-on: ubuntu-latest
+
+    permissions:
+      id-token: 'write' # for OIDC auth for GCP authentication
+
+    steps:
+      - name: Checkout securesystemslib
+        uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          cache: 'pip'
+          cache-dependency-path: 'requirements*.txt'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install --upgrade tox
+
+      - name: Authenticate to Google Cloud
+        uses: google-github-actions/auth@c4799db9111fba4461e9f9da8732e5057b394f72
+        with:
+          token_format: access_token
+          workload_identity_provider: projects/843741030650/locations/global/workloadIdentityPools/securesystemslib-tests/providers/securesystemslib-tests
+          service_account: securesystemslib@python-tuf-kms.iam.gserviceaccount.com
+
+      - run: tox -e kms
--- a/.github/workflows/test-kms.yml
+++ b/.github/workflows/test-kms.yml
@@ -12,6 +12,7 @@ jobs:
 
     permissions:
       id-token: 'write' # for OIDC auth for GCP authentication
+      issues: 'write' # for filing an issue on failure
 
     steps:
       - name: Checkout securesystemslib
@@ -36,3 +37,25 @@ jobs:
           service_account: securesystemslib@python-tuf-kms.iam.gserviceaccount.com
 
       - run: tox -e kms
+
+      - name: File an issue on failure
+        if: ${{ failure() }}
+        uses: actions/github-script@d556feaca394842dc55e4734bf3bb9f685482fa0
+        with:
+          script: |
+              const repo = context.repo.owner + "/" + context.repo.repo
+              const issues = await github.rest.search.issuesAndPullRequests({
+                q: "KMS+tests+failed+in:title+state:open+type:issue+repo:" + repo,
+              })
+              if (issues.data.total_count > 0) {
+                console.log("Issue open already, not creating.")
+              } else {
+                await github.rest.issues.create({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  title: "KMS tests failed",
+                  body: "Hey, it seems KMS tests have failed, please see - [workflow run](" +
+                        "https://github.com/" + repo + "/actions/runs/" + context.runId + ")"
+                })
+                console.log("New issue created.")
+              }
--- a/mypy.ini
+++ b/mypy.ini
@@ -9,3 +9,7 @@ files =
 # Supress error messages until enough modules
 # are type annotated
 follow_imports = silent
+
+# let's not install typeshed annotations for GCPSigner
+[mypy-google.*]
+ignore_missing_imports = True
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ Issues = "https://github.com/secure-systems-lab/securesystemslib/issues"
 
 [project.optional-dependencies]
 crypto = ["cryptography>=37.0.0"]
+gcpkms = ["google-cloud-kms"]
 pynacl = ["pynacl>1.2.0"]
 PySPX = ["PySPX==0.5.0"]
 
--- a/requirements-kms.txt
+++ b/requirements-kms.txt
@@ -1,0 +1,1 @@
+google-cloud-kms
--- a/securesystemslib/signer/__init__.py
+++ b/securesystemslib/signer/__init__.py
@@ -13,3 +13,35 @@ from securesystemslib.signer._signer import (
     Signer,
     SSlibSigner,
 )
+
+# Register supported private key uri schemes and the Signers implementing them
+SIGNER_FOR_URI_SCHEME.update(
+    {
+        SSlibSigner.ENVVAR_URI_SCHEME: SSlibSigner,
+        SSlibSigner.FILE_URI_SCHEME: SSlibSigner,
+    }
+)
+
+# Register supported key types and schemes, and the Keys implementing them
+KEY_FOR_TYPE_AND_SCHEME.update(
+    {
+        ("ecdsa", "ecdsa-sha2-nistp256"): SSlibKey,
+        ("ecdsa", "ecdsa-sha2-nistp384"): SSlibKey,
+        ("ecdsa-sha2-nistp256", "ecdsa-sha2-nistp256"): SSlibKey,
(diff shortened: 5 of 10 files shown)


securesystemslib/kms_signer.py
```diff
@@ -0,0 +1,106 @@
...
+
+        if key["keytype"] == "rsa":
+            # hash algorithm is encoded as last scheme portion
+            return key["scheme"].split("-")[-1]
+        if key["keytype"] in [
+            "ecdsa",
+            "ecdsa-sha2-nistp256",
+            "ecdsa-sha2-nistp384",
+        ]:
+            # nistp245 uses sha-256, nistp384 uses sha-384
```
jku, line 59: nistp245 :thinking:

securesystemslib/kms_signer.py
```diff
@@ -0,0 +1,106 @@
...
+    GCPSigner keyref is a fully qualified CryptoKeyVersion name prefixed with "gcpkms://"
+
+    GCPSigner is not user facing API but an implementation detail of KMSSigner.
+    """
+
+    KEYREF_SCHEME = "gcpkms://"
+
+    def __init__(self, keyref: str, public_key: Dict[str, Any]):
+        # TODO instead of hash_algo and keyid, could just take the public key dict
+        # as input and figure out the algorithm here?
```
jku, line 87: oops, I did this already: comment is obsolete

jku:
The discussion that maybe makes sense at this point is private key abstraction in general: Since we don't have a common abstraction for private key identification, we end up with having to change code when the keytype changes:

```
SSLibSigner(key_dict) # here key_dict is expected to include public and private key 
GPGSigner(gpg_key_id)
KMSSigner.from_keyref(keyref, key_dict)
```

If there was a "configurable private key" the code wouldn't have to change -- assume `privkey` here is something we read from configuration:
`Signer.from_private_key(privkey, key_dict)`

This would act as a factory method/dispatcher that returned the correct Signer subclass -- see KMSSigner in this PR for an example.

 I wonder if the simple  "keyref" string approach in this PR would actually work in general? The idea here is that the public key is available -- we only need the private key access with keyref: 
private key references:
* GCP KMS: 
  "gcpkms://projects/openssf/locations/global/keyRings/securesystemslib-test-keyring/cryptoKeys/securesystemslib-test-key/cryptoKeyVersions/1"
* GPG
  "gpg://C5A0ABE6EC19D0D65F85E2C39BE9DF5131D924E9"
* keyval.private content 
  "raw://788b012147f9afe9c1a151...f94dc32a281baa"
* file with keyval.private content
  "file:///home/jku/rsa_key"

The point of keeping this as "just a string" is to make it easy to load the private key identifier from configuration without touching code (and would be easy to expose via CLI tools if needed)

securesystemslib/kms_signer.py
```diff
@@ -0,0 +1,104 @@
...
+            "ecdsa",
+            "ecdsa-sha2-nistp256",
+            "ecdsa-sha2-nistp384",
+        ]:
+            # nistp256 uses sha-256, nistp384 uses sha-384
+            bits = key["scheme"].split("-nistp")[-1]
+            return f"sha{bits}"
+        if key["keytype"] == "ed25519" and key["scheme"] == "ed25519":
+            # specification default is sha-512
+            return "sha512"
```
jku, line 64: I'm not sure where I got this impression from...

jku: Hmm, the spec defines sha512 and the identity function as the prehash options. Using identity (i.e. no hashing) in this case would be silly...

However, spec uses "ed25519" to mean the identity prehashed scheme, and "ed25519ph" as the SHA512 prehashed scheme: It's entrirely possible that our current implementation does the former and then we would have to add a new scheme to the ed25519 keytype.

Considering that GCP KMS does not even support ed25519, I think I'll just remove this if-clause

jku:
So after a bit of consideration, I think this makes sense:
* generic private key improvements are in #447: I will remove the whole KMSSigner abstraction from this PR with the assumption that those issues will be solved in 447
* testing is still an issue... I will try to find out if there is a low-effort way to get a testing project on GCP

jku:
Oops. I thought I had figured out a way to allow running KMS tests on PRs safely but hit a very practical issue: authenticating to GCP requires the OIDC token from GitHub which understandably isn't given to PRs that come from forks.

So I don't think we can run the KMS test on PRs safely. We can run it on pushes but that's not as nice :(

jku:
Here is the test run on my repo https://github.com/jku/securesystemslib/actions/runs/3384917757/jobs/5622619919 (the GCP setup currently accepts requests from both "jku/securesystemslib" and "secure-systems-lab/securesystemslib")

jku:
Ok, final change: added issue filing on failure (otherwise no-one notices when this fails since it does not run on PRs from forks).

Setting non-draft now.

joshuagl:
@kairoaraujo would love your input on the design here, especially:
* URI scheme key reference
* Importing existing keys (i.e. SSH key or YubiKey) is not yet handled, any thoughts on that?

.github/workflows/test-kms.yml
```diff
@@ -0,0 +1,61 @@
...
+  test-kms:
+    runs-on: ubuntu-latest
+
+    permissions:
+      id-token: 'write' # for OIDC auth for GCP authentication
+      issues: 'write' # for filing an issue on failure
+
+    steps:
+      - name: Checkout securesystemslib
+        uses: actions/checkout@v2
```
lukpueh, line 19: Should we pin to commit as we do in ci.yml now?
```suggestion
        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
```

.github/workflows/test-kms.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
```
lukpueh, line 22:
```suggestion
        uses: actions/setup-python@13ae5bb136fac2878aff31522b9efb785519f984
```

.github/workflows/test-kms.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install --upgrade tox
+
+      - name: Authenticate to Google Cloud
+        uses: google-github-actions/auth@c4799db9111fba4461e9f9da8732e5057b394f72
+        with:
+          token_format: access_token
+          workload_identity_provider: projects/843741030650/locations/global/workloadIdentityPools/securesystemslib-tests/providers/securesystemslib-tests
```
lukpueh, line 36: Do you need to adopt the service name change (f48786f182e5ddf729d2fba7f7fdff3d21225ab3) in the `workload_identity_provider` path? (I did not check if or how they are related)

.github/workflows/test-kms.yml
```diff
@@ -0,0 +1,61 @@
...
+              const issues = await github.rest.search.issuesAndPullRequests({
+                q: "KMS+tests+failed+in:title+state:open+type:issue+repo:" + repo,
+              })
+              if (issues.data.total_count > 0) {
+                console.log("Issue open already, not creating.")
+              } else {
+                await github.rest.issues.create({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  title: "KMS tests failed",
```
lukpueh, line 56: minor nit: Is it worth adding an inline comment that this string is used above in the query?

securesystemslib/signer.py
```diff
@@ -266,3 +279,69 @@ def sign(self, payload: bytes) -> GPGSignature:
...
+    credentials. These will be found by google.cloud.kms, see
+    https://cloud.google.com/docs/authentication/getting-started
+    (and https://github.com/google-github-actions/auth for the relevant
+    GitHub action).
+
+    Arguments:
+        gcp_keyid: Fully qualified GCP KMS key name, like
+            projects/python-tuf-kms/locations/global/keyRings/securesystemslib-tests/cryptoKeys/ecdsa-sha2-nistp256/cryptoKeyVersions/1
+        hash_algo: Payload hashing algorithm
+        keyid: The keyid to be used in the returned Signature
```
lukpueh, line 301: Can't remember the current plan for #456, but if we require Signers to include the public key, the keyid and hash algo don't have to be passed by themselves, right?

lukpueh (approved):
This is super cool and I have nothing to add, except some very minor nits. Feel free to address or ignore as you see fit. Also, let me know if you want to defer merging until after #456.

jku:
> let me know if you want to defer merging until after https://github.com/secure-systems-lab/securesystemslib/pull/456.

I was thinking no need ... but since I'm now going to suggest splitting the code to internal submodules maybe that is indeed a good idea -- let's see how "easy" 456 looks to merge  once I get a non-draft version ready

.github/workflows/test-kms.yml
jku: I think no: service_account is the actual GCP service account that has the permissions to sign. workload_identity_provider is what allows our Github OIDC identity to act as the service account. So two different things that are only connected by "service account impersonation", not by name.

jku:
* rebased on both master and privkeyuri (so it's not really 24 commits, don't worry)
* Implemented the private key uri "gcpkms": works like charm (with a small tweak to the dispatching)
* the KMS test "works on my github project": https://github.com/jku/securesystemslib/actions/runs/3583834641/jobs/6029741653

this is now waiting for #456 merge

jku:
Changes since the first review:
* action pinning fixed per review comment
* uses priv key uri now (constructor changed accordingly)
* signer and key dispatch also tweaked slightly to make it easier to implement signers/keys in other files

Can I ask you to have another look @lukpueh?

.github/workflows/test-kms.yml
```diff
@@ -0,0 +1,61 @@
...
+    permissions:
+      id-token: 'write' # for OIDC auth for GCP authentication
+      issues: 'write' # for filing an issue on failure
+
+    steps:
+      - name: Checkout securesystemslib
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+
+      - name: Set up Python
+        uses: actions/setup-python@13ae5bb136fac2878aff31522b9efb785519f984
```
jku, line 22: ugh, I'm actually missing some required parameters here now that I use a current version I think. Will check tomorrow

jku: fixed now: https://github.com/jku/securesystemslib/actions/runs/3590694453/jobs/6044359440

lukpueh (approved):
Found one more non-blocking nit. Re-approving...

PR #443, opened 2022-10-26
Remove colorama optional dependency

lukpueh:
The coloroma library was used to optionally colorize the private key file path in a password prompt, for encryption or decryption.

Given that the feature is in a rarely used code path -- interactive calls are unlikely to be used outside of demo contexts -- and its merit it generally debatable, our broad goal to minimize dependencies weighs stronger.


jku (approved):
Ha! I was this :pinching_hand: close to making this same patch yesterday

joshuagl (approved):
Great decision, thank you.

PR #449, opened 2022-10-28
Fix GPG signature incompatibility

### LINKED ISSUE #448: GPGSigner (and gpg.functions.create_signature) produce invalid signature dicts
```python
        signer = GPGSigner(self.signing_subkey_keyid, self.gnupg_home)
        signature = signer.sign(self.test_data)
        print(signature.to_dict())
```
The result does not look like a TUF compliant signature (notice the key name "signature" instead of "sig"):
```
{
  'keyid': 'c5a0abe6ec19d0d65f85e2c39be9df5131d924e9',
  'signature': '5630c9b9436ed438cdf1386abe9ca3201612c24dee868e7294f6f4592fa05820a194cf1f87fd21f5ea0a596578a20491ca8829ef8d352df1139a953e28cf2176e3c3dc484e669782f65a1cc7ac4d1798ef3874e1f4a113aa59e5b6d350aacfa3c9c58a3ef2770995396971db3aae93d122a73795bb6b6b8d6e5b49a9be64bed8c13f849b09d43c7e916366a6290e92e41987636107703147b2fc86eec59909ddf1d002f39655c11176c18f04b927ecce48781258aa4e8f715014e2e30332839475528afa7faac0c48b425984eb726176092a2cbc3ece7d7e7eea372b8ffdf82715564536b961e5d195f1746b578ccaea19a5eb5936a2d38e0187af490ecc902a',
  'other_headers': '04000108001d162104c5a0abe6ec19d0d65f85e2c39be9df5131d924e90502635bbb2e'
}
```

Did no-one ever actually try to put a GPG signature in TUF metadata?

jku:
Fixes #448

This is an API break for gpg users. It changes the returned signature dict of `gpg.functions.create_signature()` (keyname "signature" changes to "sig") so that the GPG_SIGNATURE_SCHEMA is a superset of SIGNATURE_SCHEMA and so that the return value of `GPGSigner.sign() `is now a valid TUF signature.

It additionally removes GPGSignature as it seems to have no purpose.


jku:
**TODO**: Add a test in python-tuf that uses a GPG key to sign/verify metadata to verify this change

jku:
* I've added a compat fix so that new securesystemslib can verify GPG signatures made with old securesystemslib
* old securesystemslib will fail to verify new GPG signatures
* there is another spec incompatibility #450

I'm not 100% sure this change makes practical sense: it's going to break existing users and we already know there are other specification compatibility issues. At least the updates would need careful planning to avoid breaking intoto users.

jku:
had this failure on macos, py3.7 https://github.com/secure-systems-lab/securesystemslib/actions/runs/3362099110/jobs/5573366570: 
```
FAIL: test_gpg_available (__main__.TestGpgAvailable)
Test that GPG is available.
  File "/Users/runner/work/securesystemslib/securesystemslib/tests/check_gpg_available.py", line 39, in test_gpg_available
    self.assertTrue(securesystemslib.gpg.constants.have_gpg())
AssertionError: False is not true
```

I'll rerun to see if it's flaky as I think it is

jku:
Closing this based on https://github.com/secure-systems-lab/securesystemslib/issues/448#issuecomment-1301769776

Issue #447, opened 2022-10-28
RFE: private key reference scheme


jku:
Taking this out of #445 as I think it makes sense in general, not just in KMS:

### Idea

Let's define an identification scheme for private key materials in Signer: 
* allows applications that sign (TUF repository, or a CLI signing tool)  to reference a specific private key, regardless of private key "storage" method
* allows changing key "storage" (e.g. from RSA-key-in-file to GPG key) with configuration, no code changes needed
* simple so that it works on CLI and configuration files
* covers all private key types: private key files, GPG, yubikeys, KMS...

### sigstore ReferenceScheme

sigstore has a `ReferenceScheme` for their KMS abstraction: I think that would work for private keys in general
https://github.com/sigstore/sigstore/tree/main/pkg/signature/kms. There's no requirement to be 1:1 compatible with sigstore, just showing this as a reference of something we could copy.

### Proposal

https://docs.google.com/document/d/1KQY5v1ASXpZM5GhfATV_WJf7GaMJrfBOXr06dtxc-ps

Review welcome here or in the doc!

jku:
Almost complete proposal https://docs.google.com/document/d/1KQY5v1ASXpZM5GhfATV_WJf7GaMJrfBOXr06dtxc-ps -- still debating myself if I should propose a Key class as well: this is definitely ready for review

lukpueh:
> still debating myself if I should propose a Key class as well 

FYI: There's new `Key`-related work available downstream in [in-toto/securesystemslib](https://github.com/in-toto/securesystemslib/blob/master/securesystemslib/key.py). This is part of https://github.com/secure-systems-lab/securesystemslib/issues/370

joshuagl:
@kairoaraujo would love your input on the design here, especially:
* URI scheme key reference
* Importing existing keys (i.e. SSH key or YubiKey) is not yet handled, any thoughts on that?

jku:
> * Importing existing keys (i.e. SSH key or YubiKey) is not yet handled, any thoughts on that?

I swear I wrote about it somewhere, but can't hurt to do it here too:
* Constructing entirely new keys or importing them is inherently specific to signing system -- we can't create generic APIs and should not try
* I would add this functionality into Signer implementations: 
   * typical use case is going to need both signer and Key at same time (to insert them into their respective storage)
   * since I've made the signer already "own" a public key in the PR, we can add generic API to get that public key from Signer
[...]

PR #456, opened 2022-11-04
signer: Add abstract Key class, implement private key uri scheme for Signer

jku:
Fixes: #447

###  Summary
* move Key from python-tuf to here
* Make Key abstract so we can have multiple implementations
* Add a private key URI scheme so that Signer is actually usable without knowing the specific implementation
* Split the signer API to internal submodules (this is a large code move but not a change to public API)

The private key URI scheme allows application code to handle private keys in a generic way: the keytype, used signer implementation or private key storage system are not something the application needs to worry about. Example 
```python
# new signer usage
signer = Signer.from_priv_key_uri("envvar:PRIV_KEY", pubkey)
sig = signer.sign(b"DATA")

# Key usage is unchanged (apart from verify_signature() now being agnostic about the payload)
key = Key.from_dict(keyid, dict)
```

### Changelist

* The user facing API additions are:
  * `Signer.from_priv_key_uri()` added
  * private key URI schemes are now part of the public API
  * `SIGNER_FOR_URI_SCHEME` and `KEY_FOR_TYPE_AND_SCHEME` define supported signers and keys
  * Signer implementations are now 95% implementation details of Signer: only the (currently unimplemented) new key generation/import is "public"
  * Key no longer has verify_signature() for Metadata arguments because that is python-tuf specific: instead it works with bytes. This should not be an issue but requires a change in python-tuf Metadata-verify_delegate()
  * Key also does not have from_securesystemslib_key(): it is now in SSlibKey

* Other changes that do not really affect users
  * Key moved from python-tuf to securesystemslib
  * Key and Signer are now both abstract classes
  * signer API split to multiple files -- this is not visible to users

### Other notes:
* In normal use, Signer and Key subclasses are never needed. Instead they both have a dynamic dispatch to find the correct implementation:
  * Signer dispatch from `Signer.from_priv_key_uri()`
  * Key dispatch from `Key.from_dict()`
* Signers using priv key uris now "own" a public Key, or in other words: A Key can exist without a Signer but a Signer cannot exist without a Key. How this affects usage:
  * Keys can just be deserialized as before
  * Signers can be loaded with a private key URI and a public Key 
  * Signers can be generated from scratch or imported using Signer implementation specific API (unimplemented in this PR).  The public key can then be requested from the Signer
* SSlibKey and SSlibSigner implementations are currently the only ones, but now others can be added
* Exceptions in the whole signer API could use a review, but for this PR please look at the Key.verify_signature() exceptions: they are the important ones

### TODO later: I will file issues
* GPG support is left out of the private key uri scheme. plan is:
  * intoto can take GPGSigner and GPGKey. GPGSigner can be removed from sslib
  * intoto can hook GPGSigner and GPGKey into the mechanisms added in this PR
  * intoto takes care of the special keyid matching required by GPGKey
  * later we can add a more spec compliant GPG signer and key into sslib 
* Key.get_pre_hash_algorithm() is not implemented: this would be easy to implement and useful for KMS and HSM at least
* Signer generation and import are not yet implemented: this should be an optional feature in Signer implementations (not part of generic Signer API).
* Signer class currently does expose public key: it probably should as the A) the priv key uri scheme pretty much requires it already and B) key import would become nicer with it
* users are likely to only read Signer docs (as that's what they are using). This leads to two issues
  *  implementation specific exceptions -- how do we handle e.g. GCPSigner raising on network error?
  * documenting the URIs: once we have multiple different Signer subclasses this documentation should be collected somewhere for the supported signers...


securesystemslib/signer.py
```diff
@@ -154,42 +311,136 @@ def sign(self, payload: bytes) -> Signature:
...
-    - rsa(ssa-pss|pkcs1v15)-(md5|sha1|sha224|sha256|sha384|sha512) (12 schemes)
-    - ed25519
-    - ecdsa-sha2-nistp256
-
-    See "securesystemslib.interface" for functions to generate and load keys.
+    securesystemslib-style rsa, ed25519 or ecdsa key. See keys module
+    for the supported types, schemes and hash algorithms.
+
+    SSlibSigners should be instantiated with Signer.from_priv_key_uri().
+    Two private key URI schemes are supported:
```
jku, line 365:
```suggestion
    These private key URI schemes are supported:
```

securesystemslib/signer.py
```diff
@@ -154,42 +311,136 @@ def sign(self, payload: bytes) -> Signature:
...
+        VAR is an environment variable that contains the private key content.
+           envvar:MYPRIVKEY
+    * file:<PATH>:
+        PATH is a file path to a file that contains private key content.
+           file:path/to/file
+    * encfile:<PATH>:
+        The the private key content in PATH has been encrypted with
+        keys.encryot_key(). Application provided SecretsHandler will be
+        called to get the passphrase.
+           file:/path/to/encrypted/file
```
jku, line 376:
```suggestion
           encfile:/path/to/encrypted/file
```

securesystemslib/signer.py
```diff
@@ -137,8 +148,154 @@ def to_dict(self) -> Dict:
...
+            raise ValueError("keydict not in securesystemslib format") from e
+
+        return cls(
+            key_dict["keyid"],
+            key_meta["keytype"],
+            key_meta["scheme"],
+            key_meta["keyval"],
+        )
+
+    def is_verified(self, signature: Signature, data: bytes) -> bool:
```
jku, line 262: This requires a bit of thinking. What I did here is:
* leave `verify_signature()` in python-tuf (as it's metadata related)
* add new `is_verified()` to wrap `keys.verify_signature()`

The former raises on verification failure, latter returns False.

The other alternative would be to implement a `verify_signature()` here as well:a method that would also raise on failure, but would take bytes as input. This would require the overload in python-tuf to do some tricks: either implement the "overload dispatch" but that is not too onerous I think.

I think I like the  `verify_signature()` overload idea more than the PR here

jku: With "overload dispatch" I mean a setup like this for verify_signature:
```python
class Base:
    def func(self, i: int) -> None:
        # implement base func here

class Derived(Base):
    @overload
    def func(self, arg: int) -> None:
        ...

    @overload
    def func(self, arg: str, extra_arg: Optional[str] = None) -> None:
        ...

    def func(self, arg: Union[int, str], extra_arg: Optional[str] = None) -> None:
        if isinstance(arg, int):
            return super().func(arg)

        # implement derived func here


```

lukpueh: > The other alternative would be to implement a verify_signature() here as well:a method that would also raise on failure like the namesake in python-tuf, but would take bytes as input. This would require the overload in python-tuf to do some tricks: likely implement the "overload dispatch" (or just silence mypy) but that seems reasonable..

Can you explain why this requires overloading? In your example, what is Base, Derived, func?

securesystemslib/signer.py
```diff
@@ -137,8 +148,154 @@ def to_dict(self) -> Dict:
         }
 
 
+class Key:
```
lukpueh, line 151: In `in-toto/securesystemslib` we made [`Key`](https://github.com/secure-systems-lab/securesystemslib/compare/master...in-toto:securesystemslib:master#diff-ae2ca657fc834f3033da3502b45f73a8380f6a6cc2fde77fafb3715fa5237c65R12) an abstract interface to support multiple container formats and verification  implementations.

We made your traditional securesystemslib `Key` a subclass called [`SSlibKey`](https://github.com/secure-systems-lab/securesystemslib/compare/master...in-toto:securesystemslib:master#diff-ae2ca657fc834f3033da3502b45f73a8380f6a6cc2fde77fafb3715fa5237c65R46) and added another subclass [`GPGKey`](https://github.com/secure-systems-lab/securesystemslib/compare/master...in-toto:securesystemslib:master#diff-ae2ca657fc834f3033da3502b45f73a8380f6a6cc2fde77fafb3715fa5237c65R168) to allow independent attributes, and `verify` and `match_keyid` implementations.

The downside is that this makes deserialization harder. You'd have to call either `SSlibKey.from_dict` or `GPGKey.from_dict` based on the format of the passed dict, e.g. in python-tuf's `api.metadata.Root.from_dict`. But I think that's feasible. What do you think?

securesystemslib/signer.py
lukpueh: Btw. there's a related issue about the return type of `Key.{is_verified aka. verify aka. verify_signature}` in https://github.com/in-toto/securesystemslib/issues/8

securesystemslib/signer.py
jku: all of those links point to a 1400 LOC diff

lukpueh: Hm. They should have pointed to a line in that diff. Let me try again.

lukpueh: Updated the links. Now it should work.

jku: but I found it: my first questions is, is this just wrapping paper to hide the  fact that GPGSignature and GPGKey are incompatible with the tuf and intoto specs? In that case I'm not fond of the idea of breaking an abstraction that works fine otherwise... 

If you do want to support non-compliant keys, can't the generic Key implementation just hide that ugly detail without forcing clients to change their code?

jku: actually, I would not want Key to suddenly use non-compliant metadata in python-tuf... so I guess Key hiding that is out of the question (at least without a "i-want-non-compliant-metadata" flag somewhere).

Further thinking:
* GPGKey (as designed) existing in securesystemslib seems fine if intoto uses it already _but it should not be a Key_ otherwise the meaning of Key is lost
* If you really want a superclass I guess it would be this:
```
class IntotoKey # abstract
class Key(IntotoKey)
class PGPKey(IntotoKey)
```
these are essentially same classes as yours: it's just that I'm calling your SSLibKey Key -- that's the real key class, everything else is wrapping paper for intoto

jku:
Results from chat with Lukas:
* I will try to modify Key class in this PR to support multiple key implementations
* goal is to be compatible with the in-toto/securesystemslib forks concept of Key abstraction
* ... with the difference that everything should be abstracted: verify_signature, serialization, keyid matching, payload hash algorithm lookup ( the only exception is construction from scratch which is quite implementation-specific)

This is quite a bit more than just moving the Key class... but it fits in well with Signer change as they are both dispatcher-style abstractions (for Key and Signer respectively)

jku:
This ballooned "a bit"... I've  updated the description, hoping it makes sense to others.

jku:
Oh and I forgot to mention: Most of the GPGKey implementation in here comes from https://github.com/in-toto/securesystemslib/ so is likely @PradyumnaKrishna's work: if this PR looks like a path forward, should make sure he gets credited (I can try editing the commit history so authorship is visible in git)

securesystemslib/signer.py
```diff
@@ -137,8 +150,264 @@ def to_dict(self) -> Dict:
...
+    """
+
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(
+        self,
+        keyid: str,
+        keytype: str,
+        scheme: str,
+        keyval: Dict[str, Any],
```
lukpueh, line 181: `Key` really only needs `keytype` and `scheme` to dispatch. We could define all other fields in sublcasses for more flexibility.  `SSlibKey` can still be the default to ensure spec compliance.

securesystemslib/signer.py
```diff
@@ -137,8 +150,264 @@ def to_dict(self) -> Dict:
...
+        """Creates ``Key`` object from TUF serialization dict.
+
+        Key implementations must override this factory constructor.
+
+        Raises:
+            KeyError, TypeError: Invalid arguments.
+        """
+        keytype = key_dict.get("keytype")
+        scheme = key_dict.get("scheme")
+        if (keytype, scheme) not in KEY_FOR_TYPE_AND_SCHEME:
```
lukpueh, line 221: At least for the current set of schemes, the scheme alone is enough to dispatch. Using scheme alone here would make it easier to deprecate the redundant keytype field. 

Or do you think its better to also plan for cases where scheme alone is not enough?

securesystemslib/signer.py
```diff
@@ -137,8 +150,264 @@ def to_dict(self) -> Dict:
...
+            exceptions.UnsupportedAlgorithmError,
+        ) as e:
+            logger.info("Key %s failed to verify sig: %s", self.keyid, str(e))
+            raise exceptions.VerificationError(
+                f"Unknown failure to verify signature by {self.keyid}"
+            ) from e
+
+
+@dataclass
+class GPGKey(Key):  # pylint: disable=too-many-instance-attributes
```
lukpueh, line 308: I feels odd that `GPGKey` is a `Key` although it has different fields and cannot be instantiated with `Key.from_dict`. IMO it makes only sense to implement this in `securesystemslib` if it is relevant for `python-tuf` and DSSE as well, otherwise we should just implement it in `in-toto`.

securesystemslib/signer.py
jku: I don't understand what we gain with that -- sure, Key itself doesn't use them but every Key implementattion that is compatible with intoto or TUF spec does need them.  If Key is an implementation of TUF and in-toto specification keys, why would we remove fields defined in those specs? It just seems like every spec-compliant key has to re-implement them anyway.

securesystemslib/signer.py
jku: I'm just trying to implement what I think the specs want -- based on them, keys seem to be identified by type and scheme. It's true that the keys we know of seem to be uniquely identified by scheme only

securesystemslib/signer.py
jku: Key.from_dict does work for GPGKey -- as long as GPGKey is in the supported keys list.

I don't think the current GPGKey will ever be relevant to python-tuf: both keys and signatures break the spec and we can't change that because in-toto depends on this behaviour. I can't imagine the spec changing to allow current GPGPKey behaviour but I suppose that could happen.

But I don't think there is a problem with
* keeping GPGKey in securesystemslib as long as intoto uses it
* if there is demand, adding a compliant gpg key version (using a different Key and Signer subclass)

jku:
I think we can do both:
* Add a compliant BetterGPGKey and BetterGPGSigner classes
* keep the current gpgp implementation until intoto does not need it

As mentionedin the other reply, GPGKey does support `Key.from_dict()`.

securesystemslib/signer.py
jku: but this is definitely not an important issue: let's dispatch on scheme if you want

jku:
I think the special keyid matching is not too onerous for the `Key.verify_signature()` callers using GPGKey, so we can leave it to them: In practice in-toto just needs to embed the matching function in their sources and remember to use that

securesystemslib/signer.py
```diff
@@ -137,8 +150,264 @@ def to_dict(self) -> Dict:
...
+                f"Failed to verify signature by {self.keyid}"
+            )
+
+
+# SecretsHandler is a function the calling code can provide to Signer:
+# If Signer needs secrets from user, the function will be called
+SecretsHandler = Callable[[str], str]
+
+
 class Signer:
```
jku, line 398: Signer class definition should probably include public_key as an attribute now: 
* the rest of the API (like from_priv_key_uri()) kind of assumes Signer always has a Key
* implementing e.g. Signer creation/import becomes more straight forward if the key is somehow available Example:
    ```
    # Load an existing KMS key, import the public key from it
    signer = GCPSigner.import(gcp_key_id)
    pubkey = signer.public_key
    ```

securesystemslib/signer.py
lukpueh: > I don't understand what we gain with that

We would gain a base class that can be used in a meaningful way for non-spec compliant keys. As `GPGKey` shows this is not an unlikely use case.

I understand that you want to avoid to accidentally load a non-spec compliant key with `Key.from_dict`, but I think that this can also be achieved by making `SSlibKey` the strictly spec-conformant key, and registering only that in the default `KEY_FOR_TYPE_AND_SCHEME`. Moreover, python-tuf could require `SSlibKey` instances.

lukpueh: Btw. using `scheme` alone to dispatch, as I proposed below, would allow us even remove the `keytype` field from `Key`.

lukpueh: I won't insist on either of them. I just think that TUF and in-toto public key metadata formats are not ideal and that we can design the `Key` interface to make it easier for implementations that use a better key format (at the cost of being non-spec compliant).

securesystemslib/signer.py
lukpueh: > Key.from_dict does work for GPGKey -- as long as GPGKey is in the supported keys list, see test_key_from_to_dict().

Oh right. But do we really want to tell users to add `(None, None)` to the dispatch table?  IMO it's okay for a user of the `Key` interface to do that as a workaround for a custom `Key` subclass. But we should not require it for a `Key` subclass that we provide. It feels like an abuse of our own API.

lukpueh:
> I think we can do both:
> 
> * Add a compliant BetterGPGKey and BetterGPGSigner classes

👍 Let's do that in a follow-up PR.


> * keep the current gpgp implementation until intoto does not need it

I suggest to remove it from securesystemslib, for these reasons (as discussed above):

- `GPGKey` is not really compatible with `Key` as it has different fields and dispatching to it with `Key.from_dict` requires a work around. This should not be public securesystemslib API.
 
- `GPGSigner` was originally added to  `securesystemslib` to allow gpg signing with python-tuf's new metadata API ([#341](https://github.com/secure-systems-lab/securesystemslib/pull/341)), and recent efforts with `GPGKey`  are related to the securesystemslib DSSE implementation ([in-toto/securesystemslib](https://github.com/in-toto/securesystemslib)). However, neither python-tuf nor DSSE should use the current implementations of `GPG{Signer, Key, Signature}`, but rather a `BetterGPG{Signer, Key}`, which is spec compliant wrt metadata format. And securesystemslib should not host features that are only relevant to one of its users.

securesystemslib/signer.py
jku: Yes, GPG support as it is, is certainly a hack. 

Moving the (Keys+Signer) GPG support to in-toto sounds good to me. We just need to make sure securesystemslib allows whatever is needed to do it:
* out-of-tree Key and Signer
* Key dispatch that somehow allows dispatching to GPGKey even though it's a bit different
* keyid matching has to be made to work with GPG (this is not an issue wright now but may be with DSSE) 

I think this all looks solved right now, at least well enough

jku: wait, I maybe misunderstood your comment, you actually mention a workaround within Key... I'm not enthusiastic on paper but we can discuss this when we next talk

lukpueh:
> I think the special keyid matching is not too much to ask from the Key.verify_signature() callers using GPGKey, so we can leave it to them: In practice in-toto just needs to embed the matching function in their sources and remember to use that

Yes, this should be easy if it's only needed in in-toto. The reason why we defined the `match_keyid` interface here was, that we needed it in in-toto's metadata API and also for DSSE, which was going to be implemented in securesystemslib. But if we decide that DSSE should only use a ` BetterGPG{Signer, Key}` that does not have special keyid matching than we don't need to worry about this in securesystemslib.

securesystemslib/signer.py
lukpueh: I meant that it's okay for in-toto to implement `GPGKey` without `keytype` and `scheme` and register it under `securesystemslib.signer.KEY_FOR_TYPE_AND_SCHEME[(None, None)]` as a workaround.  But it seems odd if securesystemslib implements `GPGKey` and requires in-toto to perform that workaround.

I think we are on the same page.

jku:
Current state:
* GPGKey has been removed from the PR in the last commit, GPGSigner no longer claims to be compatible with the signer URI scheme 
* PR is ready for general "is this a good idea" sort of review  but feel free to leave line-by-line level review for until polishing happens
* the idea is that intoto can implement both GPGKey and GPGSigner in their code base and plug them in via the dispatch system (in a hacky way but still). I did not remove the original commits from this branch so the GPGKey implementation I tested with (which again was mostly copied from the intoto fork) is available if needed.

Potential review questions/improvements:
* split code to multiple files? How should users import this API? How do we make it clear these modules are  separate from the old modules?
* key import is not implemented: I think it should be Signer implementation specific functionality (but I don't see a problem if a specific Key implementation wants to add a `SomeKey.import(...)`)
* Are the privkey URIs clear enough? I think being a URI allows required complexity (and library support)... I did not try to stay "compatible" with the sigstore scheme: I don't think that is valuable and if nothing enforces it, the compatibility will deteriorate anyway over time.
* do we want to model base Key class on current TUF & intoto specs or just on what securesystemslib actually needs? I've tried to do former but this leads to Key having both a keytype and scheme which, as  lukas has mentioned, may be unnecessary for the implementation... It just felt like being spec compatible by default would be good.

lukpueh:
> * split code to multiple files? How should users import this API? How do we make it clear these modules are separate from the old modules?

In  'in-toto/secureystemslib' we split into `signer.py` for `*Signer` and `*Signature` and `key.py` for `*Key`, but I don't know if that's any better than keeping everything in one file, or a different separation.

It looks like most users will import `Signer` or `Key` and not their subclass implementations:
```python
from secureystemslib.signer import Signer
from secureystemslib.key import Key
```
This is not very nice but also not the end of the world. It might also mean some circular dependencies (at least for annotations).  And it does not make it clear that these classes belong together. So maybe they should be all in one module or in one subpackage. We've been referring to this new API as "the signer API" informally, so that name probably makes sense.

Either way, I think we should make it clear that these modules are separate from old modules above all through documentation. E.g. we could host only these new classes on readthedocs.




> * key import is not implemented: I think it should be Signer implementation specific functionality (but I don't see a problem if a specific Key implementation wants to add a `SomeKey.import(...)`)

Let's ticketize and see if there are uses cases for importing a public key from something other than metadata in a non-signing context. If yes, we should support `*Key.import`, if no pubkey import can be implemented as part of private key import.

> * Are the privkey URIs clear enough? I think being a URI allows required complexity (and library support)... I did not try to stay "compatible" with the sigstore scheme: I don't think that is valuable and if nothing enforces it, the compatibility will deteriorate anyway over time.

Should scheme name indicate the `SSlibSigner` scope or otherwise encode information about the private key values?

> * do we want to model base Key class on current TUF & intoto specs or just on what securesystemslib actually needs? I've tried to do former but this leads to Key having both a keytype and scheme which, as  lukas has mentioned, may be unnecessary for the implementation... It just felt like being spec compatible by default would be good.

See https://github.com/secure-systems-lab/securesystemslib/pull/456#discussion_r1020172612

> * exceptions should maybe get more attention -- as an example I have no idea what `from_priv_key_uri()` can currently throw. This is not super critical as this happens during signing, based on user input: it's not nearly as fatal as failing on verify based on deserialized remote data

Agreed. This is probably something we should figure out before we make a stable release.

jku:
> > do we want to model base Key class on current TUF & intoto specs or just on what securesystemslib actually needs? I've tried to do former but this leads to Key having both a keytype and scheme which, as lukas has mentioned, may be unnecessary for the implementation... It just felt like being spec compatible by default would be good.
> See #456 (comment)

I just still don't understand _why_ we would encourage new Key implementations to be incompatible with both intoto and python-tuf -- both of those projects may want to require that keyid and keyval exist...  I suppose we should then add deserialization time checks to python-tuf to make sure the key contains the required fields if securesystemslib does not promise that?

Since you have a strong opinion on this and I don't, I think I'll modify the Key class and remove keyid and keyval from the class.

securesystemslib/signer.py
jku: > I understand that you want to avoid to accidentally load a non-spec compliant key with Key.from_dict, but I think that this can also be achieved by making SSlibKey the strictly spec-conformant key, and registering only that in the default KEY_FOR_TYPE_AND_SCHEME.

Oh! I don't think this is something I want: I'd certainly want to add things like "SigstoreKey" and "CompatibleGPGKey" in the future (not necessarily in securesystemslib but somewhere), and I would not want to put them in SSlibKey

jku:
> Should scheme name indicate the SSlibSigner scope or otherwise encode information about the private key values?

The way I see private key URIs:
* scheme is a unique identifier that defines how to parse the rest of the URI
* scheme should not itself refer to the Signer implementation: the mapping of scheme to signer is a choice made by user (although signer module does present reasonable defaults)
* a Signer implementation can handle mutliple schemes if that makes the URIs easier to understand than stuffing many arguments into the rest of the URI -- this is the reason for multiple URIs in SSlibSigner
* the rest of the URI is an implementation detail of the Signer implementation, but it should be a valid URI so that generic parsing libraries like urllib are likely to work

lukpueh:
> > > do we want to model base Key class on current TUF & intoto specs or just on what securesystemslib actually needs? I've tried to do former but this leads to Key having both a keytype and scheme which, as lukas has mentioned, may be unnecessary for the implementation... It just felt like being spec compatible by default would be good.
> > 
> > 
> > See #456 (comment)
> 
> I just still don't understand _why_ we would encourage new Key implementations to be incompatible with both intoto and python-tuf -- both of those projects may want to require that keyid and keyval exist... I suppose we should then add deserialization time checks to python-tuf to make sure the key contains the required fields if securesystemslib does not promise that?
> 
> Since you have a strong opinion on this and I don't, I think I'll modify the Key class and remove keyid and keyval from the class.

Summarizing a chat with Jussi:
- The `Key` class as currently implemented mostly makes sense regardless of specification compatibility, this includes the fields strings `keyid` and `scheme` and the dict `keyval`
- Exceptions are `keytype` (redundant with `scheme`) and `keyval` being a dict (could be e.g. a string)
- These issues are not big enough to make `Key` look different than the keys in the specification
- Users can register any Key subclasses they want 
- Using the existing fields gives enough flexibility to use any desired technology
- Using different fields is possible too (Python allows this) but would look hackish, which is a fair ask for using non-spec compliant keys

TL;DR: `Key` can stay as is

securesystemslib/signer.py
jku: Just to document this for others: I am removing GPGKey from the PR. The idea is that the current GPG implementation likely moves to intoto.

securesystemslib/signer.py
jku: I am not exposing Signer.public_key or making it a required construction arg for SSllibSigner in this PR. 

I think both probably make sense (for e.g. the key import use case) but the PR is large already and does not itself require those changes: will file further issues.

jku:
Ok! I've added three more commits and updated the description: please read that to see what is in here. This is ready for review: I can't improve it alone.

But I will leave a self-review: the commit structure is a mess: the code move (to mutliple internal modules) is IMO reasonable but it happens mid-way through the PR... That means review might only make sense for the end result which is +774 LOC. Most of that is just moves of existing Key and Signer code but there are many changes on top of that. If this feels unacceptable I can try doing this in bits and pieces to make it easier to review.

securesystemslib/signer/_signer.py
```diff
@@ -111,17 +115,13 @@ class SSlibSigner(Signer):
 
     SSlibSigners should be instantiated with Signer.from_priv_key_uri().
     These private key URI schemes are supported:
-    * envvar:<VAR>:
+    * "envvar:<VAR>":
         VAR is an environment variable with unencrypted private key content.
            envvar:MYPRIVKEY
-    * file:<PATH>:
+    * "file:<PATH>?encrypted=[true|false]":
         PATH is a file path to a file with unencrypted private key content.
```
jku, line 104: this is now missing the docstring for the encrypted version

securesystemslib/signer/_signer.py
```diff
@@ -0,0 +1,264 @@
...
+
+            # read private key (may be encrypted or not) from file
+            with open(uri.path, "rb") as f:
+                private = f.read().decode()
+
+            if params["encrypted"] != "false":
+                if not secrets_handler:
+                    raise ValueError("encrypted key requires a secrets handler")
+
+                secret = secrets_handler("passphrase")
```
lukpueh, line 165: Can you explain how `secrets_handler` is supposed to be used? I don't understand why it can be passed a name for the secret, and why this is hardcoded `"passphrase"` here.

securesystemslib/signer/_signer.py
```diff
@@ -0,0 +1,264 @@
...
+    Provides a sign method to generate a cryptographic signature with a
+    securesystemslib-style rsa, ed25519 or ecdsa key. See keys module
+    for the supported types, schemes and hash algorithms.
+
+    SSlibSigners should be instantiated with Signer.from_priv_key_uri().
+    These private key URI schemes are supported:
+    * "envvar:<VAR>":
+        VAR is an environment variable with unencrypted private key content.
+           envvar:MYPRIVKEY
+    * "file:<PATH>?encrypted=[true|false]":
```
lukpueh, line 103: Looking at the code, `true` can be anything but `false`. Is this true-ish behavior intended? I guess it makes sense.

lukpueh (commented):
Here's a first batch of comments. Mostly minor nits and questions. Can't wait to get this merged. 🎉 

Still want to take a look at `_key.py` and `test_signer.py`

securesystemslib/signer/_signer.py
```diff
@@ -0,0 +1,264 @@
...
+class GPGSigner(Signer):
+    """A securesystemslib gpg implementation of the "Signer" interface.
+
+    Provides a sign method to generate a cryptographic signature with gpg, using
+    an RSA, DSA or EdDSA private key identified by the keyid on the instance.
+
+    GPGSigners should be instantiated with Signer.from_priv_key_uri().
+    Two private key URI schemes are supported:
+    * gpg:<HOMEDIR>:
+        HOMEDIR: Optional filesystem path to GPG home directory
```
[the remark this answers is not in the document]
jku: oops: it is not up-to-date

securesystemslib/signer/_signer.py
```diff
@@ -0,0 +1,264 @@
...
+        self.homedir = homedir
+
+    @classmethod
+    def from_priv_key_uri(
+        cls,
+        priv_key_uri: str,
+        public_key: Key,
+        secrets_handler: Optional[SecretsHandler] = None,
+    ) -> "GPGSigner":
+        raise NotImplementedError("Incompatible with private key URIs")
```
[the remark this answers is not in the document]
jku: I think if I remove it then mypy will complain as this is a abstractmethod in parent.

securesystemslib/signer/_signer.py
jku: * signer calls the method when it needs a secret from user
* signer providers the name of the secret (just in case there could be several secrets -- this is a bit theoretical)
* the application implements SecretsHandler: here's a simple one for a CLI app:
```python
from getpass import getpass

def my_secret_handler(secret_name:str ) -> str:
    return getpass(f"Please provide {secret_name}: ")
```

securesystemslib/signer/_signer.py
jku: I figured it's the safest option to avoid somehow ending up with unencrypted passphrases if
* we later implement key creation and 
* you use "encrypted=yes" or "encrypted=True" or something

securesystemslib/signer/_signer.py
jku: oh, and if it's not clear: secrets_handler is a method (and not a string argument to from_priv_key_uri()) because
* we don't know if or when the secret is needed: there could be cases where it's requested at signing time and sometimes it might not be requested at all (think yubikey with a short "unlock window" after first PIN input)
* we don't necessarily know which secret is required (again, this is a bit theoretical but possible)

lukpueh: Ah, now I understand, thanks for explaining! I was confused by the comment above the secrets handler definition: "The argument is the name of the secret", when really it can be anything you implement in the secrets handler. That plus the hardcoded "passphrase", which might as well be hardcoded into the secrets handler. But I guess it makes sense for some use cases. There might be cases where no arg or multiple or non-string args might also make sense.

securesystemslib/signer/_signer.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def __init__(self, key_dict: Dict):
+        self.key_dict = key_dict
+
+    @classmethod
+    def from_priv_key_uri(
+        cls,
+        priv_key_uri: str,
+        public_key: Key,
+        secrets_handler: Optional[SecretsHandler] = None,
```
MVrachev, line 125: There is no documentation for the `secrets_handler` argument.

MVrachev (commented):
I had a birds-eye view of the pr as it's a bigger one.
Still, it was a really nice experience to review given that I just had my first look at the issue and the propose changes. Great job in documenting the discussion!

About the pr: I like that `Key` is an abstract class that can be easily reused in `in-toto` and `python-tuf`. We have discussed that this should be done a while ago.
Also, I really like that another important problem is solved by the pr - the support for private keys.

I don't have any great suggestions only a small comment.

securesystemslib/signer/_signer.py
jku: Good catch. I'm torn about documenting these methods at all. The abstractmethod is already documented in the base Signer: copy-pasting to SSlibSigner seems silly... but on the other hand there would be the additional detail like specific errors raised by sslibsigner.... and on the third hand no-one is going to read these specific Signer implementations docs since their code will be using the Signer class -- they won't even know which implementation will be dispatched.

The underlying issue is that because  SSlibSigner is just an implementation of Signer _it should not need any new method docs_. But for that to be 100% true, I would have to add a "SignerError" that catches all errors from the dispatched implementation (otherwise OSErrors or google.cloud errors are bound to happen)... and that just seems wrong.

I think this might need some more thought. A good example situation is a GCPSigner failing because there is no network connection... This could be a case that a repository application wants to handle. But I also think we could accept the situation for now and try to come up with a solution before this is released

securesystemslib/signer/_signer.py
```diff
@@ -0,0 +1,264 @@
...
+        Raises:
+            OSError: Reading the file failed with "file:" URI
+            ValueError: URI is unsupported or environment variable was not set
+                with "envvar:" URIs
+
+        Returns:
+            SSlibSigner for the given private key URI.
+        """
+        if not isinstance(public_key, SSlibKey):
+            raise ValueError(f"Expected SSlibKey for {priv_key_uri}")
```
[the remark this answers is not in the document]
jku: I decided not to:
* Signer.from_priv_key_uri() already states that it  rasies `ValueError: Incorrect arguments` and that the given public key must be for this specific private key
* I think it's reasonable to expect that users will only read Signer docs (since that's the class they use) and not the implementation docs

There is a related issue here: I left a longer comment in reply to martin about exceptions from signer implementations...

securesystemslib/signer/_signer.py
lukpueh: Okay, but if the passed name of the secret is not really a canonical identifier and not documented. How should the application, which provides the handler know what to expect and how to deal with it?

securesystemslib/signer/_signer.py
lukpueh: Yes, there's an unresolvable conflict: 
- users should only know the base class
- users should know traits that are specific to subclasses (exceptions, name of a secret)

I think there are only two ways to overcome his:
- discard one of the assumptions (probably the former)
- make all subclass traits available via the base class (that does not seem desirable or feasible)

jku:
Resolved most comments with and addition commit
Added this to the TODO list:

> * users are likely to only read Signer docs (as that's what they are using). This leads to two issues
>   *  implementation specific exceptions -- how do we handle e.g. GCPSigner raising on network error?
>   * documenting the URIs: once we have multiple different Signer subclasses this documentation should be collected somewhere for the supported signers...

The latter can definitely be solved later. The former is a bit trickier... we should have some decision before we release this as public API

securesystemslib/signer/_signer.py
jku: Ok, I've mostly removed the docs for SSlibSigner: I think copy-pasting is not useful considering the users will be reading Signer docs anyway...

But I have added a TODO item and a new comment about the exception issue that this revealed.

resolving this conversation

securesystemslib/signer/_key.py
```diff
@@ -0,0 +1,199 @@
...
+
+    @classmethod
+    @abc.abstractmethod
+    def from_dict(cls, keyid: str, key_dict: Dict[str, Any]) -> "Key":
+        """Creates ``Key`` object from a serialization dict
+
+        Key implementations must override this factory constructor that is used
+        as a deserialization helper.
+
+        Users should call Key.from_dict(): it dispacthes to the actual subclass
```
lukpueh, line 82:
```suggestion
        Users should call Key.from_dict(): it dispatches to the actual subclass
```

securesystemslib/signer/_key.py
```diff
@@ -0,0 +1,199 @@
...
+        # NOTE: Explicitly not checking the keytype and scheme types to allow
+        # intoto to use (None,None) to lookup GPGKey, see issue #450
+        key_impl = KEY_FOR_TYPE_AND_SCHEME[(keytype, scheme)]  # type: ignore
+        return key_impl.from_dict(keyid, key_dict)
+
+    def to_dict(self) -> Dict[str, Any]:
+        """Returns a serialization dict.
+
+        Key implementations may override this method. This method is a
+        deserialization helper.
```
lukpueh, line 102:
```suggestion
        serialization helper.
```

securesystemslib/signer/_key.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        return {
+            "keytype": self.keytype,
+            "scheme": self.scheme,
+            "keyval": self.keyval,
+            **self.unrecognized_fields,
+        }
```
lukpueh, line 109: I think it would be safer to raise NotImplementedError or even just not implement this here and require subclasses to implement their own `to_dict`. I am worried that a default `to_dict` could lead to an undetected wrong serialization of a subclass.

securesystemslib/signer/_key.py
```diff
@@ -0,0 +1,199 @@
...
+
+    @classmethod
+    def from_securesystemslib_key(cls, key_dict: Dict[str, Any]) -> "SSlibKey":
+        """Constructor from classic securesystemslib keydict"""
+        return SSlibKey(
+            key_dict["keyid"],
+            key_dict["keytype"],
+            key_dict["scheme"],
+            key_dict["keyval"],
```
lukpueh, line 147: The python-tuf variant of this method uses `format_keyval_to_metadata`, which also removes the private portion of the key if any. Should we do this here too?

lukpueh (commented):
Here's the rest of my review. This is great! Also your recent doc updates seem very helpful. I left a few tiny comments on public keys inline...

securesystemslib/signer/_key.py
jku: oh yeah... I meant to just build keyval manually but apparently forgot

securesystemslib/signer/_key.py
jku: sure, wfm

jku:
Resolved mentioned issues as suggested. Also added a minimal type check to SSlibKey.from_dict (to verify there is a keyval.public and it is a string).

@lukpueh does this look familiar (just happened on CI, never seen before myself): 
```
FAIL: test_rsa (tests.test_interface.TestInterfaceFunctions)
Test RSA key _generation and import interface functions.
  File "/home/runner/work/securesystemslib/securesystemslib/tests/test_interface.py", line 315, in test_rsa
    self.assertTrue(
AssertionError: False is not true : expected: 'Bad decrypt. Incorrect password?' got: 'RSA (public, private) tuple cannot be generated from the encrypted PEM string: ('Could not deserialize key data. The data may be in an incorrect format, it may be encrypted with an unsupported algorithm, or it may be an unsupported key type (e.g. EC curves with explicit parameters).', [_OpenSSLErrorWithText(code=503841036, lib=60, reason=524556, reason_text=b'error:1E08010C:DECODER routines::unsupported'), _OpenSSLErrorWithText(code=1090520[27](https://github.com/secure-systems-lab/securesystemslib/actions/runs/3546570396/jobs/5955750691#step:5:28), lib=13, reason=123, reason_text=b'error:0680007B:asn1 encoding routines::header too long'), _OpenSSLErrorWithText(code=109052006, lib=13, reason=102, reason_text=b'error:06800066:asn1 encoding routines::bad object header'), _OpenSSLErrorWithText(code=109576458, lib=13, reason=524554, reason_text=b'error:0688010A:asn1 encoding routines::nested asn1 error'), _OpenSSLErrorWithText(code=109052027, lib=13, reason=123, reason_text=b'error:0680007B:asn1 encoding routines::header too long'), _OpenSSLErrorWithText(code=109052006, lib=13, reason=102, reason_text=b'error:06800066:asn1 encoding routines::bad object header'), _OpenSSLErrorWithText(code=109576458, lib=13, reason=524554, reason_text=b'error:0688010A:asn1 encoding routines::nested asn1 error')])' (row 4)
```

jku:
I'm now porting my python-tuf changes on top of this one: I think there are still tweaks needed: marking draft until then

jku:
Marking ready-for-review again. 
* python-tuf branch is now updated: it's likely not final but close enough that it shows the required changes there and gives us confidence the sslib api is decent
* the big difference for python-tuf code was that `Key.from_securesystemslib_key()` no longer exists. This feels like a correct change to me but it is an API change. We could add a workaround for backwards compat but I think I'm fine with changing API here?

lukpueh:
> @lukpueh does this look familiar (just happened on CI, never seen before myself):

No it does not.

lukpueh:
> * the big difference for python-tuf code was that `Key.from_securesystemslib_key()` no longer exists. This feels like a correct change to me but it is an API change. We could add a workaround for backwards compat but I think I'm fine with changing API here?

I think it's more than just `from_securesystemslib_key`. The python-tuf pr completely removes `tuf.api.Key`, right? It has almost an identical replacement in `SSlibKey`, but it is still a breaking change and warrants a new major release. It should only affect repositories, though. Client application code is very unlikely to directly use the `Key` class. So I'm fine with it.

jku:
> I think it's more than just `from_securesystemslib_key`. The python-tuf pr completely removes `tuf.api.Key`, right? It has almost an identical replacement in `SSlibKey`, but it is still a breaking change and warrants a new major release. It should only affect repositories, though. Client application code is very unlikely to directly use the `Key` class. So I'm fine with it.

No, Key is still part of the public API -- it's just that it comes from securesystemslib. This is not as clear as it should be since we are not explicit about API: I think it would make sense to use `tuf/api/__init__.py` for that purpose ...

So python-tuf users can do from `tuf.api.metadata import Key` and I think the object they get will quack like the duck they expect... Obviously there are small differences, from_securesystemslib_key() being the major one. 

Now that I think about it  the `Key()` constructor is probably now also useless or broken, it makes no sense... should have a look at that.

securesystemslib/signer/_key.py
```diff
@@ -0,0 +1,207 @@
...
+            "rsassa-pss-sha256", "ed25519", and "ecdsa-sha2-nistp256".
+        keyval: Opaque key content
+        unrecognized_fields: Dictionary of all attributes that are not managed
+            by Securesystemslib
+
+    Raises:
+        TypeError: Invalid type for an argument.
+    """
+
+    __metaclass__ = abc.ABCMeta
```
jku, line 39: this apparently does not do the same things as deriving from ABC: calling Key() should fail now as Key methods are not all implemented, but it does not. If Key derives from ABC, constructor starts failing as expected

jku:
Related comment on Signer:
* it has the same issue where `Signer()` seems to work but the resulting signer is useless...
* since this is the case... I wonder if we should just rename `Signer.from_priv_key_uri()` to `Signer()` -- it's clear that the constructor isn't being used since it does nothing useful?

securesystemslib/signer/_key.py
joshuagl: Nice find here. I'm genuinely surprised it took us so long to notice... I opened a draft PR in #465 (need to test it with some sslib dependents before it's ready for merge).

joshuagl (commented):
I really like this, nice work!  I left a couple of minor comments, notably a request to document the dynamic dispatch method and how folks add their own implementations of the abstract base classes and wire them in.

securesystemslib/signer/_key.py
```diff
@@ -0,0 +1,207 @@
...
+                UnverifiedSignatureError: it contains VerificationError as well
+        """
+        raise NotImplementedError
+
+
+class SSlibKey(Key):
+    """Key implementation for RSA, Ed25519, ECDSA and Sphincs keys"""
+
+    def to_securesystemslib_key(self) -> Dict[str, Any]:
+        """Internal helper, returns a classic securesystemslib keydict"""
```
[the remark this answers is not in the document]
jku: I _think_ no-one else should ever need to do this, only SSlibKey does. In any case this is from the (Key) API in python-tuf, I did not want to change that in this PR.

securesystemslib/signer/_key.py
```diff
@@ -0,0 +1,207 @@
...
+        ) or not isinstance(keyval, dict):
+            raise TypeError("Unexpected Key attributes types!")
+        self.keyid = keyid
+        self.keytype = keytype
+        self.scheme = scheme
+        self.keyval = keyval
+        if unrecognized_fields is None:
+            unrecognized_fields = {}
+
+        self.unrecognized_fields = unrecognized_fields
```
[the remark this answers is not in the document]
jku: this is also moved from python-tuf but I agree that likely makes sense

securesystemslib/signer/_key.py
jku: Ok, I'll probably try to fix this one here then.

securesystemslib/signer/_key.py
```diff
@@ -0,0 +1,207 @@
...
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(
+        self,
+        keyid: str,
+        keytype: str,
+        scheme: str,
+        keyval: Dict[str, Any],
+        unrecognized_fields: Optional[Dict[str, Any]] = None,
+    ):
```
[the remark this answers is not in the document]
jku: on the other hand -- all spec compliant keys should have these fields and we should have the type checks somewhere so maybe this does make sense here :shrug:

jku:
fixed a few latest comments (thanks!): No real functional changes here.

securesystemslib/signer/_key.py
```diff
@@ -0,0 +1,202 @@
...
+
+    @classmethod
+    def from_securesystemslib_key(cls, key_dict: Dict[str, Any]) -> "SSlibKey":
+        """Constructor from classic securesystemslib keydict"""
+        # ensure possible private keys are not included in keyval
+        return SSlibKey(
+            key_dict["keyid"],
+            key_dict["keytype"],
+            key_dict["scheme"],
+            {"public": key_dict["keyval"]["public"]},
```
jku, line 138: this looks weird: I think I made a mistake...

lukpueh: It's a *heisenpatch*

securesystemslib/signer/_key.py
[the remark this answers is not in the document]
jku: just to document this further
* to_securesystemslib_key is already pretty much internal
* from_securesystemslib_key is only needed until we add import/generation to SSlibKey #466

I'm resolving this, please make noise if you disagree

joshuagl (approved):
LGTM

securesystemslib/signer/_key.py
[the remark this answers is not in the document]
lukpueh: Not actually equivalent: https://github.com/theupdateframework/python-tuf/issues/1937

securesystemslib/signer/_key.py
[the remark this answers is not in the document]
lukpueh: > This should probably be the SSlibKey constructor: If we made the Key serialization methods abstract then construction isn't really possible and we wouldn't want a other implementations to call this constructor

You seem to have resolved this issue by yourself, but I'm curious what you mean here. Why would we not want SSlibKey or other implementations use this as default constructor?

securesystemslib/signer/_signature.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def __init__(
+        self,
+        keyid: str,
+        sig: str,
+        unrecognized_fields: Optional[Mapping[str, Any]] = None,
+    ):
+        self.keyid = keyid
+        self.signature = sig
+        self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
```
lukpueh, line 34: See https://github.com/theupdateframework/python-tuf/issues/1937

securesystemslib/signer/_key.py
```diff
@@ -0,0 +1,202 @@
...
+    ):
+        if not all(
+            isinstance(at, str) for at in [keyid, keytype, scheme]
+        ) or not isinstance(keyval, dict):
+            raise TypeError("Unexpected Key attributes types!")
+        self.keyid = keyid
+        self.keytype = keytype
+        self.scheme = scheme
+        self.keyval = keyval
+        self.unrecognized_fields = unrecognized_fields or {}
```
lukpueh, line 55: See https://github.com/theupdateframework/python-tuf/issues/1937

lukpueh (approved):
This is so neat! Many thanks, Jussi. I'm approving, although we should probably check for None instead of Falseyness when creating the default `self.unrecognized_fields` value. But this is no blocker.

Also, there is an annotation inconsistency in Key and Signature unrecognized_fields constructor argument and instance variables (Dict vs. Mapping, and type hint at declaration in one but not the other). You probably just copied those over. 🤷

securesystemslib/signer/_signature.py
jku: falsyness point is 100% true.

mapping was my annotation mistake at some point I think: We have already fixed that from similar fields in python-tuf: I'll do  the same thing here.

PR #457, opened 2022-11-14
Strip public key pem value

danbev:
This commit adds a call to strip the public key pem value in `import_ecdsakey_from_private_pem`.

The motivation for this that a later call to `import_ecdsakey_from_public_pem` will otherwise cause an incorrect `keyid` to be generated. This is because the call to `extract_pem` will strip any new line character from the public pem, which will then be used to generate the `keyid`, but this will be a different value than the keyid generated earlier due to the removal of the new line character.

### DIFF
--- a/securesystemslib/keys.py
+++ b/securesystemslib/keys.py
@@ -1725,7 +1725,7 @@ def import_ecdsakey_from_private_pem(
     # information is not included in the generation of the 'keyid' identifier.
     # Convert any '\r\n' (e.g., Windows) newline characters to '\n' so that a
     # consistent keyid is generated.
-    key_value = {"public": public.replace("\r\n", "\n"), "private": ""}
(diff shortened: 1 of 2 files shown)


jku:
this looks a lot like #453 just on a slightly higher level: I assume that one would likely fix the issue for you too?

I'll also leave the same comment as in that pr: I would not trust the keyid generation to be stable and would instead generate a keyid once and then use that consistently in the application.

danbev:
> this looks a lot like #453 just on a slightly higher level: I assume that one would likely fix the issue for you too?

Thanks for pointing that out, and yes I think that would solve this issue.

> I would not trust the keyid generation to be stable and would instead generate a keyid once and then use that consistently in the application.

Ah I see, good to know that and I'll do that. Thanks!

PR #461, opened 2022-11-23
Replaced `setuptools` with `hatchling`

KOLANICH:

- [x] Tests have been added for the bug fix or new feature - not needed
- [x] Docs have been added for the bug fix or new feature - not needed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
-requires = ["setuptools>=61.2", "setuptools_scm[toml]>=3.4.3"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling", "hatch-vcs"]
+build-backend = "hatchling.build"
 
 [project]
 name = "securesystemslib"
@@ -53,5 +53,8 @@ include-package-data = true
 [tool.setuptools.packages.find]
 exclude = ["tests"] # debian
 namespaces = false
+
+[tool.hatch.version]
+source = "vcs"
 
 [tool.setuptools_scm]


pyproject.toml
```diff
@@ -1,6 +1,10 @@
 [build-system]
-requires = ["setuptools>=61.2", "setuptools_scm[toml]>=3.4.3"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling", "hatch-vcs"]
+build-backend = "hatchling.build"
+
+# for the ones who prefer setuptools
+#requires = ["setuptools>=61.2", "setuptools_scm[toml]>=3.4.3"]
+#build-backend = "setuptools.build_meta"
```
ofek, line 7: If someone wants to do this they don't need a comment, let's remove

ofek (changes requested):

KOLANICH:
<details>
<summary>offtop</summary>
@ofek, you have done a good job creating your software (I'm considering to switch to it myself, and have even implemented support of hatchling in my software that used to be a "`setuptools` plugin", but TBH, again, you know my position, IMHO it makes no sense to switch build backends (which seem to be doing the same job and having almost the same features) from the de-facto standard `setuptools`, I'm considering switching only because `hatchling` seems to be becoming the official build backend), but your aggressive promotion of it (I won't be surprised if I'm told that it is the way it has gained adoption, Googlag Chrome has also succesfully outcompeted Firefox only by being aggressively promoted the ways Mozilla hadn't used), as well as determination to get rid of the competing build backends, worries me a bit. I understand that competition is inevittable and that the laws of nature and principles of game theory are the same for Free Software, but in Free Software ecosystem it feels odd and not right when someone uses the same methods businesses use to ensure their monopoly. In Free Software we don't compete, we cooperate. It means if there is a product that is the de-facto standard and if there are no sufficient flaws in it, there is no reason to fragment the ecosystem and make the things worse for everyone, and there is no reason to waste resources on reinventing an **own** wheel (and then aggresively promoting it), one should just improve the existing software. I see no flaws within `setuptools` significant enough to justify creation and adoption of `hatchling`, `poetry_core`, `flit_core` and some other. They all just build wheels, and the most advanced of them call plugins during it.
</details>

ofek:
Not going there with you again 😉

Please understand you are causing confusion to users by adding those commented out lines to all of your PRs. No one needs those, not even you and your custom build pipeline.

ofek (commented):
Nice! You can remove:

```toml
[tool.setuptools]
include-package-data = true
[tool.setuptools.packages.find]
exclude = ["tests"] # debian
namespaces = false

[tool.setuptools_scm]
```

ofek:
https://github.com/secure-systems-lab/securesystemslib/pull/484

Issue #463, opened 2022-11-25
Signer uses python2 "__metaclass__"


jku:
signer.py does `__metaclass__ = abc.ABCMeta`. This is python2 syntax and does nothing. So calling `Signer()` works fine but the resulting object does not implement the Signer API (the sign method is unimplemented) 

The actually supported methods are
```python
from abc import ABC
class MyABC(ABC)
```
and
```python
from abc import ABCMeta
class MyABC(metaclass=ABCMeta)
```
the latter may be safer (WRT multiple inheritance issues)

Found in #456

jku:
also StorageBackendInterface

lukpueh:
This might be a bit of a digression, but I think we should be using `typing.Protocol` in most cases, where we now use `abc`. This is at least true for `Signer` and `StorageBackendInterface), where the base classes don't provide any re-usable code and we don't care if the implementations are actually subclasses of these interfaces. We just want to make sure that the implementations quack like Signers or StorageBackendInterface.

`typing.Protocol` is not in Python 3.7, but a backport exists. What do others think?

jku:
In general I like the protocol idea for things like Signer (where the intent is to document to 3rd parties what they need to do). Caveats though:
* signer dispatch would still need to be somewhere
* anything with serialize/deserialize (like Key) seems trickier


The one  issue I have with Protocol is that the implementations never declare themselves (compare with rust Traits which you always declare with a `impl ThisTrait for ThatType`). This is very pythonic of course but it's also annoying in a very pythonic way: Does `MySpecialClass` implement the `Signer` protocol? Who knows! Go ahead and read the full source code to find out...

joshuagl:
I agree that `typing.Protocol` seems like a good fit for our ABC's which exist only to define an interface and do not provide any base or shared functionality.

PR #470, opened 2022-11-29
Convert `StorageBackendInterface` to `typing.Protocol`

joshuagl:
Fixes: #

As reported in #463 the `__metaclass__` attribute no longer works in Python 3. Furthermore, as proposed in https://github.com/secure-systems-lab/securesystemslib/issues/463#issuecomment-1330338153 `typing.Protocol` is a better fit for places we use `AbstractBaseClass` to define interfaces without any attached implementation.

This PR first fixes the ABC use in storage.py to match Python3 syntax, then in a follow-on commit changes from using an ABC to a Protocol as the base class for `StorageBackendInterface`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -34,3 +34,4 @@
 cryptography >= 37.0.0; python_version >= '3'
 pynacl
 PySPX
+typing-extensions; python_version < '3.8'
--- a/securesystemslib/storage.py
+++ b/securesystemslib/storage.py
@@ -22,9 +22,9 @@ import os
 import shutil
 import stat
 import sys
-
 from contextlib import contextmanager
 from typing import IO, BinaryIO, Iterator, List, Optional
+
 if sys.version_info < (3, 8):
     from typing_extensions import Protocol
 else:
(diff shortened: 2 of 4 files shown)


securesystemslib/storage.py
```diff
@@ -96,7 +99,7 @@ def put(
         """
         raise NotImplementedError  # pragma: no cover
 
-    @abc.abstractmethod
+
```
joshuagl, line 102:
```suggestion
```

requirements.txt
```diff
@@ -34,3 +34,4 @@
 cryptography >= 37.0.0; python_version >= '3'
 pynacl
 PySPX
+typing-extensions; python_version < '3.8'
```
joshuagl, line 37: Whoops, need to update requirements-pinned.txt

joshuagl:
It turns out that the linter doesn't like `FilesystemBackend` having functions without docstrings:
>securesystemslib/storage.py:198:4: C0116: Missing function or method docstring (missing-function-docstring)
securesystemslib/storage.py:212:4: C0116: Missing function or method docstring (missing-function-docstring)
securesystemslib/storage.py:278:4: C0116: Missing function or method docstring (missing-function-docstring)

I should probably convert storage.py to the more modern docstring style

joshuagl:
Let's abandon this. We don't make any significant improvements for API users and the contortions to use typing.Protocol with Python 3.7 are ugly.

Issue #466, opened 2022-11-29
Design key import/generation for the signer API


jku:
With #456 the signer API comes a lot closer to being a "fully featured API". The missing part is key generation and import (and maybe storing private keys)

Related issues: #446, #408, #280, #451 (and probably more)

## Current status (after #456):
* Using signers and keys is possible without knowing the signer or key implementation, and without using the "old" API
* **except** to create keys this sort of code is typically used:
  ```
  key = keys.generate_ed25519_key() # key is a dictionary. keys is a module we would like to start hiding from users
  pubkey = SSlibKey.from_securesystemslib_key(key) # pubkey is an implementation of Key
  signer = Signer.from_priv_key_uri("file:path/to/file?encrypted=false", pubkey) # signer is an implementation of Signer
  
  ```
* the key dictionary is only used as an intermediate step

## Proposal
* Signer implementations should provide implementation specific methods to create new keys or import them from other systems
* these methods may have whatever arguments they need
* Signer should expose the public key so callers can access it
* typical usage:
  ```
  # call implementation specific API (in this case SSlibsigner to generate a ed25519 key)
  signer = SSlibSigner.new_ed25519(...) 
  pubkey = signer.public_key
  ```
* this hides the `keys` module from users, and makes `SSlibKey.from_securesystemslib_key` unneeded
* AFAICT this should work for KMS, HSM, GPG, ...
* The implementation specific method could live elsewhere as well, but it seemed like it would make most sense in signer.

---

The part I have not 100% figured out is private key storage: how and when do you store the private key content in a way that you can then use the `from_priv_key_uri()` to load it again.  This is somewhat different depending on the key type and in many cases does not even make sense (you can't really store environment variables)... possibly it should happen during  the constructor, so `SSlibSigner.new_ed25519() ` could actually take a priv_key_uri argument and and the generated key would immediately get stored ?

jku:
As background for why the public_key should be a Signer property, and why I think import/generate-methods should be Signer-constructors:

_In real-world scenarios, if you generate/import a Key, you always want to "connect" that to a specific signer, or at least the private key uri: you never want to just generate a Key because then how would you know how to sign for that Key? Somehow you need to be able to connect a signer private key uri to a keyid: we can leave that to the user of course but if both are properties of the Signer, that gets solved neatly_

As an example if you import a public key from a Yubikey, you then always want to store configuration like
[...]

Issue #445, opened 2022-10-26
should support KMS signing


jku:
Signing using a KMS would be useful: as an example a TUF repository could sign metadata with online keys without having access to the private key material.

The implementation should not be difficult but there are a couple of tricky bits:
* testing requires a KMS (mocking is not very useful in this case)
* the authentication to KMS is going to complicate setting up (but likely not the code)
* Signing (private) key identification may require a rethink to make the API most usable. See the comment below
* signing is the most important part, but smooth key changes (rotation etc) may require some additional functionality: key generation and public key extraction. See #446

I have a draft PR #442

jku:
## Signing (private) key identification

The discussion that maybe makes sense at this point is private key abstraction in general -- in other words how to decide which  Signer to use and how to initialize that signer. Since we don't have a solution for this, we end up with having to change code when the keytype changes:

```python
SSLibSigner(key_dict) # here key_dict is expected to include public and private key 
GPGSigner(gpg_key_id)
KMSSigner.from_keyref(keyref, key_dict)
```

If there was a global way to identify a private key (_privkeyref_) that we can load from configuration, we could just have a single call that just works:
`Signer.from_private_keyref(privkeyref, key_dict)
`
This method would act as a factory method/dispatcher that returns the correct Signer subclass -- see KMSSigner in this PR for an example that currently dispatches to just one subclass.

I wonder if the simple "privkeyref" string approach in the linked PR would actually work in general? The idea here is that the public key is provided -- we only need the private key details from the private key storage:

* GCP KMS:
  `"gcpkms://projects/openssf/locations/global/keyRings/securesystemslib-test-keyring/cryptoKeys/securesystemslib-test-key/cryptoKeyVersions/1"`
* GPG
  `"gpg://C5A0ABE6EC19D0D65F85E2C39BE9DF5131D924E9"`
* keyval.private content
  `"raw://788b012147f9afe9c1a151...f94dc32a281baa"`
* file with keyval.private content
  `"file:///home/jku/rsa_key"`

The point of keeping this as "just a string" is to make it easy to load the private key identifier from configuration and get a Signer without modifying code (this should make it easy to expose all of this via CLI tools if needed)

jku:
I've moved the private key id proposal to #447

jku:
This is done now I suppose: Google Cloud KMS is supported and working fine. Other KMS implementations can be added.

PR #472, opened 2022-12-01
Add HSMSigner

lukpueh:
*based on #229*
*updated PR title and description on 12/7/2022*
*un-draft on 12/12/2022*

--
**Description**
Adds basic HSMSigner implementation to generate signatures on a hardware security module and to export the corresponding public key (as SSlibKey).

HSMSigner uses the first token it finds, if multiple tokens are available, and can be instantiated directly or with `Signer.from_priv_key_uri("hsm:", public_key, secret_handler)`. The former can be passed a keyid, the latter uses the key on the digital signature default piv slot 9c.

Supported keys are ecdsa on SECG curves secp256r1 (NIST P-256) or secp384r1 (NIST P-384), which correspond to securesystemslib signing schemes "ecdsa-sha2-nistp256" and "ecdsa-sha2-nistp384".

Tests are performed on SoftHSM (virtual hsm).

**Future work**
* If we want to support schemes that allow arbitrary token reader slots, e.g.
  `hsm:YubiKey%20PIV%20%2315835999?piv_slot=9c` we need to change the HSMSigner constructor, to e.g.  add an `hsm_label` argument. This could be an optional non-API-breaking addition.
* If we want to export public keys automatically, e.g. if no key is passed to 'from_priv_key_uri', we also need  to pass the sslib keyid via uri parameter, or use generate a random/canonical keyid (the choice should maybe also be parameterizable  

**Notes to reviewer**
- ~~Keeping this as draft until [#442](https://github.com/secure-systems-lab/securesystemslib/pull/442) is merged (please ignore Jussi's commits until then)~~
- The first of my commits is what started this PR. Everything afterwards is new.
- I don't recommend reviewing commit by commit. They tell an engineering story, but are not all relevant for reviewers.
- If desired, I can spend some time on rewriting the commit history, which will likely be just squashing everything. :)
- Ping me if you have questions.


lukpueh:
<details>
<summary>Usage example (tried on YubiKey 5 Nano)</summary>

```bash
# Create and sign key
yubico-piv-tool -a generate -s 9c -A ECCP256 -k -o public.pem
yubico-piv-tool -a verify-pin -a selfsign -s 9c -i public.pem -S /CN=a.b.c/OU=abc/O=b.c/ -o cert.pem
yubico-piv-tool -k -a import-certificate -s 9c -i cert.pem
```

```python
from PyKCS11 import PyKCS11
from securesystemslib.signer import HSMKey, HSMSigner


lib = "<PATH/TO/PKCS11/LIBRARY>"
#     e.g. "/usr/local/Cellar/yubico-piv-tool/2.3.0/lib/libykcs11.dylib" (via `brew`)
hsm_keyid = "<HSM KEY IDTUPLE>"
#     e.g. `(2,)`; see `pkcs15-tool --list-key`
pin = "<YOUR USER PIN>"

sslib_keyid = "0"*64

pkcs11 = PyKCS11.PyKCS11Lib()
pkcs11.load(lib)
slot = pkcs11.getSlotList(tokenPresent=True)[0]
session = pkcs11.openSession(slot)

pubkey = HSMKey.from_hsm(session, hsm_keyid, sslib_keyid)

session.login(pin)
signer = HSMSigner(session, hsm_keyid, pubkey)
sig = signer.sign(b"DATA")
session.logout()
session.closeSession()

pubkey.verify_signature(sig, b"DATA")
```
</details>

pyproject.toml
```diff
@@ -46,6 +46,7 @@ Issues = "https://github.com/secure-systems-lab/securesystemslib/issues"
 crypto = ["cryptography>=37.0.0"]
 pynacl = ["pynacl>1.2.0"]
 PySPX = ["PySPX==0.5.0"]
+hsm = ["asn1crypto", "cryptography", "PyKCS11"]
```
lukpueh, line 49: Should this be: 
```suggestion
asn1crypto = ["asn1crypto"]
PyKCS11 = ["PyKCS11"]
```
?

For the sake of consistency, it should. Also,  specifying optional dependencies per feature is prone to redundancy (see cryptography).

Either way, we should document which optional dependency is needed for which currently available feature.

lukpueh (commented):
Some comments about testing, CI and dependencies...

jku:
> identifying the correct private key (aka what needs to be in private key URI for this signer). I'm really not sure what these are exactly so have no opinions... Have you found good docs on what these really are?
> * id tuple?
> * slot?
> * label?

So first confusion source is that pkcs11 slot seems to be completely different from what yubikey docs call a slot??? I suppose PKCS11 slot is something that can e.g. change over reboots etc? 

So we want to be able to potentially select
* token  _(but a good first implemenation would be to assume only one token is available)_
* the key within the token (by CKA_ID or CKA_LABEL?) which _maybe_ maps to yubico slots like "9c" . _(If we can tell which CKA_ID is "9c" we could start by only supporting "9c"?)_

jku:
> The HSMSigner could still receive the pin as constructor argument. Or would you just pass on the handler?

:shrug: I think handler as argument to constructor makes sense?

lukpueh:
> > identifying the correct private key (aka what needs to be in private key URI for this signer). I'm really not sure what these are exactly so have no opinions... Have you found good docs on what these really are?
> > 
> > * id tuple?
> > * slot?
> > * label?
> 
> So first confusion source is that pkcs11 slot seems to be completely different from what yubikey docs call a slot??? I suppose PKCS11 slot is something that can e.g. change over reboots etc?

Yeah, YubiKey docs refer to [PIV certificate slots](https://developers.yubico.com/PIV/Introduction/Certificate_slots.html) (PIV is a separate standard). In [PKCS11](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html) slots and tokens are defined as:

- Slot    --   A logical reader that potentially contains a token.
- Token   --  The logical view of a cryptographic device defined by Cryptoki.

Not sure what you mean by reboot, but I saw that slot ids can change even between pkcs11 sessions.

> 
> So we want to be able to potentially select
> 
> * token  _(but a good first implemenation would be to assume only one token is available)_

👍 

> * the key within the token (by CKA_ID or CKA_LABEL?) which _maybe_ maps to yubico slots like "9c" . Can the user see CKA_ID or CKA_LABEL anywhere? _(Potential first implementation: If we can tell which CKA_ID is "9c" we could start by only supporting "9c"?)_

In my example I just used `"9c"` when creating the key, because YubiKey docs suggest that it is semantically the correct slot for a signing key. I don't think our API needs to care about the semantics of PIV slots. They are not related to the hsm keyid.

The hsm keyid can be  determined e.g. with 

```bash
 $ pkcs11-tool -lO
Using slot 0 with a present token (0x0)
Logging in to "host.example.com".
Please enter User PIN:
Private Key Object; EC
  [...]
  ID:         03
  [...]
Public Key Object; EC  EC_POINT 256 bits
  [...]
  ID:         03
  [...]
```

It's the same for public and private key and remains stable.

lukpueh:
Thanks for digging, @jku! Just tried `pkcs11-tool` with `libykcs11.dylib` instead of the default `opensc-pkcs11.so`  module and now I see a lot more keys on the yubikey, and the mapping seems to match the table. 🎉  

```
$ pkcs11-tool --module /usr/local/lib/libykcs11.dylib -O
[...]
Public Key Object; EC  EC_POINT 256 bits
  label:      Public key for Digital Signature
  ID:         02
  [...]
Public Key Object; EC  EC_POINT 256 bits
  label:      Public key for Key Management
  ID:         03
  [...]
```

I was already wondering, why I can pass IDs 2 and 3, to sign, when, according to pkcs11-tool, I only have one signing key at ID 3.

jku:
So then the version 1 implementation could
* assume first token is the one we want
* assume key id is 02
* document that this has been tested with libykcs and PIV slot "9c"

then the first version would need no path or query params for the private key URI: `pkcs11:` or `hms:` would be enough. Then later we could add optional paths/params if needed: `hms:YubiKey%20PIV%20%2315835999?token_slot=9c`

lukpueh:
Just rebased on top of #445 to pull in 16734bb (for sub-modules) and added a bunch of WIP-commits based on our latest discussion. Will ping for re-review when ready.

lukpueh:
Okay, this is ready for re-review. See updated PR description for what changed.

tox.ini
```diff
@@ -18,7 +18,7 @@ deps =
 commands =
     python -m tests.check_gpg_available
     coverage run tests/aggregate_tests.py
-    coverage report -m --fail-under 97
+    coverage report -m --fail-under 95
```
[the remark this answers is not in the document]
lukpueh: This still needs to be 95%, but the reason changed. The PR now also pulls in #442, which is excluded from `aggregate_tests`. I should be able to increase the number was that #442 is merged.

securesystemslib/signer/_hsm_signer.py
```diff
@@ -0,0 +1,301 @@
...
+    )
+except ImportError:
+    ASN1_IMPORT_ERROR = "'asn1crypto' required"
+# pylint: enable=wrong-import-position
+
+
+_PYKCS11LIB = None
+
+
+def PYKCS11LIB():
```
lukpueh, line 70: Question for reviewers: Any better ideas than this? See 6b825a0ee4aea04107f9354b91c566ac7ed52294 for context.

pyproject.toml
```diff
@@ -44,8 +44,11 @@ Issues = "https://github.com/secure-systems-lab/securesystemslib/issues"
 
 [project.optional-dependencies]
 crypto = ["cryptography>=37.0.0"]
+gcpkms = ["google-cloud-kms"]
 pynacl = ["pynacl>1.2.0"]
 PySPX = ["PySPX==0.5.0"]
+asn1 = ["asn1crypto"]
+pykcs11 = ["PyKCS11"]
```
jku, line 51: feels like this isn't the correct way to use this system. pykcs11 without asn1 doesn't make sense so why offer that as option?

jku (approved):
Looks good to merge to me. Left comments but I don't think any of them are blockers -- although the optional-dependencies discussion might be good to have before merge

lukpueh:
Just force-pushed a pure rebase without changes and marked as non-draft.

securesystemslib/signer/_hsm_signer.py
```diff
@@ -0,0 +1,301 @@
...
+        key = cls._find_key(session, keyid)
+        params, point = session.getAttributeValue(
+            key, [PyKCS11.CKA_EC_PARAMS, PyKCS11.CKA_EC_POINT]
+        )
+        return ECDomainParameters.load(bytes(params)), bytes(point)
+
+    @classmethod
+    def pubkey_from_hsm(
+        cls, sslib_keyid: str, hsm_keyid: Optional[int] = None
+    ) -> SSlibKey:
```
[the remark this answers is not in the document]
lukpueh: I agree with you. Let's fix this with #466. Until then, do you think it's better to merge as non-public `_pubkey_from_hsm`?

securesystemslib/signer/_hsm_signer.py
```diff
@@ -0,0 +1,301 @@
...
+from urllib import parse
+
+from securesystemslib import KEY_TYPE_ECDSA
+from securesystemslib.exceptions import UnsupportedLibraryError
+from securesystemslib.signer._key import Key, SSlibKey
+from securesystemslib.signer._signature import Signature
+from securesystemslib.signer._signer import SecretsHandler, Signer
+
+# pylint: disable=wrong-import-position
+CRYPTO_IMPORT_ERROR = None
```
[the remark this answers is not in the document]
lukpueh: I think the separation makes sense wrt current extra dependency handling. Let's fix those together.

pyproject.toml
lukpueh: `asn1crypto` is only an immediate dependency for pubkey export. For signing we don't use it.

lukpueh: But I agree, telling users to run
```
python securesystemslib[crypto, asn1, pykcs11]
```
is not a lot better, than
```
python cryptography asn1crypto pykcs11 securesystemslib
````

lukpueh: Listing extra dependencies by feature, otoh, seems prone to redundancy. `cryptography`, for instance, is used for many different features.

lukpueh:
Thanks for the review, @jku! I addressed your comments wrt docs and secrets_handler. Would you mind if we take a look at [extra dependency definition](https://github.com/secure-systems-lab/securesystemslib/pull/472#discussion_r1043231971) (and [related error handling](https://github.com/secure-systems-lab/securesystemslib/pull/472#discussion_r1043179453)) in a follow-up? It's not obvious to me, what the best solution is.

securesystemslib/signer/_hsm_signer.py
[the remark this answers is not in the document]
jku: I think it's fine as is: if we fix before next release it's a non-issue; if we fix in some later release it's a minor API break

jku:
I'm happy with this, LGTM.

Issue #363, opened 2021-06-23
Deprecate ecdsa-sha2-nistp256 keytype


avelichka:
**Description of issue or feature request**:
Support for ecdsa as keytype was added in https://github.com/secure-systems-lab/securesystemslib/pull/267 and we should deprecate the old keytype.

PratGit1606:
I'd like to work on this. Planning to add a DeprecationWarning in the relevant codepaths and update docs/changelog. Will open a PR shortly.

PR #480, opened 2022-12-12
GCPSigner import

jku:
Implement import  (loading the whole key, including public key, from the KMS) for GCPSigner.

This is the implementation of #466 for this specific Signer -- in my opinion this can be merged: if we decide to change something in #466 that affects this Signer we can do that later.

At this point GCPSigner is fully usable: I don't think it needs anything more.

### DIFF
--- a/securesystemslib/signer/_gcp_signer.py
+++ b/securesystemslib/signer/_gcp_signer.py
@@ -66,7 +66,6 @@ class GCPSigner(Signer):
 
         self.hash_algorithm = self._get_hash_algorithm(public_key)
         self.gcp_keyid = gcp_keyid
-        self.priv_key_uri = f"{self.SCHEME}:{gcp_keyid}"
         self.public_key = public_key
         self.client = kms.KeyManagementServiceClient()
 
--- a/securesystemslib/signer/_gcp_signer.py
+++ b/securesystemslib/signer/_gcp_signer.py
@@ -104,6 +104,7 @@ class GCPSigner(Signer):
 
         self.hash_algorithm = self._get_hash_algorithm(public_key)
         self.gcp_keyid = gcp_keyid
+        self.priv_key_uri = f"{self.SCHEME}:{gcp_keyid}"
         self.public_key = public_key
         self.client = kms.KeyManagementServiceClient()
 
--- a/securesystemslib/signer/_gcp_signer.py
+++ b/securesystemslib/signer/_gcp_signer.py
@@ -1,7 +1,7 @@
 """Signer implementation for Google Cloud KMS"""
 
 import logging
-from typing import Optional
+from typing import Optional, Tuple
 from urllib import parse
 
 import securesystemslib.hash as sslib_hash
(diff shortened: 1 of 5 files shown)


jku:
test run in https://github.com/jku/securesystemslib/actions/runs/3677912828/jobs/6220505707

jku:
Marking ready for review

jku:
I'll add a couple of commits that try to "standardize" GCP and HSM "key creation / import": these change HSMSigner mostly but I think it makes some sense as part of this PR: the goal is to keep the offered Signer APIs coherent, even if there is no specific signature that the signers are implementing.

* use the same method name `import_()`
* same return values: priv key uri and public key
* and finally, handle keyid creation internally

The last one, keyid change,  is a separate commit so it can be dropped if you don't like it. My reasons for including it: 
* It's true that callers should have the opportunity to set keyids, but I think they do have that: they can just set the keyid of the key to whatever they want after getting it from import_()
* providing a reasonable default still seems nice (the implementation may not be pretty but now it's an implementation detail that we can change without API changing)

Neither signer has been released yet so API changes are fine.

securesystemslib/signer/_gcp_signer.py
```diff
@@ -71,6 +84,79 @@ def from_priv_key_uri(
...
+        public_key = SSlibKey(keyid, keytype, scheme, keyval)
+
+        return f"{cls.SCHEME}:{gcp_keyid}", public_key
+
+    @staticmethod
+    def _get_keytype_and_scheme(algorithm: int) -> Tuple[str, str]:
+        """Return keytype and scheme for the KMS algorithm enum"""
+        keytypes_and_schemes = {
+            CryptoKeyVersion.CryptoKeyVersionAlgorithm.EC_SIGN_P256_SHA256: (
+                "ecdsa-sha2-nistp256",
```
lukpueh, line 118: Minor nit: https://github.com/secure-systems-lab/securesystemslib/issues/363 suggests to deprecate `ecdsa-sha2-nistp256` as keytype in favor of the more succinct `ecdsa`.

securesystemslib/signer/_gcp_signer.py
```diff
@@ -71,6 +84,79 @@ def from_priv_key_uri(
...
+            CryptoKeyVersion.CryptoKeyVersionAlgorithm.RSA_SIGN_PKCS1_4096_SHA256: (
+                "rsa",
+                "rsa-pkcs1v15-sha256",
+            ),
+            CryptoKeyVersion.CryptoKeyVersionAlgorithm.RSA_SIGN_PKCS1_4096_SHA512: (
+                "rsa",
+                "rsa-pkcs1v15-sha512",
+            ),
+        }
+        return keytypes_and_schemes[algorithm]
```
lukpueh, line 158: Is it worth creating a ticket to test these all schemes?

lukpueh (commented):
LGTM. Please comment on `priv_key_uri` signer attribute and I'll approve. Other inline comments are non-blockers.

securesystemslib/signer/_gcp_signer.py
jku: oh, I must have gotten these mixed up

securesystemslib/signer/_gcp_signer.py
jku: I suppose that would be nice -- it will require creating all of these keys in the KMS though

securesystemslib/signer/_gcp_signer.py
```diff
@@ -54,6 +66,7 @@ def __init__(self, gcp_keyid: str, public_key: Key):
 
         self.hash_algorithm = self._get_hash_algorithm(public_key)
         self.gcp_keyid = gcp_keyid
+        self.priv_key_uri = f"{self.SCHEME}:{gcp_keyid}"
```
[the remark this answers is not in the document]
jku: I think there is no benefit -- I added that before we arrived at the conclusion that it makes sense as import return value. I will  remove the attribute

jku:
interesting, I broke the kms test. Investigating...

lukpueh:
[Test still failing](https://github.com/jku/securesystemslib/actions/runs/3894425830/jobs/6648399377). Could it be that you have to recalculate the keyid? 

https://github.com/secure-systems-lab/securesystemslib/blob/8f85dc8c84ad8b57641209ab556c5cb63751afa0/tests/check_kms_signers.py#L28

tests/check_kms_signers.py
```diff
@@ -18,13 +18,25 @@
...
-from securesystemslib.signer import Key, Signer
+from securesystemslib.signer import GCPSigner, Key, Signer
 
 
 class TestKMSKeys(unittest.TestCase):
     """Test that KMS keys can be used to sign."""
 
-    def test_gcp(self):
+    pubkey = Key.from_dict(
+        "23b650fcc538c0c6d423d0886448172fd51d81f43ef749851a0a84978cdcd8e0",
```
lukpueh, line 28:
```suggestion
        "218611b80052667026c221f8774249b0f6b8b310d30a5c45a3b878aa3a02f39e",
```
should work

jku:
Okay: the test was failing as it should since I changed the key content. I fixed the test content to match as suggested, test now passes: 

https://github.com/jku/securesystemslib/actions/runs/3894855218/jobs/6649416664

lukpueh (approved):

PR #484, opened 2022-12-14
build: use hatchling, don't use vcs integration

### LINKED ISSUE #483: build: Use hatchling (or get rid of setuptools_scm anyway)
setuptools_scm tries to be way more smart than it should and breaks `pip install git+https://github.com/jku/securesystemslib.git` which is useful for testing PR branches in downstream projects. Specifically it imagines the git tag 'debian/0.16.0-1' is a version number and asserts because it doesn't quite work.

I suggest we get rid of setuptools_scm...
* we currently use it for automated version setting from git describe  (and I sort of like that -- getting development version numbers for free is nice)
* hatchling has VCS support as well, so could just switch to hatchling (since I think that was the plan anyway) and hope the VCS support is better
* if VCS support is found to be a bad idea, hatchcling can read the version from `__init__.py`

jku:
* removes build dependencies
* makes securesystemslib build more like the one in python-tuf
* makes `pip install git+https://github.com/secure-systems-lab/securesystemslib.git` work again
* requires setting version in `__init__.py`

Fixes #483


lukpueh (approved):
Thanks, Jussi! I tried this locally and it works as expected. I also looked into sdist and it includes everything we want. 

Moving from setuptools to hatchling obviates the need for MANIFEST.in. Want to remove it?

Issue #370, opened 2021-08-31
Add support for DSSE


adityasaky:
**Description of issue or feature request**:

[Dead Simple Signing Envelope (DSSE)](https://github.com/secure-systems-lab/dsse) is the new signature wrapper recommended for use in in-toto (https://github.com/in-toto/ITE/blob/master/ITE/5/README.adoc) and TUF (pending https://github.com/theupdateframework/taps/pull/138). Naturally, it makes sense for its implementation to live in securesystemslib.

With ITE-5 up for acceptance (https://github.com/in-toto/ITE/issues/22), we are trying to define a transition period for in-toto during which the reference implementation supports generating and verifying metadata in both formats. This time period is dependent at least in part on the DSSE functionality being present in securesystemslib for in-toto to use. We can tentatively define a relative transition period, perhaps a year after in-toto starts supporting both wrappers.

**Current behavior**:

securesystemslib provides some crypto-specific primitives used by TUF and in-toto. Both projects maintain their own implementation of the current signature wrapper.

**Expected behavior**:

securesystemslib continues providing the functionality it currently does and adds DSSE support, allowing the python implementations of both TUF and in-toto to transition to the new wrapper.

cc @SantiagoTorres @joshuagl @mnm678 @trishankatdatadog

adityasaky:
For reference, there's currently support for DSSE in in-toto-golang: https://github.com/in-toto/in-toto-golang/tree/master/pkg/ssl. The DSSE specification also includes an implementation we can likely borrow from here: https://github.com/secure-systems-lab/dsse/tree/master/implementation

lukpueh:
Related: https://github.com/secure-systems-lab/securesystemslib/issues/272

lukpueh:
A working DSSE implementation is available in https://github.com/secure-systems-lab/securesystemslib/compare/master...in-toto:securesystemslib:master courtesy of @PradyumnaKrishna! ❤️  (see [GSoC blog post](https://onpy.in/posts/gsoc-22/) for context). The implementation was tailored for in-toto (https://github.com/in-toto/in-toto/pull/503) but should work for python-tuf as well.

The result is a ~1,4K LOC diff, which I have reviewed in the course of multiple downstream PRs. Other maintainers should still take a look before we upstream. So I suggest to either:
- PR all at once and keep the nicely curated commit history, or
- PR module by module and make reviews more fun.

[...]

PR #486, opened 2022-12-15
Make GPGSigner.sign return Signature

### LINKED ISSUE #448: GPGSigner (and gpg.functions.create_signature) produce invalid signature dicts

### LINKED ISSUE #471: Move GPGSigner to intoto
I think the plan is as follows (@lukpueh can correct):
* Move GPGSigner from securesystemslib to somewhere under intoto umbrella
* Also add GPGKey in the same place
* intoto can then use the new dispatch mechanisms (from #456) to support gpg in intoto with just a small amoun of hacks
  * e.g. use (None, None) as the keytype/scheme-tuple
  * handle the special keyid matching  

This is done because the gpg key and signature have serialization formats that are not quite compatible with the specifications. The goal is:
* intoto can keep supporting the current key and sig formats for backwards compat
* we can later add more specification compatible GPG support into securesystemslib (but this won't be compatible with the format in intoto)

lukpueh:
Fixes #448
Closes #471 

GPGSigner.sign now returns Signature, which maps to a in-toto and tuf spec compliant signature object. The lower level gpg signing and verifying API stays as is, to allow existing gpg users (in-toto) to create and verify signature using the old format.

This is an acceptable API break, because GPGSigner is not yet used by in-toto or python-tuf.

To make `GPGSigner` useable for tuf and DSSE ([#370](https://github.com/secure-systems-lab/securesystemslib/issues/370)) a spec compatible GPGKey implementation will be added in a follow-up PR.


jku (approved):
Seems good to me.

jku:
Although: the serialized signature format contains some extra fields when created with GPGSigner, right? It would be nice if these were obvious in the tests

lukpueh:
@jku, let me know if the last commit roughly aligns with what you had in mind.

jku:
Yeah LGTM. It's not clear from the test what is actually in the custom data but it's now visible to reader that there's something there: I'm happy with that.

PR #488, opened 2022-12-16
Add GPGKey and use with GPGSigner

### LINKED ISSUE #450: GPG public key schemas are incompatible with both TUF and in-toto specifications
Example:
```
        {
            "hashes": [
                "pgp+SHA2"
            ],
            "keyid": "7b3abb26b97b655ab9296bd15b0bd02e1c768c43",
            "keyval": {
                "private": "",
                "public": {
                    "e": "010001",
                    "n": "e9ad391502ae32bd4fcc41a0f9970f8901ed6ad1c5c128c02add22721cdc22318b64bec9f9467b6949b19fc2e98ce41906125ad45d0b138f1ad6c5da7bde38092d9e3e697ce8b8373b150b57342dd921d634b873f258f5c15559b52921fa4bb7f482ec43a1c85c3385bd520cedbdc16b2524a64aecf32ac5690e6dd4ee0210a975e1b6c5af164ea69ca64533422432070511068730594793885567bb8f7cffacf6eb5ffdc640e898e599579b21b15e497f5c052112c5fdf7974e7056cd1564fe84f207cb946d1efc521e5031299e6275936e6f9464a735bd4edc8e0cde3fe5b1bf6d3bc1ed12993b865d8fcb9d9a2b2ef2df30cb7f0ab4c0dea819ea017ff195"
                }
            },
            "method": "pgp+rsa-pkcsv1.5",
            "type": "rsa"
        }

```
and the schemas it uses:
```
GPG_RSA_PUBKEYVAL_SCHEMA = SCHEMA.Object(
    object_name="GPG_RSA_PUBKEYVAL_SCHEMA", e=SCHEMA.AnyString(), n=HEX_SCHEMA
)
_GPG_RSA_PUBKEY_SCHEMA = SCHEMA.Object(
    object_name="GPG_RSA_PUBKEY_SCHEMA",
    type=SCHEMA.String("rsa"),
    method=SCHEMA.String(GPG_RSA_PUBKEY_METHOD_STRING),
    hashes=SCHEMA.ListOf(SCHEMA.String(GPG_HASH_ALGORITHM_STRING)),
    creation_time=SCHEMA.Optional(UNIX_TIMESTAMP_SCHEMA),
    validity_period=SCHEMA.Optional(SCHEMA.Integer(lo=0)),
    keyid=KEYID_SCHEMA,
    keyval=SCHEMA.Object(
        public=GPG_RSA_PUBKEYVAL_SCHEMA, private=SCHEMA.String("")
    ),
)
```

This contradicts with the "rsa" type defined in both TUF and in-toto specifications (which define keyval.public as a PEM string). These keys work in in-toto because there's code in in-toto that uses a schema check to figure out which verify function to call, but the metadata is still not spec compliant, and would be impossible for another implementation to use without reverse engineering the reference implementation.

lukpueh:
*EDIT 2023/01/20:  remove key id param in private key URI*

*EDIT 2023/03/06:*
  - move min code coverage back to 90%, gpg signer is at 100%, nothing more we can do for coverage in this PR
  - remove optional fields from GPGKey, we shouldn't use advanced PKI concepts from GnuPG -- subkeys, key expiration, etc.  -- in tuf/in-toto metadata, which use these concepts themselves, instead assert that passed and returned key id match on import_ and sign
  - remove obsolete "hashes" field from GPGKey (this is  static for existing schemes, and should be encoded in the scheme for other schemes)

see https://github.com/secure-systems-lab/securesystemslib/pull/488#issuecomment-1398169012 pp. and commit messages for more details

Closes #450 

Kept as draft to,
- [x] rebase on merged [#486](https://github.com/secure-systems-lab/securesystemslib/pull/486)
- [x] figure out GPGKey main/sub-key id mapping to signer (see [discussion](https://github.com/secure-systems-lab/securesystemslib/pull/456#issuecomment-1315138451))
- [x] add more tests

GPGKey is a regular Key with additional GnuPG specific key fields,
and verification method. It also has conversion helpers to translate
from and to a non-in-toto/tuf-spec compliant key format, which is
still used by the underlying securesystemslib.gpg subpackage.

GPGSigner is updated to:
- take a GPGKey as constructor argument, and implement
- ``from_priv_key_uri``, to load signer from `gnupg:[<GnuPG homedir>]`
- ``import_``, to import a public key from a GnuPG keyring and return it along with a uri to create the signer.

### DIFF
--- a/securesystemslib/signer/__init__.py
+++ b/securesystemslib/signer/__init__.py
@@ -5,7 +5,7 @@ This module provides extensible interfaces for public keys and signers:
 Some implementations are provided by default but more can be added by users.
 """
 from securesystemslib.signer._gcp_signer import GCPSigner
-from securesystemslib.signer._gpg_signer import GPGSigner
+from securesystemslib.signer._gpg_signer import GPGKey, GPGSigner
 from securesystemslib.signer._hsm_signer import HSMSigner
 from securesystemslib.signer._key import KEY_FOR_TYPE_AND_SCHEME, Key, SSlibKey
 from securesystemslib.signer._signature import Signature
@@ -23,6 +23,7 @@ SIGNER_FOR_URI_SCHEME.update(
         SSlibSigner.FILE_URI_SCHEME: SSlibSigner,
         GCPSigner.SCHEME: GCPSigner,
         HSMSigner.SCHEME: HSMSigner,
+        GPGSigner.SCHEME: GPGSigner,
     }
 )
 
@@ -47,5 +48,8 @@ KEY_FOR_TYPE_AND_SCHEME.update(
         ("rsa", "rsa-pkcs1v15-sha384"): SSlibKey,
         ("rsa", "rsa-pkcs1v15-sha512"): SSlibKey,
         ("sphincs", "sphincs-shake-128s"): SSlibKey,
+        ("rsa", "pgp+rsa-pkcsv1.5"): GPGKey,
+        ("dsa", "pgp+dsa-fips-180-2"): GPGKey,
+        ("eddsa", "pgp+eddsa-ed25519"): GPGKey,
     }
 )
--- a/securesystemslib/signer/_gpg_signer.py
+++ b/securesystemslib/signer/_gpg_signer.py
@@ -1,10 +1,10 @@
 """Signer implementation for OpenPGP """
 
 import logging
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, Optional, Tuple
 from urllib import parse
 
-from securesystemslib import exceptions
+from securesystemslib import exceptions, formats
 from securesystemslib.gpg import exceptions as gpg_exceptions
 from securesystemslib.gpg import functions as gpg
 from securesystemslib.signer._key import Key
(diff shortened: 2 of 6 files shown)


securesystemslib/signer/_gpg_signer.py
```diff
@@ -35,41 +240,78 @@ def from_priv_key_uri(
...
+            raise ValueError(f"expected GPGKey for {priv_key_uri}")
+
+        uri = parse.urlparse(priv_key_uri)
+
+        if uri.scheme != cls.SCHEME:
+            raise ValueError(f"GPGSigner does not support {priv_key_uri}")
+
+        if secrets_handler is not None:
+            raise ValueError("GPGSigner does not support a secrets handler")
+
```
jku, line 253: I think this is not useful -- code that uses from_priv_key_uri() is aiming to work with as many signers as it can, using the shared API. This seems to force me to write two code paths for no gain.

jku (changes requested):
Looks pretty good to me: it's obviously a bit of a mess with going back and forth beween legacy gpg format and and shared one, but that's all hidden behind the signer  API now... 

I think requiring secretshandler to be none is wrong but that's the only complaint really.

Is there a test that ensures that the client usage (desdrialize + verify) works without gnupg being installed?

lukpueh:
rebased without changes

lukpueh:
Thanks for the review.

> it's obviously a bit of a mess with going back and forth beween legacy gpg format and and shared one, but that's all hidden behind the signer API now...

Agreed. This is not nice but can be easily fixed, once in-toto also transitions to the signer API.

> 
> I think requiring secretshandler to be none is wrong but that's the only complaint really.

Will change.

> Is there a test that ensures that the client usage (desdrialize + verify) works without gnupg being installed?

Will (fix and) add tests before un-drafting this PR.

securesystemslib/signer/_gpg_signer.py
lukpueh: @jku, do you think we should just ignore it? Maybe log a warning/info? With current implementation of `securesystemslib.gpg` we cannot pass any secrets to the gpg command.

jku: No need for warning as far as I am concerned: the way I imagine this being used (and the way I am using it) is that the signing code doesn't care which signers are used or what they do:
* online signing app likely does not have a secrets handler
* an offline signing tool does provide a secrets handler
* even the offline signing tool _does not care whether the secrets handler is used or not_: a gpg signing might not use it but a yubikey signing will

lukpueh:
Thanks for the comments, @jku! I've addressed your concerns and added plenty of test, which also revealed a serialization bug  (fixed in 3d140d5306cd805f0b15cd43f119d205f7bc8959).

Beyond that, I removed the keyid field from GPGSigner, given that it also has a GPGKey attached, which has the same user-assigned keyid.

I am still unsure how to best resolve the issue that this signer keyid may differ from the keyid of a signature generated by that signer. This is because gpg may use a subkey of the key identified by the passed keyid and we assign the subkey keyid to the signature, so that we know, at verification time, which key or subkey of the public key bundle we should use.

I have a feeling that the signer API should just not support the subkey-masterkey relationship. It adds, for no benefit, a PKI hierarchy to tuf/in-toto, which already have their own PKI hierarchies.

**This solution**  entails the following changes

- On public key import, users must pass the exact keyid of the key, they want to use for verification. This means, if gpg will use a subkey of a key bundle for signing, they must pass that subkey keyid.
- Public key import only returns the key identified by the keyid, and not the full public key bundle ([as it does now even if a subkey keyid is passed](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/gpg/common.py#L628-L632)).
- Signing fails, if gpg uses a different key for signing, than the one attached to the signer public key (e.g. because the user explicitly imported a master key, but the signing capability in gpg is set on a subkey).

caveat: Explicitly identifying a public key before signing requires additional tooling to allow signing with the default key, which is currently possible in in-toto.

An **alternative solution**, which allows both PKI hierarchies alongside each other, would be to accept public key bundles with subkeys, but always assign the master key keyid to the signature, regardless of what key/subkey was used. However, this would require a different way of matching the correct key in the public key bundle with a signature at verification time.

cc @jku,  @adityasaky, @SantiagoTorres

lukpueh:
One more minor thing I thought about is the `hashes` field of a GPGKey and whether we should merge the info with its `scheme` field.

For context:
- The field is currently [hardcoded](https://github.com/secure-systems-lab/securesystemslib/blob/fe86ea1e1a6f89cbb31b839925c81cdded2012df/securesystemslib/gpg/common.py#L147) to `["pgp+SHA2"]`, but ignored at signature [creation](https://github.com/secure-systems-lab/securesystemslib/blob/fe86ea1e1a6f89cbb31b839925c81cdded2012df/securesystemslib/gpg/constants.py#L79) and [verification](https://github.com/secure-systems-lab/securesystemslib/blob/fe86ea1e1a6f89cbb31b839925c81cdded2012df/securesystemslib/gpg/functions.py#L260), which both hard-code sha256 as digest algorithm.
- This PR already changes the gpg key format in a backwards-incompatible way, to make it in-toto/tuf specification compatible (see [#450](https://github.com/secure-systems-lab/securesystemslib/issues/450)). Metadata backwards-compatibility will be guaranteed through [in-toto/in-toto#534](https://github.com/in-toto/in-toto/issues/534).
-  [#308](https://github.com/secure-systems-lab/securesystemslib/issues/308) describes, among other things, how scheme can be used to encode the hash digest scheme. This is also what other supported keys in securesystemslib do. 


cc @adityasaky, @SantiagoTorres

lukpueh:
> would this change impact our notions of subkeys during verification?

Yes, both of my proposals are not compatible with current in-toto usage of GPG keys and signatures. in-toto currently considers all keys in a public key bundle (masterkey + subkeys), both, to check if a key is authorized via layout to provide a signature, and to identify the exact key or subkey in the bundle to verify the signature.

But this is not a concern, because the key and signature metadata formats of the GPGSigner API in securesystemslib are already incompatible with in-toto,. For in-toto-backwards- and securesystemslib-signer-compatibility, I proposed adding a custom _LegacyGPGSigner API to the in-toto code base  (see my [other comment](https://github.com/secure-systems-lab/securesystemslib/pull/488#issuecomment-1398195422) above).

lukpueh:
Another thought about parallel PKI: We export gpg keys with `creation_time` and `validity_period` fields from the gpg keychain, if present. This information is then considered at signature verification time and may raise `KeyExpirationError`.

Given that in-toto and TUF already have a mechanism to expire signing keys, that is via an expiration date in the delegating metadata, I wonder, if we should ignore expiration information of gpg keys, when verifying signatures, and leave it to the user to not add gpg keys to delegating metadata that would expire earlier than the metadata.

jku:
> I have a feeling that the signer API should just not support the subkey-masterkey relationship.

I agree with this, and both of your proposed solutions seem solid (although I admit I'm not fully immersed in the problem space of PGP subkeys).

> I wonder, if we should ignore expiration information of gpg keys

This is a reasonable point: increasing the complexity of the serialization format doesn't seem useful if the same practical security advantage can be had by making the key-import code just slightly more complex.

securesystemslib/signer/_gpg_signer.py
```diff
@@ -35,41 +110,123 @@ def from_priv_key_uri(
...
+        """
+        uri = f"{cls.SCHEME}:{homedir or ''}"
+
+        raw_key = gpg.export_pubkey(keyid, homedir)
+        raw_keys = [raw_key] + list(raw_key.pop("subkeys", {}).values())
+        keyids = []
+
+        for key in raw_keys:
+            if key["keyid"] == keyid:
+                # TODO: Raise here if key is expired, revoked, incapable, ...
```
lukpueh, line 190: I'll create a ticket for this. We can also partially re-use #190)

securesystemslib/signer/__init__.py
```diff
@@ -47,5 +48,8 @@
         ("rsa", "rsa-pkcs1v15-sha384"): SSlibKey,
         ("rsa", "rsa-pkcs1v15-sha512"): SSlibKey,
         ("sphincs", "sphincs-shake-128s"): SSlibKey,
+        ("rsa", "pgp+rsa-pkcsv1.5"): GPGKey,
+        ("dsa", "pgp+dsa-fips-180-2"): GPGKey,
+        ("eddsa", "pgp+eddsa-ed25519"): GPGKey,
```
jku, line 53: are these the same keytype/scheme combos that were used before, but the serialization format is now slightly different? Would it make sense sense to use e.g. a different keyname?

I guess in-toto is the only existing user... so do what makes sense there.

jku (approved):
:clap: looks fine to me. Left a question about the keytype  but since pgp key compatibility is mostly a in-toto issue, I'll leave it to you

securesystemslib/signer/__init__.py
lukpueh: The thing is, these keytype/scheme combos are completely new. We did use the same values as type/method combos, but I think that's okay.

PR #493, opened 2023-01-11
Misc fixes in dependency definitions

### LINKED ISSUE #482: Specify optional dependencies per feature
We use optional dependencies (aka. [extras](https://peps.python.org/pep-0508/#extras)) to allow a default installation with limited capabilities (e.g. verify ed25519 signatures using a pure Python implementation), which is useful for securesystemslib dependents (e.g. python-tuf), that need to be installable without any native dependencies (e.g. vendored into pip).

The full or a specific larger feature-set (e.g. signing) can be installed by specifying those extras like so:
```
pip install securesystemslib[crypto, asn1, pykcs11] # for hsm signing support
```
Except for hiding version constraints this is hardly more useful than telling user to install like so
```
pip install cryptography asn1crypto PyKCS11
```

**Current behavior**:

Specified per library:
https://github.com/secure-systems-lab/securesystemslib/blob/e2a08b123476ff05d1c3163eb6eeb3c705754952/pyproject.toml#L45-L51

**Expected behavior**:
- figure out what "features(s)" `cryptography`, `pynacl` should be mapped to (or leave those as they are)
- specify per feature, 
```python
[project.optional-dependencies]
[...]
gcpkms = ["google-cloud-kms"]
pq = ["PySPX==0.5.0"]
hsm = ["asn1crypto", "PyKCS11", "cryptography"]
```

**Caveat:**
Specifying them per feature is prone to redundancy. E.g. `cryptography` is needed for an hsm signing, but also for a traditional securesystemslib signing and gpg singing.

lukpueh:
- fixes #482
- plus two unrelated fixes in dependency definitions 

See commit messages for details.


pyproject.toml
```diff
@@ -45,10 +45,9 @@ Issues = "https://github.com/secure-systems-lab/securesystemslib/issues"
 [project.optional-dependencies]
 crypto = ["cryptography>=37.0.0"]
 gcpkms = ["google-cloud-kms", "cryptography>=37.0.0"]
+hsm = ["asn1crypto", "cryptography>=37.0.0", "PyKCS11"]
 pynacl = ["pynacl>1.2.0"]
-PySPX = ["PySPX==0.5.0"]
-asn1 = ["asn1crypto"]
-pykcs11 = ["PyKCS11"]
+PySPX = ["PySPX>=0.5.0"]
```
lukpueh, line 50: @rugo, was there a specific reason for pinning this to 0.5.0. See 58857cafc7075ef9b8c190c16d932155bc79271c for why I changed this.

rugo: Not really, all good.

jku (approved):

PR #495, opened 2023-01-12
Prepare v0.26.0 release

### LINKED ISSUE #475: New release
I don't think there is a need to rush a release but:
* the next release contains API breaks: we should build a good changelog 
* we should make a conscious decision on what is a release blocker

For the changelog, I think the commit messages from python-tuf PR https://github.com/theupdateframework/python-tuf/pull/2165 are probably a good start for documenting the signer API changes

As release blocker proposals:
* #442,  #472 Signer implementations would be good to have to verify the signer API design
* #466 I don't think private key generation/import/storage needs to be fully implemented, but we should have the story ready
* #468 likewise, at least a high level description of "this is how exceptions should work in signer API, and this is the current situation" would be helpful

lukpueh:
- bump version in __init__.py
- add changelog entry

----

Closes #475

This release should not break `in-toto` or `python-tuf` implementations, based on maintainer judgement and running their respective test suites against this release.

It does, however, require two updates in in-toto tests ([patch available](https://github.com/in-toto/in-toto/pull/529)) and DataDog integrations, as [discussed](https://github.com/secure-systems-lab/securesystemslib/pull/437#pullrequestreview-1147072331) and [agreed upon ](https://github.com/secure-systems-lab/securesystemslib/pull/437#issuecomment-1284111257) with corresponding maintainers.  cc @trishankatdatadog, @ofek

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,29 @@
 # Changelog
+
+## securesystemslib v0.26.0
+
+### Added
+* Private key URI schemes for signer instantiation (#456)
+* Public key container class for signature verification (#456)
+* Post-quantum sphincs+ signing scheme (#427)
+* Hardware Security Module (HSM) signing (#472)
+* Google Cloud KMS signing (#442, #480)
+
+### Changed
+* Use pyproject.toml for build configuration (#253)
+* Use hatchling as build backend (#484)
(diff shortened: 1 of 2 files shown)


jku (approved):
:+1:

PR #500, opened 2023-02-02
Correct public_key annotation

MVrachev:

Correct public_key annotation in SSlibSigner.from_priv_key_uri() as of right now, the "public_key"  argument is defined as a "Key" type, but then on line 166 it's validated that it's actually of an "SSlibKey" type.

This can lead to confusion as to whether other "Key" types can be used as the "public_key" value.

### DIFF
--- a/securesystemslib/signer/_signer.py
+++ b/securesystemslib/signer/_signer.py
@@ -153,7 +153,7 @@ class SSlibSigner(Signer):
     def from_priv_key_uri(
         cls,
         priv_key_uri: str,
-        public_key: Key,
+        public_key: SSlibKey,
         secrets_handler: Optional[SecretsHandler] = None,
     ) -> "SSlibSigner":
         """Constructor for Signer to call


MVrachev:
It seems as if black has updated its requirements and many parts of the codes that I did not touch are affected.
If there is a pr in the works will wait.
I could have time to look at this, but I am not sure when.

jku:
Yup, black 23 changed empty line rules and securesystemslib does not pin linters

jku (commented):
Is mypy ok with this change?

If it is, then LGTM

MVrachev:
Well, check it locally and it seems `mypy` is not happy:
```
securesystemslib/signer/_signer.py:153: error: Argument 2 of "from_priv_key_uri" is incompatible with supertype "Signer"; supertype defines the argument type as "Key"  [override]
securesystemslib/signer/_signer.py:153: note: This violates the Liskov substitution principle
securesystemslib/signer/_signer.py:153: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```
It seems you cannot use a more precise type when you override a function in the child class.

jku:
That makes sense.  Also keeping it a Key actually makes sense in another way: Real application code that uses this API only knows it has a Key (because that's what Metadata contains) so it's going to call this method with a Key... requiring a runtime type check in the calling code would be just annoying. 

I think the API is roughly right, closing.

PR #502, opened 2023-02-09
Improve gpg subprocess timeout and stop using custom process module

lukpueh:

gpg interface function to sign and export public keys now have an optional subprocess timeout. This means, users no longer need to patch a global in order to configure this value. Moreover, the default timeout is now defined inside the gpg subpackage, and increased to 10.
--> The should reduce test failures on slow runners, and makes it more intuitive to override the default.

Internally the gpg subpackage now no longer uses the custom subprocess wrapper, which provides very little benefit over calling directly into stdlib `subprocess`. 
--> This is in preparation for deprecation of the process module

see commit message for details

### DIFF
--- a/securesystemslib/gpg/constants.py
+++ b/securesystemslib/gpg/constants.py
@@ -24,13 +24,20 @@ from securesystemslib import process
 
 log = logging.getLogger(__name__)
 
+GPG_TIMEOUT = 10
+
 
 @functools.lru_cache(maxsize=3)
 def is_available_gnupg(gnupg: str) -> bool:
     """Returns whether gnupg points to a gpg binary."""
     gpg_version_cmd = gnupg + " --version"
     try:
-        process.run(gpg_version_cmd, stdout=process.PIPE, stderr=process.PIPE)
+        process.run(
+            gpg_version_cmd,
+            stdout=process.PIPE,
+            stderr=process.PIPE,
+            timeout=GPG_TIMEOUT,
+        )
         return True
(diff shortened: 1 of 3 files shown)


jku (approved):
The process of building strings just to then split them with shlex seems a little unusual... but I can see how some of the option handling would need more refactoring if you wanted to fix that. LGTM.

lukpueh:
> The process of building strings just to then split them with shlex seems a little unusual... but I can see how some of the option handling would need more refactoring if you wanted to fix that.

Exactly. I'd have to refactor how `keyarg` and `homearg` are passed, e.g.:
https://github.com/secure-systems-lab/securesystemslib/blob/6199f462282ea65df06b2e37d003cccdbebd5dba/securesystemslib/gpg/constants.py#L87 

Also, I wasn't sure how we want to treat gpg commands specified in an environment variable. Maybe they include some custom flag, which would behave differently if we stopped using shlex. But maybe that's okay. 🤷  

I mentioned in the commit message that I'm just trying to replicate the original behavior. I can also add a comment in the code.

jku:
> I can also add a comment in the code.

Not strictly needed IMO, I think the option and env var handling issues should be pretty obvious to anyone who does try to improve this

lukpueh:
- Rebased w/o change on https://github.com/secure-systems-lab/securesystemslib/pull/504
- Fixed linter issues
- Applied Jussi's suggestion to use `capture_output` instead of stdout/stderr PIPEs

CI passes. Merging...

Issue #337, opened 2021-02-22
in-toto Github Actions: different builds for windows fail with the same errors


SolidifiedRay:
**Description of issue or feature request**:
As [issue #407](https://github.com/in-toto/in-toto/issues/407) suggests, I set up the Github Actions Workflow for in-toto in this [draft PR](https://github.com/in-toto/in-toto/pull/426).  However, each time I run the workflow on my [fork](https://github.com/SolidifiedRay/in-toto/tree/gh-actions), different builds for windows fail, and they all give the same errors:

```
======================================================================
ERROR: test_timeout_setting (tests.test_runlib.TestLinkCmdExecTimeoutSetting)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "d:\a\in-toto\in-toto\.tox\py\lib\site-packages\securesystemslib\process.py", line 220, in run_duplicate_streams
    raise subprocess.TimeoutExpired(cmd, timeout)
subprocess.TimeoutExpired: Command '['d:\\a\\in-toto\\in-toto\\.tox\\py\\scripts\\python.exe', '-c', 'while True: pass']' timed out after 0.1 seconds

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "D:\a\in-toto\in-toto\tests\test_runlib.py", line 463, in test_timeout_setting
    in_toto.runlib.execute_link([sys.executable, '-c', 'while True: pass'], True)
  File "D:\a\in-toto\in-toto\in_toto\runlib.py", line 362, in execute_link
    timeout=float(in_toto.settings.LINK_CMD_EXEC_TIMEOUT))
  File "d:\a\in-toto\in-toto\.tox\py\lib\site-packages\securesystemslib\process.py", line 230, in run_duplicate_streams
    os.remove(stdout_name)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpzhlq0vde'
```

Here is the link for the full test results: https://github.com/SolidifiedRay/in-toto/runs/1948989796?check_suite_focus=true

Not sure if it is relevant, but another person also encounter a [test failure](https://github.com/secure-systems-lab/securesystemslib/pull/316#issuecomment-777294641) in the same process module, and it is also related to the `run_duplicate_streams` function.

**Current behavior**:
Failing test: test_timeout_setting (tests.test_runlib.TestLinkCmdExecTimeoutSetting)

[...]

PR #505, opened 2023-02-10
Remove `securesystemslib.process` module

### LINKED ISSUE #345: Default timeout setting helper in process module does not work as intended
#202 added a helper function with the intention of deferring the evaluation of the default argument `SUBPROCESS_TIMEOUT`, used in two functions of the `process` module, until invocation of these functions. 
This should have made it possible to modify `SUBPROCESS_TIMEOUT` via the `settings` module, where it is defined originally, even after importing the `process` module.

Unfortunately, this does not work as intended, because the helper function and thus the default `SUBPROCESS_TIMEOUT` is still evaluated only once at import time.

**Note:** This issue is a side-effect of the non-intuitive usage pattern of expecting user interaction via patching globals in the `settings` module, which is documented in #219. IMHO it is okay to fix this issue together with #219.

**Current behavior**:
Default argument is evaluated at function import time.

**Expected behavior**:
Default argument is evaluated at function invocation time.

lukpueh:
This is a BREAKING CHANGE for in-toto, must coordinate release! (cc in-toto/in-toto#544)

---
This module was originally developed in in-toto and transferred to securesystemslib in #174, primarily as Py2/Py3-agnostic wrapper around stdlib's `subprocess.run`, to run the `gpg` command. In #502 we switched to using `subprocess.run` directly.

Another wrapper around `subprocess.run`, provided by the module, allows capturing standard streams AND write them to a terminal. It was developed as specific `in-toto-run` feature and does not need to be public API in sslib. in-toto/in-toto#544 moves the function back to in-toto.

closes #345,
transfers #337 (to in-toto)


lukpueh:
cc @trishankatdatadog, @ofek

trishankatdatadog:
> cc @trishankatdatadog, @ofek

Noted, thanks, shouldn't affect us unless we upgrade @yzhan289 @iliakur

adityasaky (approved):
Approving on behalf of in-toto/in-toto.

PR #522, opened 2023-02-28
signer: add basic sigstore signer and verifier

lukpueh:
Adds SigstoreSigner and SigstoreKey implementations to the securesystemslib signer interface. They can be used to create and verify signatures using an OIDC token and an identity/issuer pair respectively, as described in the TAP:
["Ephemeral identity verification using sigstore's Fulcio for TUF developer key management"](https://github.com/theupdateframework/taps/blob/3b0d8b108758266e36790b3a782b943d633ce585/candidate-fulcio-tap.md)

IMPORTANT NOTES:
- ~the created signature does not strictly follow the typing requirements of the interface~
- currently, only offline verification is supported

This patch also adds a GitHub Action workflow that runs a tailored test, using ambient credentials for signing.

--- 

*This should already work, but I'd like to fix a few things before I mark this "ready for review"*.
- [x] fix signature interface incompatibility
- [x] ~~fix circular imports and inline-import workaround~~ (local scope import seems like the right choice here)
- [x] make sure signer remains importable even if sigstore is not installed 
- [ ] implement from_priv_key_uri (maybe in follow-up PR)
- [x] ~make online/offline verification configurable (maybe in follow-up PR)~ (let's discuss this in the TAP)


securesystemslib/signer/_sigstore_signer.py
```diff
@@ -0,0 +1,156 @@
...
+        Raises:
+            Various errors from sigstore-python.
+
+        Returns:
+            Signature.
+
+            NOTE: The ``signature`` attribute of the returned object
+            contains the ``dict`` representation of a sigstore ``Bundle`.
+            This is incompatible with the TUF specification and the
+            ``Signature` interface, which expect the attribute to be of type `str`.
```
jku, line 147: Could store the bundle in a separate "bundle" field in the signature, nothing forces us to use "sig", right?

To keep spec conformance at 100% could even fill "sig" with the actual signature value -- even if it's useless without the bundle

securesystemslib/signer/_sigstore_signer.py
```diff
@@ -0,0 +1,156 @@
...
+            logger.info("Key %s failed to verify sig: %s", self.keyid, str(e))
+            raise exceptions.VerificationError(
+                f"Unknown failure to verify signature by {self.keyid}"
+            ) from e
+
+
+class SigstoreSigner(Signer):
+    """Sigstore signer."""
+
+    def __init__(self, token: str, public_key: Key):
```
lukpueh, line 116: Requiring the token to be created beforehand is not very nice UX. The interface here should look more like the sigstore sign cli, where the user can pass identity/issuer, or instruct it to use ambient credentials.

jku: Yeah makes sense -- I guess the most important bit is how it works with from_private_key_uri(). The ways to deliver the "arguments" are then
* signing key specific details included in private_key_uri
* signing system variables can also be "ambient" (so environment variables etc)

lukpueh: You are right, the preferred signer creation UI is `from_private_key_uri`. So creating the token there and passing it to the constructor should work.

lukpueh:
Okay, I think this is mergable as self-contained feature. To be a fully usable signer, we need to add from_priv_key_uri, but I suggest to do that in a follow-up PR.

The "on: pull_request" test-sigstore job should start working once this PR is merged. A successful run is already available in my fork: https://github.com/lukpueh/securesystemslib/actions/runs/4355175849/jobs/7611437818

securesystemslib/signer/_sigstore_signer.py
```diff
@@ -0,0 +1,167 @@
...
+            result = verifier.verify(materials, identity)
+            if not result:
+                logger.info(
+                    "Key %s failed to verify sig: %s", self.keyid, result.reason
+                )
+                raise UnverifiedSignatureError(
+                    f"Failed to verify signature by {self.keyid}"
+                )
+        except UnverifiedSignatureError:
+            raise
```
jku, line 109: Nit: you could probably handle `result` outside of the try block to avoid this raise-except-raise pattern.

tox.ini
```diff
@@ -45,6 +45,19 @@ passenv =
...
     python -m tests.check_kms_signers
 
+[testenv:sigstore]
+deps =
+    -r{toxinidir}/requirements-pinned.txt
+    -r{toxinidir}/requirements-sigstore.txt
+passenv =
+    GITHUB_ACTIONS
+    ACTIONS_ID_TOKEN_REQUEST_TOKEN
+    ACTIONS_ID_TOKEN_REQUEST_URL
```
jku, line 55: these are used by ambient credentials detection?

securesystemslib/signer/_sigstore_signer.py
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+    def verify_signature(self, signature: Signature, data: bytes) -> None:
+        # pylint: disable=import-outside-toplevel,import-error
+        try:
+            from sigstore.verify import VerificationMaterials, Verifier
+            from sigstore.verify.policy import Identity
+            from sigstore_protobuf_specs.dev.sigstore.bundle.v1 import Bundle
+
+            verifier = Verifier.production()
```
jku, line 92: seeing this code it seems clear that the content in our key should include everything needed to choose the sigstore "instance".

That said, it's quite unclear to me what that all means (rekor and tuf urls? fulcio?): this is really more a question for the TAP than this implementation which should focus on supporting the default instance for now

So no action requested here.

jku (approved):
Looks good.

I think it's likely that the key format is not final but I think we should get this merged and just keep it "unstable": 
* TAP really needs an implementation so this serves that need 
* the implementation should work fine already as long as you don't want anything else than the public good instance

Thinking about the future changes: we can just add keytypes/schemes if we want to support this implementation and a newer one at some point

jku:
Should the module or the classes be marked "unstable API" or something?

.github/workflows/test-sigstore.yml
```diff
@@ -0,0 +1,38 @@
+name: Run Sigstore Signer tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
jku, line 7: should remove this as it will unfortunately keep failing for PRs

We should also maybe file an issue for managing these tests that can be only run on upstream branches... so that an issue would be raised when a test fails

jku: oh, I actually did that in kms-tests: you can probably copy the file-ticket-on-failure step from there

lukpueh: > should remove this as it will unfortunately keep failing for PRs

Is that the case? Shouldn't it work, once the workflow is merged?

jku: I don't think so: pull requests have read only permissions and you need `id-token: write`

or in more OIDC terms, pull requests aren't allowed to "authenicate" as the project (even though sometimes it would make sense, since the authentication is not for the whole project but for the specific workflow which could be dedicated for PRs...)

jku: but feel free to test by merging: we can fix with another PR if it does not work

securesystemslib/signer/_sigstore_signer.py
lukpueh: Good point! Can you check if the commit I just pushed resolves this to your liking.

jku: yeah that looks good to me

.github/workflows/test-sigstore.yml
lukpueh: @jku: Looks like this actually works:
https://github.com/secure-systems-lab/securesystemslib/actions/runs/4382839569

Build was triggered via pull request and apparently had the required permissions.

Or maybe it only worked because the pr didn't come from a fork. 🤷

jku: > maybe it only worked because the pr didn't come from a fork. shrug

yes, correct: I expect branches in upstream repo to work, but branches in forks  to fail

PR #523, opened 2023-03-02
Adopt default branch change master to main

### LINKED ISSUE #515: Rename default branch to `main`
Context and guidance: https://github.com/github/renaming

**Current behavior**:
Default branch is `master`

**Expected behavior**:
Default branch is `main`

lukpueh:
Closes #515 

- Update ci badge link in README
- Update branch restriction in ci push event listener
- Update log statement in check-upstream-ed25519.yml workflow (they have also changed their default branch name)


jku (approved):
LGTM. There might be other places that refere to master but
1. that probably does not even break things -- github is pretty good at this
2. it won't be critical, this is main branch after all, not a release branch

PR #526, opened 2023-03-03
Hsm identification

jku:
I bought a second yubikey so need a way to tell them apart... so: Define a URI that contains identifying information for the HSM token:
 * URI path is the keyid
 * URI query can contain any token field filters (this is a bit overkill but since there's apparently no standards for this it seemed reasonable)

Example URI: `hsm:2?label=YubiKey+PIV+%2315835999`

This would use keyid 2 from a token with label "YubiKey PIV #15835999". This example is also what gets created by HSMSigner.import_() by default.

import_() stays backwards compatible, and old URIs keep working but there are some changes in functionality:
* Running import_() now fails if there are more than 1 tokens (but a filter can be provided there as well). Because of this the tests needed some changes (softhsm creates a new token in InitToken???) -- unfortunately this means the default import_() filter is not tested as I couldn't figure out how to remove the extra softhsm token.
* The constructor has a new required argument (this could be fixed but I didn't see it as that important).

I can be persuaded to do these differently.

### DIFF
--- a/securesystemslib/signer/_hsm_signer.py
+++ b/securesystemslib/signer/_hsm_signer.py
@@ -96,6 +96,7 @@ class HSMSigner(Signer):
       "YubiKey+PIV+%2315835999"
 
     Usage::
+
         # Store public key and URI for your HSM device for later use. By default
         # slot 9c is selected.
         uri, pubkey = HSMSigner.import_()
@@ -150,13 +151,10 @@ class HSMSigner(Signer):
         self.pin_handler = pin_handler
 
     @staticmethod
-    @contextmanager
-    def _get_session(filters: Dict[str, str]) -> Iterator["PyKCS11.Session"]:
-        """Context manager to handle a HSM session.
+    def _find_pkcs_slot(filters: Dict[str, str]) -> int:
+        """Return the PKCS slot with initialized token that matches filter
 
-        The slot/token is selected by filtering by token info fields.
-        ValueError is raised if not matching slot/token is not found, or if
-        more than one are found.
+        Raises ValueError if more or less than 1 PKCS slot is found.
         """
         lib = PYKCS11LIB()
         slots: List[int] = []
@@ -178,13 +176,24 @@ class HSMSigner(Signer):
 
         if len(slots) != 1:
             raise ValueError(
-                f"Found {len(slots)} slots/tokens matching filter {filters}"
+                f"Found {len(slots)} cryptographic tokens matching filter {filters}"
             )
 
-        session = lib.openSession(slots[0])
+        return slots[0]
+
+    @staticmethod
+    @contextmanager
+    def _get_session(filters: Dict[str, str]) -> Iterator["PyKCS11.Session"]:
+        """Context manager to handle a HSM session.
+
+        The cryptographic token is selected by filtering by token info fields.
+        ValueError is raised if matching token is not found, or if more
+        than one are found.
+        """
+        slot = HSMSigner._find_pkcs_slot(filters)
+        session = PYKCS11LIB().openSession(slot)
         try:
             yield session
-
         finally:
             session.closeSession()
 
@@ -229,26 +238,18 @@ class HSMSigner(Signer):
 
     @classmethod
     def _build_token_filter(cls) -> Dict[str, str]:
-        """Builds a token filter for the found slot/token.
+        """Builds a token filter for the found cryptographic token.
 
         The filter will include 'label' if one is found on token.
 
-        raises ValueError if less or more than 1 token/slot is found
+        raises ValueError if less or more than 1 token is found
         """
 
         lib = PYKCS11LIB()
-        slots: List[int] = []
-        for slot in lib.getSlotList(tokenPresent=True):
-            tokeninfo = lib.getTokenInfo(slot)
-            if not tokeninfo.flags & PyKCS11.CKF_TOKEN_INITIALIZED:
-                # useful for tests (softhsm always has an unitialized token)
-                continue
-            slots.append(slot)
+        slot = cls._find_pkcs_slot({})
+        tokeninfo = lib.getTokenInfo(slot)
 
-        if len(slots) != 1:
-            raise ValueError(f"Expected 1 token/slot, found {len(slots)}")
         filters = {}
-        tokeninfo = lib.getTokenInfo(slots[0])
         # other possible fields include manufacturerID, model and serialNumber
         for key in ["label"]:
             try:
@@ -266,8 +267,8 @@ class HSMSigner(Signer):
     ) -> Tuple[str, SSlibKey]:
         """Import public key and signer details from HSM.
 
-        Either only one token/slot must be present when importing or a
-        token_filter must be provided.
+        Either only one cryptographic token must be present when importing or a
+        token_filter that matches a single token must be provided.
 
         Returns a private key URI (for Signer.from_priv_key_uri()) and a public
         key. import_() should be called once and the returned URI and public
@@ -276,15 +277,14 @@ class HSMSigner(Signer):
         Arguments:
             hsm_keyid: Key identifier on the token. Default is 2 (meaning PIV key slot 9c).
             token_filter: Dictionary of token field names and values used to
-                filter the correct slot/token. If no filter is provided one is built
-                from the fields found on the token.
+                filter the correct cryptographic token. If no filter is
+                provided one is built from the fields found on the token.
 
         Raises:
             UnsupportedLibraryError: ``PyKCS11``, ``cryptography`` or ``asn1crypto``
                     libraries not found.
-            ValueError: No compatible key for ``hsm_keyid`` found on HSM.
+            ValueError: A matching HSM device could not be found.
             PyKCS11.PyKCS11Error: Various HSM communication errors.
-
         """
         if CRYPTO_IMPORT_ERROR:
             raise UnsupportedLibraryError(CRYPTO_IMPORT_ERROR)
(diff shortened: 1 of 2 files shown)


jku:
Two main questions:
* is filtering tokens with a dictionary overkill and should I have just hard coded filtering on `label` for now?
* is the "API break" ok or should I try to keep the constructor compatible (my reasoning was that the real API is the private key URL and import_() which do stay compatible)

securesystemslib/signer/_hsm_signer.py
```diff
@@ -192,15 +225,49 @@ def _find_key_values(
...
+
         Returns a private key URI (for Signer.from_priv_key_uri()) and a public
         key. import_() should be called once and the returned URI and public
         key should be stored for later use.
 
         Arguments:
             hsm_keyid: Key identifier on the token. Default is 2 (meaning PIV key slot 9c).
+            token_filter: Dictionary of token field names and values used to
+                filter the correct slot/token. If no filter is provided one is built
+                from the fields found on the token.
```
lukpueh, line 270: > If no filter is provided one is built from the fields found on the token.

Why? Can't we just not filter, if no filter is provided?

tests/test_hsm_signer.py
```diff
@@ -114,6 +114,11 @@ def setUpClass(cls):
         slot = lib.getSlotList(tokenPresent=True)[0]
         lib.initToken(slot, so_pin, token_label)
 
+        # suddenly there are two slots (why?), use label to filter in tests
+        tokeninfo = lib.getTokenInfo(slot)
+        cls.token_filter = {"label": getattr(tokeninfo, "label")}
+        print(cls.token_filter)
```
lukpueh, line 120: Did you mean to leave the `print` here?

tests/test_hsm_signer.py
```diff
@@ -114,6 +114,11 @@ def setUpClass(cls):
         slot = lib.getSlotList(tokenPresent=True)[0]
         lib.initToken(slot, so_pin, token_label)
 
+        # suddenly there are two slots (why?), use label to filter in tests
```
lukpueh, line 117: > suddenly there are two slots (why?)

I don't quite understand this comment here. I do have an idea, but only because I read the commit message too. Please rephrase or remove.

lukpueh (changes requested):
Cool stuff! 

I think the filter dictionary is quite neat and doesn't require that much more code compared to a hardcoded a label filter. Plus, your argument that there is no one obvious way to filter tokens makes sense. So even if it's a bit over kill, I'm fine with it.

But I don't understand the purpose of `_build_token_filter`. Couldn't you just not filter, if no filter is provided?

Could you also re-explain the issue you had with suddenly(?) SoftHSM creating two slots? And, why that doesn't let you test the default case?

tests/test_hsm_signer.py
jku: no

securesystemslib/signer/_hsm_signer.py
jku: sure we could but my assumption is that filtering on "label" is the best default:
* it works fine if you have a single HW key (whether it has a label or not)
* when you buy a second HW key, things keep working and the signer URIS (that you have already created before) are still correctly identified if they have labels

If that assumption turns out false we can change the default to no filter.

jku:
> Could you also re-explain the issue you had with suddenly(?) SoftHSM creating two slots? And, why that doesn't let you test the default case?

What I mean is that before initToken() there is a slot 0 already with some default token I suppose. Then when we call `lib.initToken(0,...)` that slot 0 contains our testing token. Unfortunately a new slot 1 also appears with some default content.

Since I'm now trying to be more strict about finding a specific key during `import_()`, I can't use the normal import_() style in tests (this is what would work in normal applications):
```python
uri, key = HSMSigner.import_()
```
because that now fails since there are two slots and two tokens that could be used and I think the safe option is to fail in that case. So to work around I have to do 
```python
uri, key = HSMSigner.import_(self.hsm_keyid_default, self.token_filter)
```
which does not test the default.

lukpueh:
Okay, thanks for the explanation.

jku:
oops sorry for editing your comment by accident

I tried to rply to this: 

> I'm fine with ticketizing this issue, unless you have resources to debug SoftHSM.

This turns out to be a softhsm feature (that allows you to "create" all the new tokens you want without a separate API): the way to distinguish the uninitialized token from usable tokens is PyKCS11.CKF_TOKEN_INITIALIZED flag which we could check for. I'll try this.

securesystemslib/signer/_hsm_signer.py
```diff
@@ -83,26 +83,34 @@ class HSMSigner(Signer):
...
+    FILTERS are optional. Example URIs:
     * "hsm:":
-      Sign with key on PIV digital signature slot 9c.
+      Sign with a key with default keyid 2 (PIV digital signature slot 9c) on the
+      only token/smartcard available.
+    * "hsm:2?label=YubiKey+PIV+%2315835999":
+      Sign with key with keyid 2 (PIV slot 9c) on a token with label
+      "YubiKey+PIV+%2315835999"
 
     Usage::
```
lukpueh, line 98: nit: IIRC sphinx needs a blank line here, to render this as code block. This isn't really relevant yet, as we don't generate docs, but since I have another comment below, I might as well point this out. :)

securesystemslib/signer/_hsm_signer.py
```diff
@@ -133,17 +145,41 @@ def __init__(
...
 
     @staticmethod
     @contextmanager
-    def _default_session():
-        """Context manager to handle default HSM session on reader slot 1."""
+    def _get_session(filters: Dict[str, str]) -> Iterator["PyKCS11.Session"]:
+        """Context manager to handle a HSM session.
+
+        The slot/token is selected by filtering by token info fields.
+        ValueError is raised if not matching slot/token is not found, or if
```
lukpueh, line 158: double negation?

securesystemslib/signer/_hsm_signer.py
```diff
@@ -192,15 +228,56 @@ def _find_key_values(
         return ECDomainParameters.load(bytes(params)), bytes(point)
 
     @classmethod
-    def import_(cls, hsm_keyid: Optional[int] = None) -> Tuple[str, SSlibKey]:
+    def _build_token_filter(cls) -> Dict[str, str]:
+        """Builds a token filter for the found slot/token.
+
+        The filter will include 'label' if one is found on token.
+
+        raises ValueError if less or more than 1 token/slot is found
```
lukpueh, line 236: Is it worth disambiguating the terms "slot" and "token" somewhere in the HSM signer documentation? I must admit, I'm still confused. Also because they mean different things in related standards (pkcs11 and piv).

securesystemslib/signer/_hsm_signer.py
```diff
@@ -192,15 +228,56 @@ def _find_key_values(
...
+    @classmethod
+    def import_(
+        cls,
+        hsm_keyid: Optional[int] = None,
+        token_filter: Optional[Dict[str, str]] = None,
+    ) -> Tuple[str, SSlibKey]:
         """Import public key and signer details from HSM.
 
+        Either only one token/slot must be present when importing or a
+        token_filter must be provided.
```
lukpueh, line 270: nit: Even if a token_filter is provided, it could match multiple tokens. IMO the important thing to convey here is that the function must unambiguously identify a single token/slot.

securesystemslib/signer/_hsm_signer.py
```diff
@@ -192,15 +228,56 @@ def _find_key_values(
...
+
         Returns a private key URI (for Signer.from_priv_key_uri()) and a public
         key. import_() should be called once and the returned URI and public
         key should be stored for later use.
 
         Arguments:
             hsm_keyid: Key identifier on the token. Default is 2 (meaning PIV key slot 9c).
+            token_filter: Dictionary of token field names and values used to
+                filter the correct slot/token. If no filter is provided one is built
+                from the fields found on the token.
```
lukpueh, line 280: Only "label" at the moment, right? But maybe that's not important to mention.

lukpueh (approved):
Very nice enhancement of the hsm signer!  I'll approve despite a few minor nits. Although the double negation in the docstring seems like a mistake and should probably be fixed. The rest is nice to have.

securesystemslib/signer/_hsm_signer.py
jku: I guess I could remove any mention of slots. 

The slot I mean here is the PKCS11 slot, aka the smart card reader: since we don't filter readers I don't think we even need to mention them

securesystemslib/signer/_hsm_signer.py
```diff
@@ -192,15 +228,56 @@ def _find_key_values(
...
+        """
+
+        lib = PYKCS11LIB()
+        slots: List[int] = []
+        for slot in lib.getSlotList(tokenPresent=True):
+            tokeninfo = lib.getTokenInfo(slot)
+            if not tokeninfo.flags & PyKCS11.CKF_TOKEN_INITIALIZED:
+                # useful for tests (softhsm always has an unitialized token)
+                continue
+            slots.append(slot)
```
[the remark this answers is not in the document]
jku: yeah maybe it does

securesystemslib/signer/_hsm_signer.py
jku: yeah only label is currently added: I'm not mentioning that in docstring

jku:
Refactoring suggestion was good, I even expanded that a bit. There is now a `_find_pkcs_slot(filters: dict[str, str]) -> int` and that makes the calling code much easier to reason about. 

The failing test is Windows struggling with gpg -- I'm merging this

PR #528, opened 2023-03-06
tests: rm HSM availability assertion in tox.ini

### LINKED ISSUE #527: tests should not fail by default (PYKCS11LIB env var)
Currently if you install securesystemslib and run the tests as instructed (`tox` or `tox -e py`), you  get:

```
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.10/os.py", line 680, in __getitem__
    raise KeyError(key) from None
KeyError: 'PYKCS11LIB'
```

This is wrong and likely really hard to figure out for a new developer
* CI should keep running with HSM tests enabled
* By default developer testing must not include HSM tests if the testing setup can't be made fool proof (AFAICT it can't)

lukpueh:
Currently, tox asserts that PYKCS11LIB env var is set to not silently skip tests on SoftHSM (virtual HSM).

This is is unnecessary for CI, where we control the env var, and also hard to figure out locally, where we can't expect everyone to setup SoftHSM.

This patch removes the assertion and adds a note to ci.yml to point out the consequences of unsetting the env var.

Fixes #527

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,8 @@ jobs:
       - name:  Install system dependencies
         shell: bash
         run: |
+          # NOTE: HSM tests are skipped silently, if PYKCS11LIB is unset.
+
           if [ "$RUNNER_OS" == "Linux" ]; then
             sudo apt-get install -y softhsm2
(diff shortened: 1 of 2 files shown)


jku (approved):
seems reasonable to me

Issue #533, opened 2023-03-13
SigstoreSigner: define URI and some import functionality


jku:
NOTE: SigstoreSigner is unstable and will likely stay that way for a bit.... It's already very interesting though so would be nice to to start implementing the quality of life improvements:
* signer should define a private key URI. As far as I can tell it should only encode whether the signing should use ambient credentials or not -- other suggestions are welcome though. So maybe just `sigstore:?ambient=true` or `sigstore:?ambient=false`
* from_priv_key_uri() would either use `detect_credentials()` or `issuer.identity_token()` based on ambient value
* the generic import method would take `identity`, `issuer` and `ambient` as arguments and return a public key (with a generated keyid) and URI  

Further improvements:
* later down the line we may want to provided more specific helpers like e.g. `import_github_action()` that could take repository, workflow and branch name instead (because constructing the identity string is a bit complex)
* signing really should be able to choose the identity and issuer -- user should not be able to choose GitHub as the issuer if we know we want issuer to be Google: user should only decide to authenticate or not.

lukpueh:
Mostly addressed by #535. Further improvements ticketized in #536 and #537

Issue #520, opened 2023-02-26
Windows CI keeps failing (flaky)


jku:
At least three failures since the CI was enabled on windows, all look roughly like this:
```
y: commands[1]> python -c "import os; os.environ[\"PYKCS11LIB\"]"
py: commands[2]> coverage run tests/aggregate_tests.py
................................Exporting master key '8465a1e2e0fb2b40adb2478e18fb3f537e0c8a17' including subkeys '6a112fd3390b2e53afc2e57f8fc8e12099aeceea, c5a0abe6ec19d0d65f85e2c39be9df5131d924e9' for passed keyid 'C5A0ABE6EC19D0D65F85E2C39BE9DF5131D924E9'.
..........................E........................s..........................Exporting master key '8465a1e2e0fb2b40adb2478e18fb3f537e0c8a17' including subkeys '6a112fd3390b2e53afc2e57f8fc8e12099aeceea, c5a0abe6ec19d0d65f85e2c39be9df5131d924e9' for passed keyid 'C5A0ABE6EC19D0D65F85E2C39BE9DF5131D924E9'.
...................s............
LoadLibrary() failed with error 126: The specified module could not be found.

======================================================================
ERROR: setUpClass (tests.test_hsm_signer.TestHSM)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\a\securesystemslib\securesystemslib\tests\test_hsm_signer.py", line 113, in setUpClass
    lib = PYKCS11LIB()
  File "D:\a\securesystemslib\securesystemslib\securesystemslib\signer\_hsm_signer.py", line 76, in PYKCS11LIB
    _PYKCS11LIB.load()
  File "D:\a\securesystemslib\securesystemslib\.tox\py\lib\site-packages\PyKCS11\__init__.py", line 488, in load
    raise PyKCS11Error(rv, pkcs11dll_filename)
PyKCS11.PyKCS11Error: Load (C:\SoftHSM2\lib\softhsm2-x64.dll)
```

* "LoadLibrary() failed" line seems to jump around so this could be a timing issue.
* happened on at least python 3.7 and 3.10

https://github.com/secure-systems-lab/securesystemslib/actions/runs/4240908943/jobs/7370472661

lukpueh:
Thanks for reporting!

> happened on at least python 3.7 and 3.10

I've seen it on 3.9 as well.

lukpueh:
[...]

PR #535, opened 2023-03-14
signer: add SigstoreSigner from_priv_key_uri and import_ methods

lukpueh:
addresses most of #533 

- `from_priv_key_uri`supports token detection from ambient credentials (default), and token creation from OAuth2 + OpenID browser login (with '?ambient=false' URI param).
- `import_` creates  sigstoreKey from passed identity and issuer, and signer URI based on passed bool to toggle ambient credential usage.

TODOs (not necessarily in this PR)
- [ ] figure out appropriate CI trigger conditions
- [ ] add tests
- [ ] use identity/issuer from public key, passed to `from_priv_key_uri`, in the OAuth2 + OpenID flow
- [ ] validate if public key and token passed to SigstoreSigner's from_priv_key_uri and constructor match
- [ ] [#533](https://github.com/secure-systems-lab/securesystemslib/issues/533) also suggests adding an `SigstoreSigner.import_github_action()`

### DIFF
--- a/tests/check_sigstore_signer.py
+++ b/tests/check_sigstore_signer.py
@@ -9,18 +9,9 @@ import os
 import unittest
 
 from securesystemslib.signer import (
-    KEY_FOR_TYPE_AND_SCHEME,
     SIGNER_FOR_URI_SCHEME,
-    Key,
     Signer,
-    SigstoreKey,
     SigstoreSigner,
-)
-
-KEY_FOR_TYPE_AND_SCHEME.update(
-    {
-        ("sigstore-oidc", "Fulcio"): SigstoreKey,
-    }
 )
 
 SIGNER_FOR_URI_SCHEME.update({SigstoreSigner.SCHEME: SigstoreSigner})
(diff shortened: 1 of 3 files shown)


lukpueh:
A passing ci run is available in my fork: https://github.com/lukpueh/securesystemslib/actions/runs/4417013232/jobs/7742109165

tests/check_sigstore_signer.py
```diff
@@ -8,20 +8,13 @@
...
+    Signer,
     SigstoreSigner,
 )
 
-KEY_FOR_TYPE_AND_SCHEME.update(
-    {
-        ("sigstore-oidc", "Fulcio"): SigstoreKey,
-    }
-)
+SIGNER_FOR_URI_SCHEME.update({SigstoreSigner.SCHEME: SigstoreSigner})
```
jku, line 17: very minor nit: this might be a bit simpler for updating a single key
```suggestion
SIGNER_FOR_URI_SCHEME[SigstoreSigner.SCHEME] = SigstoreSigner
```

jku (approved):
Looks great! Comments below should not block merging this IMO.

> * figure out appropriate CI trigger conditions

yes, pull_request trigger has to go I think.

> * use identity/issuer from public key, passed to from_priv_key_uri, in the OAuth2 + OpenID flow
> * validate if public key and token passed to SigstoreSigner's from_priv_key_uri and constructor match

I've asked about this in a couple of channels, let's see what happens




 > * ... adding a SigstoreSigner.import_github_action()
 
no need to rush with this (since maybe there are alternatives like  just providing some useful constants) but this might be useful -- producing the github action identity string is not completely trivial, and requiring the issuer string is just making the user do work we could avoid.

lukpueh:
Thanks for the review. I accepted your inline suggestion, disabled "on pr", and created separate issues for the remaining TODOs. CI failure is unrelated (#520). Merging...

PR #542, opened 2023-03-20
ci/tests: workarounds for flaky windows tests

lukpueh:
Works around issues that would frequently fail Windows CI as documented in #520, by:

- disabling hsm tests on Windows,
- generally increaseing gpg subprocess timeout in tests


lukpueh:
The failing test is an unrelated platform-independently flaky rsa test 😢 It is documented in #462.

jku (approved):
:+1: let's try this

PR #543, opened 2023-03-21
gpg: set default in function to allow global patch

lukpueh:
Patching globals to modify function behavior is an anti-pattern and should not be recommended to a user!!

But it seems acceptable in tests (see #542). For this to work, the global must not be used before being modified, which, AFAIK, is not possible, if the global is defined and used (e.g. in a function signature) in the same scope.

This commit changes the relevant function to access the default value in local scope, so that it global can be imported and modified before it.

More invasive but probably more correct solutions would be:
- move either GPG_TIMEOUT or is_available_gnupg to a different module, or
- call relevant functions in tests with an explicit timeout arg

### DIFF
--- a/securesystemslib/gpg/constants.py
+++ b/securesystemslib/gpg/constants.py
@@ -20,7 +20,7 @@ import logging
 import os
 import shlex
 import subprocess  # nosec
-from typing import List
+from typing import List, Optional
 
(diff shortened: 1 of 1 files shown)


jku (approved):
This seems correct even if the other improvements are later planned

Issue #512, opened 2023-02-14
Drop md5 and sha1 schemes from supported key schemes


jku:
Now that we have a key dispatch mechanism we could drop some ancient schemes: Let's remove all md5 and sha1 using schemes from KEY_FOR_TYPE_AND_SCHEME. This means:
* Keys using those schemes will fail to be recognised by default
* But since KEY_FOR_TYPE_AND_SCHEME is modifiable, applications can override decisions like this (as long as the underlying libraries keep working)

CC @MVrachev

PR #554, opened 2023-03-30
Drop md5 and sha1 schemes from supported key schemes

ckilcoin:
Fixes: #512 

Tested by running tox -e py311, no additional tests needed since it is a removal of schemes.


jku:
Quick comment: 
* I suggested removing the support from signer module  (so the signer/__init__.py key dispatch table)
* this PR also removes support from keys and hashes modules

This is probably still fine: nobody should be using these hash algorithms in the year 2023 (the theoretical danger is that _existing_ keys used somewhere could stop working because of this... but I do think it's theoretical)
 
 @lukpueh opinions?

jku:
Also, please have a look at the lint results (you can check them locally too with `tox -e lint`)

jku (approved):
LGTM

PR #557, opened 2023-03-31
Move _get_keyid helper to Signer base class and use in implementations

lukpueh:
Signer base class now provides a static `_get_keyid` method which may be used in the `import_` or `generate` methods of Signer implementations.

This PR also changes GCPSigner in HSMSigner to use the new method instead of the legacy `_get_keyid`, which does overly complicated and intransparent mangling of the input data, and thus produces different hashes.

Note that keyid computation is not public API, and verification of existing signatures with existing keys is not affected. If this does disrupt existing users, who expect a certain default keyid for a give key, I'd rather annoy them now, when the signer API is still experimental.

### DIFF
--- a/securesystemslib/signer/_gcp_signer.py
+++ b/securesystemslib/signer/_gcp_signer.py
@@ -6,7 +6,6 @@ from urllib import parse
 
 import securesystemslib.hash as sslib_hash
 from securesystemslib import exceptions
-from securesystemslib.keys import _get_keyid
 from securesystemslib.signer._key import Key
 from securesystemslib.signer._signer import (
     SecretsHandler,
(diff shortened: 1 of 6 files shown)


jku (approved):
LGTM. 

I still hope we can move to free form keyids in the future -- but even in that case it probably makes sense to have a "default" that should just work for the cases where the application is not interested in doing the extra work of uniquely naming the keys.

lukpueh:
> I still hope we can move to free form keyids in the future -- but even in that case it probably makes sense to have a "default" that should just work for the cases where the application is not interested in doing the extra work of uniquely naming the keys.

Agreed. And I think the signer API already allows you to use free form keyids.

PR #558, opened 2023-03-31
signer: _from/_to_dict helpers for Key base fields

lukpueh:
Add de/serialization helpers for common fields in the Key base class, which may be used in subclass from/to_dict methods for convenience.

* _to_dict: This could be Key.to_dict, which subclasses default to, but we prefer that subclasses must implement a to_dict, to avoid unnoticed serialization errors.

* _from_dict: Other than from_dict, this one does not create a Key instance, but just returns the popped off fields. This makes usage in subclasses more flexible and less mystical (lessons learned from https://github.com/theupdateframework/python-tuf/commit/ace25e4ad3d731a66171f7b9e6978384e4629efb).

### DIFF
--- a/securesystemslib/signer/_gpg_signer.py
+++ b/securesystemslib/signer/_gpg_signer.py
@@ -32,19 +32,11 @@ class GPGKey(Key):
 
     @classmethod
     def from_dict(cls, keyid: str, key_dict: Dict[str, Any]) -> "GPGKey":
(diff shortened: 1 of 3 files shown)


jku (approved):
Method naming is a bit suspicious maybe... but I've tried to come up with better suggestions for a week and came up with nothing so LGTM.

PR #565, opened 2023-04-13
Export Base64 signatures in Envelope

PradyumnaKrishna:

As per DSSE spec, Signatures in Envelope should be base64 encoded. While securesystemslib implementation uses hex signatures for signing and verifying them. Best way to achieve this by using base64 encoded signatures externally and not changing any internal representation of signature.

This PR adds two helper method that encodes and converts signature into base64 dict and vice versa. Testcases are added and updated as well.

Reference: https://github.com/in-toto/in-toto/issues/563#issuecomment-1487815979

### DIFF
--- a/securesystemslib/dsse.py
+++ b/securesystemslib/dsse.py
@@ -58,22 +58,26 @@ class Envelope:
         payload = b64dec(data["payload"])
         payload_type = data["payloadType"]
 
-        signatures = [
-            Signature.from_base64_dict(signature)
-            for signature in data["signatures"]
-        ]
+        signatures = []
+        for signature in data["signatures"]:
+            signature["sig"] = b64dec(signature["sig"]).decode("utf-8")
+            signatures.append(Signature.from_dict(signature))
 
         return cls(payload, payload_type, signatures)
 
     def to_dict(self) -> dict:
         """Returns the JSON-serializable dictionary representation of self."""
 
+        signatures = []
+        for signature in self.signatures:
+            sig_dict = signature.to_dict()
+            sig_dict["sig"] = b64enc(sig_dict["sig"].encode("utf-8"))
+            signatures.append(sig_dict)
+
         return {
             "payload": b64enc(self.payload),
             "payloadType": self.payload_type,
-            "signatures": [
-                signature.to_base64_dict() for signature in self.signatures
-            ],
+            "signatures": signatures,
         }
 
     def pae(self) -> bytes:
--- a/securesystemslib/dsse.py
+++ b/securesystemslib/dsse.py
@@ -59,7 +59,8 @@ class Envelope:
         payload_type = data["payloadType"]
 
         signatures = [
-            Signature.from_dict(signature) for signature in data["signatures"]
+            Signature.from_base64_dict(signature)
+            for signature in data["signatures"]
         ]
 
         return cls(payload, payload_type, signatures)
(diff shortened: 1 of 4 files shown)


lukpueh (commented):
Thanks for the great patch, @PradyumnaKrishna!

What do you think about adding the conversion logic as internal helpers to Envelope? We don't need it elsewhere, so why not keep the API smaller.

Also, apart from the docstring, which says "hex", the Signature class itself has been agnostic to the encoding of the sig field so far. I suggest we keep it that way.

jku:
> Best way to achieve this by using base64 encoded signatures externally and not changing any internal representation of signature.

Just bike shedding here but I suppose you could have a DSSESignature that is just a Signature with overridden `to_dict()` and `from_dict()` ?  I didn't really think this completely through so I'm not sure if that actually gives you real value over what lukas suggested -- feel free to disregard.

securesystemslib/dsse.py
```diff
@@ -58,21 +58,26 @@ def from_dict(cls, data: dict) -> "Envelope":
...
 
         return cls(payload, payload_type, signatures)
 
     def to_dict(self) -> dict:
         """Returns the JSON-serializable dictionary representation of self."""
 
+        signatures = []
+        for signature in self.signatures:
+            sig_dict = signature.to_dict()
+            sig_dict["sig"] = b64enc(sig_dict["sig"].encode("utf-8"))
```
adityasaky, line 74: I think this is incompatible with other DSSE envelope signing libraries. If I'm reading this right, we're writing the b64 encoding of the hex representation of the signatures. I think we should first decode the hex value before passing the decoded bytes to b64 encoding.

lukpueh: Good point, Aditya! DSSE spec doesn't go into details about the format, but I would also expect other implementations to base64-encode the "raw" signature bytes.

securesystemslib/dsse.py
```diff
@@ -71,7 +74,7 @@ def to_dict(self) -> dict:
         signatures = []
         for signature in self.signatures:
             sig_dict = signature.to_dict()
-            sig_dict["sig"] = b64enc(sig_dict["sig"].encode("utf-8"))
+            sig_dict["sig"] = b64enc(binascii.unhexlify(sig_dict["sig"]))
```
lukpueh, line 77:
```suggestion
            sig_dict["sig"] = b64enc(bytes.fromhex(sig_dict["sig"]))
```

securesystemslib/dsse.py
```diff
@@ -1,6 +1,7 @@
 """Dead Simple Signing Envelope
 """
 
+import binascii
```
lukpueh, line 4:
```suggestion

```

securesystemslib/dsse.py
```diff
@@ -60,7 +61,9 @@ def from_dict(cls, data: dict) -> "Envelope":
 
         signatures = []
         for signature in data["signatures"]:
-            signature["sig"] = b64dec(signature["sig"]).decode("utf-8")
+            signature["sig"] = binascii.hexlify(
+                b64dec(signature["sig"])
+            ).decode("utf-8")
```
lukpueh, line 66:
```suggestion
            signature["sig"] = b64dec(signature["sig"]).hex()
```

lukpueh (commented):
Since Python 3.5 this is built into the bytes type.

lukpueh (approved):

adityasaky:
Thanks @PradyumnaKrishna!

@lukpueh, can we cut a release with this PR included? I'd like to release in-toto-py with `--use-dsse` enabled.

PR #568, opened 2023-04-14
signer: move spx keygen, sign and verify API

lukpueh:
This was previously part of the `securesystemslib.keys` interface. To prepare for deprecation of this interface, and use of the new signer API as replacement, spx-related functionality and tests are moved.

This is a backwards incompatible change and a slight feature degradation:

- spx is now no longer available via the keys module, and thus no longer available via SSlibSigner/SSlibKey (i.e. the legacy API bridge). Given that spx was added recently, support via the legacy API does not seem necessary. The advantage is more modular code and less case-handling, also wrt different optional dependencies.

- Currently, SpxSigner does not implement from_priv_key_uri, which was previously available via SSlibSigner, and supported loading signers from file or envvar. Support may be added back later.

- The default keyid is computed differently than before (see 6c29cae63a151b739bbe4abb7f1b83bdd945c731)

See usage example in docstring.


lukpueh:
cc @rugo

securesystemslib/signer/_spx_signer.py
```diff
@@ -0,0 +1,131 @@
...
+    def __init__(self, private: bytes, public: SpxKey):
+        self.private_key = private
+        self.public_key = public
+
+    @classmethod
+    def new_(cls) -> "SpxSigner":
+        """Generate new SPHINCS+ key pair and return as SpxSigner.
+
+        NOTE: The Signer API is still experimental and key generation in
+        particular (see #466).
```
lukpueh, line 88: @jku: I'm still not sure what's most useful here. This method is a bit of a compromise between a `generate_raw`, which only returns private and public bytes and requires additional tooling to be usable by the signer; and the full set of `generate_file`, `import_file`, `generate_envvar`, `import_envvar` methods plus corresponding `import_from_priv_key_uri` as you suggest in https://github.com/secure-systems-lab/securesystemslib/issues/466#issuecomment-1367288846, which seemed like a lot of work and unclear to deduplicate from SSlibSigner.

Any opinions?

lukpueh: Okay, I changed my mind here. `SpxSigner.new_`  neither makes a lot of sense for the intended usage pattern of the Signer API, nor is it very useful to create tools on top (see details in c0671c4 commit message).

jku (approved):
I think this looks good and is probably the right way to go about it.

Following is just related thinking out loud:

It's a little worrying that we don't have a good story for how to handle e.g. keys in files in a general way but that's not this PRs fault... maybe Signer.from_priv_key_uri() actually could select the signer implementation _also_ based on the public key -- it seem like in some cases that would be the smart thing. So in case of "file:" URIs, it would do another lookup based on the keytype to find out if it should use SPXSigner or something else?

jku:
Yeah I think we're talking about roughly the same thing. The core issues are A) dispatch and B) signer construction: I was trying to say
* adding keytypes to the URI feels wrong: URI is supposed to answer the question _"how do I access the private key corresponding to this public key"_. Putting the public keytype in the uri just feels plain incorrect -- maybe this is me being too tied to the original idea but that is where I'm coming from: the URI was originally meant to define the private key storage mechanism
* the keytype (of the public key) has to be taken into account in the signer dispatch though
* The construction of a new signers/keys (and the potential storage of private key material that requires specific API) of course has to take keytype somehow into account as well
* a couple of options I can see:
  * `Signer.from_priv_key()` handles the "public key based dispatch" too. The shared implementations of e.g. a file-based-private-key could be in a Mixin... The thing to note here is that those Mixins are not fully internal: they form a part of the public API (API to e.g. store the private key in a file, or to get the envvar content)
  * signers like SpxSigner don't implement a `from_priv_key()`: new mid-level classes are made that handle "IO", e.g. FileSigner. Then FileSigner is the one that is dispatched to from Signer, FileSigner has an internal dispatch list of keytype-to-actual-signer, FileSigner has custom API to store the private key bytes in an encrypted file, and knows how to read and provide those bytes to the actual implementation like SpxSigner when it constructs one.

It's entirely possible that being less rigid about the private key URI meaning would be less work...

lukpueh:
Let's continue this conversation in #466

PR #578, opened 2023-05-01
signer: adopt upstream sigstore API change

lukpueh:
Since sigstore 1.1.2 `detect_credential` requires an "audience" argument. This patch passes the internal (!) DEFAULT_AUDIENCE, which is also used by the related sigstore cli.

Using the non-public constant is only a temporary fix until I hear back from upstream devs.

The PR also adds a missing permission in a related GitHub workflow .

### DIFF
--- a/.github/workflows/test-sigstore.yml
+++ b/.github/workflows/test-sigstore.yml
@@ -15,6 +15,7 @@ jobs:
 
     permissions:
       id-token: 'write' # ambient credential is used to sign
+      issues: 'write' # for filing an issue on failure
 
     steps:
       - name: Checkout securesystemslib
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ gcpkms = ["google-cloud-kms", "cryptography>=37.0.0"]
 hsm = ["asn1crypto", "cryptography>=37.0.0", "PyKCS11"]
 pynacl = ["pynacl>1.2.0"]
 PySPX = ["PySPX>=0.5.0"]
-sigstore = ["sigstore"]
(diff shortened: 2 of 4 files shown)


jku:
This is fixed in main now: https://github.com/sigstore/sigstore-python/pull/641: the argument is removed in next release (2.0 unless we really want a point release fix).

So we should add arguments to our calls :) Maybe the easy fix to get builds working is prevent sigstore 1.1.2 from being used in requirements?

lukpueh:
> So we should add arguments to our calls :) Maybe the easy fix to get builds working is prevent sigstore 1.1.2 from being used in requirements?

I'm fine either way. Given that our sigstore signer is highly experimental and uses other non-public API in sigstore-python too, I think, this fix is okay. Let's have our users of this (you) decide. ;)

If we don't merge here, I can PR the ci fixing commit of this branch separately.

jku:
> > So we should add arguments to our calls :) Maybe the easy fix to get builds working is prevent sigstore 1.1.2 from being used in requirements?
> 
> I'm fine either way. Given that our sigstore signer is highly experimental and uses other non-public API in sigstore-python too, I think, this fix is okay. Let's have our users of this (you) decide. ;)

Oops, my brain left a "not" out of that original reply: I meant we should **not** add arguments to our calls because that will then break on next release

lukpueh:
Oh! I didn't look at the fix. I assumed they made the argument optional. :D Yeah, then excluding 1.1.2 makes sense. I'll do it in this PR if you don't mind.

jku (approved):
lgtm, let's see what we get this time...

PR #585, opened 2023-05-24
key: refactor SSlibKey.verify_signature 

lukpueh:
EDIT 2023/05/20: 
- This is ready for review and can be merged as is.
- Current python-tuf tests pass without changes.
- Legacy code is not removed in this PR for the reasons outlined in 692b6acf2faced30e45f689582e5c13e01ff3954. See [full diff](https://github.com/secure-systems-lab/securesystemslib/compare/main..lukpueh:refactor-sslib-signature-verification-rm-legacy) including the removal for how much more lightweight the code becomes: `13 additions and 1,010 deletions.`(!)
---

Signature verification for "securesystemslib keys" was previously implemented in 'rsa_keys', 'ecdsa_keys' and 'ed25519_keys' modules, which were called from `SSlibKey.verify_signature` via the legacy interface function `keys.verify_signature()`.

This commit moves the entire implementation to SSlibKey, which will allow us (in a subsequent commit) to drastically decrease LOC count and drop 'nacl' optional dependency for ed25519 keys, in favour of 'pyca/cryptography', which we already use for all other sslib keys.

An alternative design for this refactor used separate RSAKey, ECDSAKey and ED25510Key classes to replace SSlibKey, but that mostly added redundant boilerplate code. To the user it shouldn't matter, so let's do what makes sense from maintainer perspective.

### DIFF
--- a/securesystemslib/signer/_key.py
+++ b/securesystemslib/signer/_key.py
@@ -253,85 +253,81 @@ class SSlibKey(Key):
 
         return padding
 
+    def _verify_ed25519_fallback(self, signature: bytes, data: bytes) -> None:
+        """Helper to verify ed25519 sig if pyca/cryptography is unavailable."""
+        try:
+            public_bytes = bytes.fromhex(self.keyval["public"])
+            checkvalid(signature, data, public_bytes)
+
+        except SignatureMismatch as e:
+            raise UnverifiedSignatureError from e
+
+    def _verify(self, signature: bytes, data: bytes) -> None:
+        """Helper to verify signature using pyca/cryptography (default)."""
+        try:
+            key: PublicKeyTypes
+            if self.scheme in [
+                "rsassa-pss-sha224",
+                "rsassa-pss-sha256",
+                "rsassa-pss-sha384",
+                "rsassa-pss-sha512",
+                "rsa-pkcs1v15-sha224",
+                "rsa-pkcs1v15-sha256",
+                "rsa-pkcs1v15-sha384",
+                "rsa-pkcs1v15-sha512",
+            ]:
+                key = cast(RSAPublicKey, self._from_pem())
+                padding_name, hash_name = self.scheme.split("-")[1:]
+                hash_algorithm = self._get_hash_algorithm(hash_name)
+                padding = self._get_rsa_padding(padding_name, hash_algorithm)
+                key.verify(signature, data, padding, hash_algorithm)
+
+            elif self.scheme in [
+                "ecdsa-sha2-nistp256",
+                "ecdsa-sha2-nistp384",
+            ]:
+                key = cast(EllipticCurvePublicKey, self._from_pem())
+                hash_name = f"sha{self.scheme[-3:]}"
+                hash_algorithm = self._get_hash_algorithm(hash_name)
+                signature_algorithm = ECDSA(hash_algorithm)
+                key.verify(signature, data, signature_algorithm)
+
+            elif self.scheme in ["ed25519"]:
+                public_bytes = bytes.fromhex(self.keyval["public"])
+                key = Ed25519PublicKey.from_public_bytes(public_bytes)
+                key.verify(signature, data)
+
+            else:
+                raise ValueError(f"unknown scheme '{self.scheme}'")
+
+        except InvalidSignature as e:
+            raise UnverifiedSignatureError from e
+
     def verify_signature(self, signature: Signature, data: bytes) -> None:
-        if signature.keyid != self.keyid:
-            raise VerificationError from ValueError(
-                f"keyid mismatch: 'key id: {self.keyid}"
-                f" != signature keyid: {signature.keyid}'"
-            )
+        try:
+            if signature.keyid != self.keyid:
+                raise ValueError(
+                    f"keyid mismatch: 'key id: {self.keyid}"
+                    f" != signature keyid: {signature.keyid}'"
+                )
 
-        sig = bytes.fromhex(signature.signature)
+            signature_bytes = bytes.fromhex(signature.signature)
 
-        if CRYPTO_IMPORT_ERROR:
-            try:
+            if CRYPTO_IMPORT_ERROR:
                 if self.scheme != "ed25519":
                     raise UnsupportedLibraryError(CRYPTO_IMPORT_ERROR)
 
-                public_bytes = bytes.fromhex(self.keyval["public"])
-                checkvalid(sig, data, public_bytes)
+                return self._verify_ed25519_fallback(signature_bytes, data)
 
-            except SignatureMismatch as e:
-                raise UnverifiedSignatureError(
-                    f"Failed to verify signature by {self.keyid}"
-                ) from e
+            return self._verify(signature_bytes, data)
 
-            except Exception as e:
-                logger.info(
-                    "Key %s failed to verify sig: %s", self.keyid, str(e)
-                )
-                raise VerificationError(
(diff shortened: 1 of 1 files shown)


securesystemslib/signer/_key.py
```diff
@@ -180,22 +215,80 @@ def from_dict(cls, keyid: str, key_dict: Dict[str, Any]) -> "SSlibKey":
...
+                algo = self._hash_algo(algo_name)()
+                key.verify(sig, data, ECDSA(algo))
+
+            elif self.scheme in ["ed25519"]:
+                public_bytes = bytes.fromhex(self.keyval["public"])
+                key = Ed25519PublicKey.from_public_bytes(public_bytes)
+                key.verify(sig, data)
+
+            else:
+                raise ValueError(f"unknown scheme '{self.scheme}'")
```
lukpueh, line 280: I tried different variants here: One, where I didn't dissect the scheme string to get hash and padding objects, but bluntly added a separate elif block per scheme; or another one, where I hid the verification argument assembly in a helper and called `key.verify` only once using some `*args` magic.

I opted for this one here, because it felt like a good trade-off between readability, succinctness and mypy-friendlyness. That said, I'm happy to take any code style advice.

lukpueh: Turns out, the refactored code in `SSlibSigner.sign` looks 90% the same. This means, having helpers that dissect the scheme string in order to call the right pyca/cryptography functions definitely makes sense and should be available to SSlibKey and SSlibSigner. And IIRC other signers use the same scheme strings and pyca/cryprography too.

I'll try to deduplicate here.

lukpueh: > having helpers that dissect the scheme string in order to call the right pyca/cryptography functions definitely makes sense and should be available to SSlibKey and SSlibSigner.

They should actually be available to other signers too, e.g. _gcp_signer:

https://github.com/secure-systems-lab/securesystemslib/blob/135567fa04f10d0c6a4cd32eb45ce736e1f50a93/securesystemslib/signer/_gcp_signer.py#L161-L172
 
Let's move on with some duplicate helpers here, and ticketize this issue for later.

securesystemslib/signer/_key.py
```diff
@@ -180,22 +215,99 @@ def from_dict(cls, keyid: str, key_dict: Dict[str, Any]) -> "SSlibKey":
...
+                )
+            sig = bytes.fromhex(signature.signature)
+
+            if CRYPTO_IMPORT_ERROR:
+                if self.scheme == "ed25519":
+                    public_bytes = bytes.fromhex(self.keyval["public"])
+                    checkvalid(sig, data, public_bytes)
+                    return
+
+                raise exceptions.UnsupportedLibraryError(CRYPTO_IMPORT_ERROR)
```
jku, line 267: I guess another potential structure would be 

```
 if CRYPTO_IMPORT_ERROR:
    if self.scheme != "ed25519":
        raise exceptions.UnsupportedLibraryError(CRYPTO_IMPORT_ERROR)
    try:
        # call vendored ed25519
    except SignatureMismatch as e:
        # handle
    except Exception as e:
        # handle
else:
    try:
        # the actual switch for cryptography calls goes here
    except InvalidSignature as e:
        # handle
    except Exception as e:
        # handle
```
in pseudocode that feels easier to understand (since the try-block is not shared) but in practice I'm unsure if it helps

jku (approved):
I think this looks nice. Approving as is but commented one possible change.

I have a question about testing though. How well is `Key.verify_signature()` getting tested? Obviously we're not adding new API here but... before this it wasn't really that critical that `Key.verify_signature()` had good coverage as the implementation was known to be a wrapper over `keys.verify_signature()` with some testing of its own. Now the situation is a bit different.

So... how do you feel about the test coverage for the "new" implementation? Is it good enough?

lukpueh:
> So... how do you feel about the test coverage for the "new" implementation? Is it good enough?

Oh, I forgot to mention that I have already added more tests for `SSlibKey.verify_signature` tests in preparation for this PR beforehand, also to show that the refactor really doesn't change the behaviour: see #555, #556.

So, I do feel quite confident.

securesystemslib/signer/_key.py
lukpueh: @jku: is 73f84e2 what you had in mind? I'm not sure if it makes things much easier to read.

lukpueh: Okay, I think this is good enough. Thanks for your help, Jussi! Lmk if this still looks good to you.

jku (approved):
Yes, feels much easier to understand.

The exceptions still feel a bit clunky (like handling generic Exception) but I think my suggestions are not very actionable... so LGTM

securesystemslib/signer/_key.py
```diff
@@ -180,23 +219,115 @@ def from_dict(cls, keyid: str, key_dict: Dict[str, Any]) -> "SSlibKey":
...
-            exceptions.CryptoError,
-            exceptions.FormatError,
-            exceptions.UnsupportedAlgorithmError,
-            exceptions.UnsupportedLibraryError,
-        ) as e:
-            logger.info("Key %s failed to verify sig: %s", self.keyid, str(e))
-            raise exceptions.VerificationError(
+
+            signature_bytes = bytes.fromhex(signature.signature)
+
```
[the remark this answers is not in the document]
lukpueh: > this could all be outside of the try block... raising things just to catch them in the same method seems odd
> 
> * just raising a ValueError from this method _might_ be fine: the idea is that it is a programming error to try to verify a signature that does not match the key --

I added this here to pass python-tuf tests. Previously, the keyid mismatch triggered a FormatError, which was caught and re-raised as VerificationError and asserted for in tests. Without the keyid check, this will try verification and fail as UnverifiedSignatureError.

Obviously, we can change the python-tuf test. Then we should maybe update the Key.verify_signature docs.

> * if you don't want to raise ValueError from this method, you could just raise VerificationError before the try block

Yes, I considered that too. But I thought it is semantically more correct to raise a ValueError as VerificationError, and that I could re-use the generic log.info + raise VerificationError block used for everything that is not an UnverifiedSignatureError. But now that you point this out, it does not seem fully correct to re-raise this ValueError as "Unknown failure"

> 
> That said, doing this does have similarities to how the `_verify*` methods raise ValueError etc that are then handled in this method... so maybe it's fine

Yes, it is consistent, which does not mean it is right. :) I could also raise a VerificationError in `_verify*`, which I don't catch here and just let fall through.

securesystemslib/signer/_key.py
```diff
@@ -180,23 +219,115 @@ def from_dict(cls, keyid: str, key_dict: Dict[str, Any]) -> "SSlibKey":
...
+        except UnverifiedSignatureError as e:
+            raise UnverifiedSignatureError(
+                f"Failed to verify signature by {self.keyid}"
+            ) from e
+
+        except Exception as e:
+            logger.info("Key %s failed to verify sig: %s", self.keyid, e)
+            raise VerificationError(
                 f"Unknown failure to verify signature by {self.keyid}"
             ) from e
```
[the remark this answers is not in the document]
lukpueh: Yeah, you are right. The question is, what do we actually want to raise as VerificationError?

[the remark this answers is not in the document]
lukpueh: Should UnsupportedLibraryError be re-raised as VerificationError? Was https://github.com/secure-systems-lab/securesystemslib/pull/556/commits/e6529cd31ad11642a42eb1c4a4005c613bffc56b a mistake?

Issue #468, opened 2022-11-29
signer API: review exceptions


jku:
This is after #456:

Signer API exceptions may not be 100% coherent: it's an API that's grown over time. It's also not as critical that the exceptions are well designed as e.g. python-tuf verification API (as signing happens in much more controlled environments: new keys may be added but that never happens as a surprise to user). Still, would be good to review the exception usage.

One especially tricky case is Signer implementation specific exceptions -- like a KMS signer raising on network error. Currently we just document this as a possibility.

jku:
There is, AFAICT, no really nice solution to this. We will not be able to handle all exceptions that e.g. google-cloud-kms might raise in a generic and useful way. So we can either tell callers to handle all errors or we can handle all the errors ourselves, and raise a generic SignerException...

The latter would normally be preferable (it's what we do in python-tuf serialization for example) but I can also imagine cases where caller wants to handle some specific errors: in that case wrapping the actual error just makes things more difficult for the caller.

So I'm leaning towards documenting Signers like 
> callers should prepare for `sign()` and `import_()` to raise Signer implementation specific errors that may be undocumented: <add generic try-except example here>

it's not amazing but I can live with that

lukpueh:
Let's also review if the exceptions we do know are properly documented and tested.

PR #588, opened 2023-05-26
Add Azure Signer support

malancas:
Fixes: #

This pull request adds an Azure signer implementation so we can use Azure's KMS offering Key Vault. This signer currently only supports ECDSA keys. RSA key support can be added later in a follow up pull request.


jku:
Based on a quick glance, looks nice!
Some early comments:
* tests are failing because the signer modules should be importable even without the optional deps (and cryptography is an optional dep -- you could probably move the import inside the try-except you already have)
* `tox -e lint` runs the same lint tests locally as the CI runs
* I think the `import_()` should not take the fully built "private key URI" as argument -- that forces the caller to understand and be able to build the URI. Instead it should take the logical arguments that azure vault users will understand and the import method should build the URI 


Finally a suggestion/comment, feel free to disregard:
* from experience with GCP, I would suggest you might want to use the full azure keyid that contains the vault and key version, if that is easily available to users (because we likely do want a specific key version): 'https://myvault.vault.azure.net/keys/my-key/my-key-version'
  * this full keyid might work as the import argument
  * you probably won't need to call KeyClient() during `__init__()` since you can use the full keyid to create CryptographyClient

jku:
Since you're allowing maintainer pushes (thanks!), I'll take the liberty of adding some annotation and lint fixes: 
* most changes come from running black and isort (like `tox -e lint` does) but there are some lint fixes and some annotation tweaks as well
* I've only touched things that should not affect intended function
* the unrelated tests should pass after this
* there are still linter issues (both pylint and mypy) that require actual changes

jku:
> IIRC most of them are about some azure class not having attributes

scratch that comment: these were seemingly all in the code that you managed to remove in the last commits

malancas:
@jku All of the linting errors and warnings should now be fixed. Thanks for your feedback and help with this PR

kommendorkapten:
We can take this out of "draft" now @malancas right?

malancas:
@kommendorkapten yep, I'll change that now.

securesystemslib/signer/_azure_signer.py
```diff
@@ -0,0 +1,272 @@
...
+        # length of at most nLen with the most significant byte first (i.e.
+        # big endian)
+        # https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/cs01/pkcs11-curr-v3.0-cs01.html#_Toc30061178
+        r_s_len = int(len(response.signature) / 2)
+        r = int.from_bytes(response.signature[:r_s_len], byteorder="big")
+        s = int.from_bytes(response.signature[r_s_len:], byteorder="big")
+
+        # Create an ASN.1 encoded Dss-Sig-Value to be used with
+        # pyca/cryptography
+        dss_sig_value = binascii.hexlify(encode_dss_signature(r, s)).decode(
```
lukpueh, line 268: nit: the `bytes` type has a hex method these days.

securesystemslib/signer/_azure_signer.py
```diff
@@ -0,0 +1,272 @@
...
+            cred = DefaultAzureCredential()
+            self.crypto_client = CryptographyClient(
+                az_key_uri,
+                credential=cred,
+            )
+            self.signature_algorithm = self._get_signature_algorithm(
+                public_key,
+            )
+            self.hash_algorithm = self._get_hash_algorithm(public_key)
+        except UnsupportedKeyType as e:
```
lukpueh, line 80: Should UnsupportedKeyType be public API? Currently it isn't, so a user can't really handle it. Would a built-in exception type also do the trick?

Also, is it really worth to catch, log and re-raise? Doesn't the original error message, which you control, provide enough info? This pattern seems more useful below, where you contextualise the 3rd-party HttpResponseError, although even there I wonder if the stack trace alone might be enough. 🤷 

FYI: we have a discussion ticket about Signer exceptions in https://github.com/secure-systems-lab/securesystemslib/issues/468. Curious about your opinions.

lukpueh (commented):
Great work, @malancas!

jku (commented):
I think the code looks good. left some notes but nothing that I think requires changes if you'd like to merge as is.

The one thing I'd like to bring up is testing... I know it's tricky to setup a proper test with the KMS service dependency (I talk a bit more about that in a comment) and we can't _require_ a real time testing setup. But right now we as maintainers are in a situation where we can't see
* example code -- how is it supposed to be used
* proof that it actually works anywhere -- logs of a successful signing

Do you have good ideas about this? 
Are there public logs of this working? 
Should we add testing code (similar to `check_kms_signers`) even if it is disabled and only works in your fork?

securesystemslib/signer/_azure_signer.py
jku: I left a note about this: https://github.com/secure-systems-lab/securesystemslib/pull/588#discussion_r1211249464

TLDR: I think it's not really public API, it's just part of the larger group of exceptions that we know we may raise that callers are not really meant to handle

lukpueh: Agreed. I was actually trying to challenge the need for a custom exception and the way it is handled internally. But I'm okay with it. The code is still very readable and extra info for the caller is nice.

jku:
> I can share credentials with you if you want to replicate the test locally.

Oh no need for that -- would just be good to know that _if_ someone needs to test this, they mostly just need to setup the KMS (and maybe enable the tests or run another tox env or something simple like that).

securesystemslib/signer/_azure_signer.py
```diff
@@ -0,0 +1,272 @@
...
+    )
+except ImportError:
+    AZURE_IMPORT_ERROR = (
+        "azure-cloud-kms library required to sign with Azure Cloud keys."
+    )
+
+logger = logging.getLogger(__name__)
+
+
+class UnsupportedKeyType(Exception):
```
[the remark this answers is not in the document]
malancas: That's correct, I just wanted to add a little bit more information around why fatal failure would occur in a few spots.

jku (approved):
:+1:  thanks for the test: it works exactly like I expected but it's good to actually see it. 

The `requirements-kms.txt` changes are now not needed (since it is used as requirements for tests.check_kms_signers which doesn't currently include azure). I can remove those changes before merging.

malancas:
Great, thank you for the reviews!

PR #590, opened 2023-05-30
signer: refactor SSlibSigner

lukpueh:
*edit: ticketized and removed TODOs below (#593, #594, #592)*

---

This patch refactors the SSlibSigner signing implementation, akin to #585, which refactored signature verification (see PR for details about legacy code and rationale).

Unlike, #585 this patch does not only condense and move the code fo singing, but creates a new hierarchy of signers to achieve tw additional goals:

1. Provide tiny rsa, ecdsa and ed25519 pyca/cryptography Signer implementations, which are independent of private key serialization formats, above all, of the proprietary legacy keydict format. This is particularly interesting, when refactoring existing or designing new key generation or import interfaces, where it would be annoying to move back and forth over the legacy keydict.

2. Preserve SSlibSigner including its internal legacy keydict data structure.  SSlibSigner is and remains a backwards-compatibility crutch. Breaking its existing users to make it a little less awkward would defeat that purpose.  And even though the Signer API doc says that the internal data structure is not part of the public API, users may rely on it (python-tuf actually does so at least in tests and demos).

To achieve these goals, SSlibSigner becomes a container for the newly added CryptoSigner class, whose implementations can also be used as independent Signers, and above all created or imported, with very few lines of pyca/cryptography code.

**Caveat:**
Latest python-tuf tests pass against this patch, except for one, which expects a keydict deserialization failure in `sign`, which now happens in `__init__` initialization time. This seems feasible to fix in python-tuf.

Also note that private key format errors are now ValueErrors and no longer unreliably either FormatErrors or sometimes UnsupportedAlgorithmErrors.

*See [full diff](https://github.com/secure-systems-lab/securesystemslib/compare/main..lukpueh:securesystemslib:refactor-sslib-sign-rm-legacy) including the removal of related legacy code for how much more lightweight the code becomes: `194 additions and 1,264 deletions`(!)*

### DIFF
--- a/securesystemslib/signer/_signer.py
+++ b/securesystemslib/signer/_signer.py
@@ -3,14 +3,48 @@
 import logging
 import os
 from abc import ABCMeta, abstractmethod
-from typing import Any, Callable, Dict, Optional, Type
+from typing import Any, Callable, Dict, Optional, Type, cast
 from urllib import parse
 
 import securesystemslib.keys as sslib_keys
+from securesystemslib.exceptions import UnsupportedLibraryError
 from securesystemslib.formats import encode_canonical
 from securesystemslib.hash import digest
 from securesystemslib.signer._key import Key, SSlibKey
 from securesystemslib.signer._signature import Signature
+
+CRYPTO_IMPORT_ERROR = None
+try:
+    from cryptography.hazmat.primitives.asymmetric.ec import (
+        ECDSA,
+        EllipticCurvePrivateKey,
+    )
+    from cryptography.hazmat.primitives.asymmetric.ed25519 import (
+        Ed25519PrivateKey,
+    )
+    from cryptography.hazmat.primitives.asymmetric.padding import (
+        MGF1,
+        PSS,
+        AsymmetricPadding,
+        PKCS1v15,
+    )
+    from cryptography.hazmat.primitives.asymmetric.rsa import (
+        AsymmetricPadding,
+        RSAPrivateKey,
+    )
+    from cryptography.hazmat.primitives.asymmetric.types import PrivateKeyTypes
+    from cryptography.hazmat.primitives.hashes import (
+        SHA224,
+        SHA256,
+        SHA384,
+        SHA512,
+        HashAlgorithm,
+    )
+    from cryptography.hazmat.primitives.serialization import (
+        load_pem_private_key,
+    )
+except ImportError:
+    CRYPTO_IMPORT_ERROR = "'pyca/cryptography' library required"
 
 logger = logging.getLogger(__name__)
 
@@ -164,6 +198,7 @@ class SSlibSigner(Signer):
 
     def __init__(self, key_dict: Dict):
         self.key_dict = key_dict
+        self._crypto_signer = CryptoSigner.from_securesystemslib_key(key_dict)
 
     @classmethod
     def from_priv_key_uri(
@@ -212,6 +247,7 @@ class SSlibSigner(Signer):
 
         keydict = public_key.to_securesystemslib_key()
         keydict["keyval"]["private"] = private
+
         return cls(keydict)
 
     def sign(self, payload: bytes) -> Signature:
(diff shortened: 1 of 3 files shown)


lukpueh:
Just went through all legacy sign/verify tests, of which ~98% assert FormatError for bad function arguments. Most of them are obsolete or different now, given the class-based design and use of mypy. Still, we could probably better test error scenarios depending on our decisions in  https://github.com/secure-systems-lab/securesystemslib/issues/468 and https://github.com/secure-systems-lab/securesystemslib/issues/595 (I'll make a not in those issues).

Apart from that, basic signature verification testing is already fully covered by new signer API tests, using pre-generated signatures (see https://github.com/secure-systems-lab/securesystemslib/pull/585#issuecomment-1570174679).

Similar sign -> verify tests already exist for all key types, but not all schemes. I can change the latter, and maybe port the "auto rsa pss salt length"-test from #436.

lukpueh:
> Similar sign -> verify tests already exist for all key types, but not all schemes. I can change the latter, and maybe port the "auto rsa pss salt length"-test from #436.

Added sign-verify tests for all schemas here and rsa pss salt length test in #651. I think this PR is good to go.

securesystemslib/signer/_signer.py
```diff
@@ -225,5 +261,146 @@ def sign(self, payload: bytes) -> Signature:
                 securesystemslib.exceptions.UnsupportedAlgorithmError:
                 Signing errors.
         """
```
lukpueh, line 263: TODO: Update exceptions in docstring

securesystemslib/signer/_signer.py
```diff
@@ -225,5 +261,146 @@ def sign(self, payload: bytes) -> Signature:
...
-        sig_dict = sslib_keys.create_signature(self.key_dict, payload)
-        return Signature(**sig_dict)
+        return self._crypto_signer.sign(payload)
+
+
+class CryptoSigner(Signer, metaclass=ABCMeta):
```
jku, line 267: What's the advantage of deriving here? You say that the individual CryptoSigners could be used as Signers directly but if they don't implement `from_priv_key_uri()` I'm not sure I agree.  The class hierarchy then has a lot of "Signers"  but only some can be used as documented: others will throw a RuntimeError

I think the division of responsibility makes sense but could this just be a clearly internal class that isn't a Signer?

jku (commented):
I think the core idea makes sense: separate two responsibilities:
* loading the private keys from storage (handled by SSLibSigner at the moment, could change in future)
* actual cryptography API usage

Code looks good. Two questions:
* is it not confusing that CryptoSigner claims to be a Signer? (more in a code comment)
* should we maybe move SSlibSigner out of _signer.py? The interface definition might be easier to understand without a screenful of crypto imports

securesystemslib/signer/_signer.py
```diff
@@ -3,15 +3,49 @@
 import logging
 import os
 from abc import ABCMeta, abstractmethod
-from typing import Any, Callable, Dict, Optional, Type
+from typing import Any, Callable, Dict, Optional, Type, cast
 from urllib import parse
 
 import securesystemslib.keys as sslib_keys
```
jku, line 9: As a note (unrealted to this PR), I was wondering what was still keeping this import in:
We call `decrypt_key()` for the encrypted private key use case (there's also a matching `securesystemslib.keys.encrypt_key()` call).

I'm sure cryptography has something we could use instead?

lukpueh:
> Code looks good. Two questions:
> 
> * is it not confusing that CryptoSigner claims to be a Signer? (more in a code comment)

Agreed, when looking at this PR alone `CryptoSigner` looks like an implementation detail of `SSlibSigner`.

However, I did intend for `CryptoSigner` to become a reasonable replacement for `SSlibSigner`. I just didn't want to open the can of worms of private key de/serialization just yet. This PR is only about `CryptoSigner.sign`. #604 adds, among other things, a `CryptoSigner.from_priv_key_uri` implementation.


> * should we maybe move SSlibSigner out of _signer.py? The interface definition might be easier to understand without a screenful of crypto imports

Yes, we should. I had a version of this PR, where I did move it out, but then reconsidered to keep the diff simpler and also due to some circular import hiccups, which surely can be fixed.

securesystemslib/signer/_signer.py
lukpueh: > I'm sure cryptography has something we could use instead?

Yes, it has, and #604 uses it consistently for all 3 supported key types to import encrypted private key files from a standard PEM format simply by...

https://github.com/secure-systems-lab/securesystemslib/blob/a1e280fbb9751c5bfc286e0e3f7c96df9debf0b8/securesystemslib/signer/_signer.py#L305

#604 does not yet provide API to serialise and dump corresponding private keys to disk but that would be extremely easy to to add, using `CryptoSigner.private_key.private_bytes` as [described](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#key-serialization).

`securesystemslib` really shouldn't implement any more complex encryption or decryption logic than this.

lukpueh:
superseded by #604

PR #599, opened 2023-06-08
Bumped cryptography dep to 40 as it's requried

kommendorkapten:
This line https://github.com/secure-systems-lab/securesystemslib/blob/main/securesystemslib/signer/_key.py#L36 requires cryptography v40 and higher.


jku:
I don't see an issue with doing this.

If we do want to keep supporting older cryptography, we could pretty easily just not use the new union type that is the problem.

jku (approved):

PR #622, opened 2023-08-21
Automate Signer API docs for RTD

### LINKED ISSUE #271: Revise and automate docs with sphinx and host on readthedocs

lukpueh:
**Description**

Adds configuration for Sphinx and Readthedocs, to automate securesystemslib API documentation. This PR includes a minimal landing page and enables docstring-autodocs for interface classes and variables of Signer API only. Additionally, docstrings are modified slightly to make them Sphinx-friendly.

- See commit messages for details
- See preview of this branch at https://python-securesystemslib.readthedocs.io.

fixes #271 (Note: The ticket is a bit broader than this. I still suggest to close and ticketize remaining tasks as described below)

**TODO** (in follow-up PRs)

- Configure web hook for auto builds on RTD
- Finalize Signer API (docs)
  Should Signer subclasses be added? The idea behind the URI system was that the actual implementations can be hidden from the user. But currently, the implementations  have some important public methods, e.g. `import_`, `generate`, etc., to even construct the URIs. Asked differently, what docs are missing for the Signer API to be fully usable?
- Add usage examples
   Ideally, basic usage examples are included in the API docs. See Jussi's blog post and examples added in [#604](https://github.com/secure-systems-lab/securesystemslib/pull/604). 
- Add other API docs 
   At least `dsse` and `formats.encode_canonical`.
- Add/update GitHub community doc (readme, maintainers, contribution, etc.)

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,21 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+version: 2
+build:
+  os: ubuntu-22.04
+  apt_packages:
+    - swig
+    - softhsm2
+  tools:
+    python: "3.11"
+
+sphinx:
+  builder: html
+  configuration: docs/conf.py
+  fail_on_warning: true
+
+python:
+  install:
+  - requirements: requirements-docs.txt
--- a/securesystemslib/signer/_signature.py
+++ b/securesystemslib/signer/_signature.py
@@ -15,6 +15,12 @@ class Signature:
     Provides utility methods to easily create an object from a dictionary
     and return the dictionary representation of the object.
 
+    Args:
+        keyid: HEX string used as a unique identifier of the key.
+        sig: HEX string representing the signature.
+        unrecognized_fields: Dictionary of all attributes that are not managed
+            by securesystemslib.
+
     Attributes:
         keyid: HEX string used as a unique identifier of the key.
         signature: HEX string representing the signature.
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ env/*
 tests/htmlcov/*
 .DS_Store
 .python-version
+
+# Sphinx documentation
+docs/_build/
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line, and also
+# from the environment for the first two.
+SPHINXOPTS    ?=
+SPHINXBUILD   ?= sphinx-build
+SOURCEDIR     = .
+BUILDDIR      = _build
+
+# Put it first so that "make" without argument is like "make help".
+help:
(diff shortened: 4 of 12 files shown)


securesystemslib/signer/_signature.py
```diff
@@ -15,6 +15,12 @@ class Signature:
     Provides utility methods to easily create an object from a dictionary
     and return the dictionary representation of the object.
 
+    Args:
+        keyid: HEX string used as a unique identifier of the key.
+        sig: HEX string representing the signature.
+        unrecognized_fields: Dictionary of all attributes that are not managed
+            by securesystemslib.
```
lukpueh, line 22: Repeating Args and Attributes seems silly. I'd prefer to only use "Args" here and say *"All parameters named below are not just constructor arguments but also instance attributes."* as we do in the `Key` class.

The problem is we use `sig` as arg name and `signature` as attribute name. I think we should consolidate them now. Changing the (positional) arg name seems less disrupting.

.readthedocs.yaml
```diff
@@ -0,0 +1,21 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+version: 2
+build:
+  os: ubuntu-22.04
+  apt_packages:
+    - swig
+    - softhsm2
```
lukpueh, line 10: `swig` and `softhsm2` are needed for `pykcs11`, which we install in requirements-docs.txt via `requirements-pinned.txt`. 

We can probably shuffle requirements* files so that we don't need to install any optional requirements.

jku:
> Add usage examples
> Ideally, basic usage examples are included in the API docs. See Jussi's blog post and examples added in #604.

I've been thinking we should have an example app (like a tool to sign and verify files) to really show why the design makes sense. 

Sigstore would be the perfect signer implementation to use (because the example would be simplest since there are no private key files to worry about) but that would imply making SigstoreSigner non-experimental which I don't really want yet...

.readthedocs.yaml
jku: suggestion makes sense but if rtd is fine with us installing random software... I suppose this is ok

securesystemslib/signer/_signature.py
jku: yeah, consolidate now sounds good: No-one outside of securesystemslib should should be calling Signature() so the fallout should be minimal.

This doesn't have to be in this PR though

jku (approved):
Looks like a great start.

I think the `make.bat+Makefile` are probably a worse solution than `tox -e docs` used in python-tuf because 
* tox setup is something we can test on CI 
* tox setup is actually less code to maintain

but this is also fine.

jku:
I'll merge this as, and file a new issue for at least the example

PR #627, opened 2023-08-25
Allow more AWS Signer options

MVrachev:
Allowing "endpoint_url" and "region_name" arguments to the AWS signer will allow the use of AWSSigner for a broader set of use cases such as local testing with awslocal and localstack which emulates multiple AWS services.

I required those arguments when working on support for AWS KMS for RSTUF.

For reference: https://github.com/repository-service-tuf/repository-service-tuf-worker/issues/344

### DIFF
--- a/securesystemslib/signer/_aws_signer.py
+++ b/securesystemslib/signer/_aws_signer.py
@@ -45,6 +45,8 @@ class AWSSigner(Signer):
     Arguments:
         aws_key_id (str): AWS KMS key ID or alias.
         public_key (Key): The related public key instance.
+        endpoint_url: (str): AWS endpoint URL.
+        region_name: (str): AWS region to use. Example: us-east-1.
 
     Returns:
         AWSSigner: An instance of the AWSSigner class.
@@ -58,14 +60,22 @@ class AWSSigner(Signer):
 
     SCHEME = "awskms"
 
-    def __init__(self, aws_key_id: str, public_key: Key):
+    def __init__(
+        self,
+        aws_key_id: str,
+        public_key: Key,
+        endpoint_url: Optional[str] = None,
+        region_name: Optional[str] = None
+    ):
         if AWS_IMPORT_ERROR:
             raise UnsupportedLibraryError(AWS_IMPORT_ERROR)
 
         self.hash_algorithm = self._get_hash_algorithm(public_key)
         self.aws_key_id = aws_key_id
         self.public_key = public_key
-        self.client = boto3.client("kms")
+        self.client = boto3.client(
+            "kms", endpoint_url=endpoint_url, region_name=region_name
+        )
         self.aws_algo = self._get_aws_signing_algo(self.public_key.scheme)
 
     @classmethod
@@ -83,7 +93,13 @@ class AWSSigner(Signer):
         return cls(uri.path, public_key)
 
     @classmethod
-    def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
+    def import_(
+        cls,
+        aws_key_id: str,
+        local_scheme: str,
+        endpoint_url: Optional[str] = None,
+        region_name: Optional[str] = None
+    ) -> Tuple[str, Key]:
         """Loads a key and signer details from AWS KMS.
 
         Returns the private key uri and the public key. This method should only
@@ -92,6 +108,8 @@ class AWSSigner(Signer):
         Arguments:
             aws_key_id (str): AWS KMS key ID.
             local_scheme (str): Local scheme to use.
+            endpoint_url: (str): AWS endpoint URL.
+            region_name: (str): AWS region to use. Example: us-east-1.
 
         Returns:
             Tuple[str, Key]: A tuple where the first element is a string
@@ -107,7 +125,9 @@ class AWSSigner(Signer):
         if AWS_IMPORT_ERROR:
             raise UnsupportedLibraryError(AWS_IMPORT_ERROR)
 
-        client = boto3.client("kms")
+        client = boto3.client(
+            "kms", endpoint_url=endpoint_url, region_name=region_name
+        )
         request = client.get_public_key(KeyId=aws_key_id)
         kms_pubkey = serialization.load_der_public_key(request["PublicKey"])
 


jku:
Two ideas/comments:
* this doesn't add the support into the privkeyuri. the uri could be like `awskms://<path>?endpoint=<endpoint>&region=<region>` with the args being optional. However...
* There might be another option if these variables are more _application config_ than key specific secrets: these systems often support environment variables and config files for something like this and looks like boto does too https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables . I think if documented, that's a fine way to support this with zero code in the library: this allows the application to configure every detail of AWS and not just the ones we expose... However, if these are key-specific variables (likely to be different if you use another AWS key), then I agree Signer attributes make sense

jku:
> There might be a better option if these variables are more _application config_ than key specific secrets/variables

In case this needs more context: the way I imagined Signers to work is that the signers handle _key specific_ details. The signing system may require other more generic configuration but that is by design "off-band" (as an example both AWS and GCP require the authentication to the KMS to be done: authentication data is not included in the Signer arguments). 

Things like "region" might be argued to be in either group but they feel more generic config: you might change the region your app uses at some point but you are unlikely to choose a different region for different keys

MVrachev:
> > There might be a better option if these variables are more _application config_ than key specific secrets/variables
> 
> In case this needs more context: the way I imagined Signers to work is that the signers handle _key specific_ details. The signing system may require other more generic configuration but that is by design "off-band" (as an example both AWS and GCP require the authentication to the KMS to be done somehow: authentication data is not included in the Signer arguments). This approach means we don't constrain the signing system configurability with the variables we define.
> 
> Things like "region" might be argued to be in either group but they feel more "generic config": you might change the region your app uses at some point but you are unlikely to choose a different region for different keys

Makes sense. I will make experiments with environmental variables and config files.

securesystemslib/signer/_aws_signer.py
```diff
@@ -58,14 +60,22 @@ class AWSSigner(Signer):
...
     SCHEME = "awskms"
 
-    def __init__(self, aws_key_id: str, public_key: Key):
+    def __init__(
+        self,
+        aws_key_id: str,
+        public_key: Key,
+        endpoint_url: Optional[str] = None,
+        region_name: Optional[str] = None
+    ):
```
kairoaraujo, line 69:
```suggestion
    def __init__(
        self,
        aws_key_id: str,
        public_key: Key,
        **kwargs,
    ):
```

securesystemslib/signer/_aws_signer.py
```diff
@@ -83,7 +93,13 @@ def from_priv_key_uri(
         return cls(uri.path, public_key)
 
     @classmethod
-    def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
+    def import_(
+        cls,
+        aws_key_id: str,
+        local_scheme: str,
+        endpoint_url: Optional[str] = None,
+        region_name: Optional[str] = None
```
kairoaraujo, line 101:
```suggestion
    def import_(
        cls,
        aws_key_id: str,
        local_scheme: str,
        **kwargs,
```

securesystemslib/signer/_aws_signer.py
```diff
@@ -107,7 +125,9 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
         if AWS_IMPORT_ERROR:
             raise UnsupportedLibraryError(AWS_IMPORT_ERROR)
 
-        client = boto3.client("kms")
+        client = boto3.client(
+            "kms", endpoint_url=endpoint_url, region_name=region_name
+        )
```
kairoaraujo, line 130:
```suggestion
        client = boto3.client("kms", **kwargs)
```

securesystemslib/signer/_aws_signer.py
```diff
@@ -58,14 +60,22 @@ class AWSSigner(Signer):
...
         if AWS_IMPORT_ERROR:
             raise UnsupportedLibraryError(AWS_IMPORT_ERROR)
 
         self.hash_algorithm = self._get_hash_algorithm(public_key)
         self.aws_key_id = aws_key_id
         self.public_key = public_key
-        self.client = boto3.client("kms")
+        self.client = boto3.client(
+            "kms", endpoint_url=endpoint_url, region_name=region_name
+        )
```
kairoaraujo, line 78:
```suggestion
        self.client = boto3.client("kms", **kwargs)
```

securesystemslib/signer/_aws_signer.py
```diff
@@ -45,6 +45,8 @@ class AWSSigner(Signer):
     Arguments:
         aws_key_id (str): AWS KMS key ID or alias.
         public_key (Key): The related public key instance.
+        endpoint_url: (str): AWS endpoint URL.
+        region_name: (str): AWS region to use. Example: us-east-1.
```
kairoaraujo, line 49:
```suggestion
       **kwargs: keyword arguments of `boto3.session.Session.client`
```

securesystemslib/signer/_aws_signer.py
```diff
@@ -92,6 +108,8 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
         Arguments:
             aws_key_id (str): AWS KMS key ID.
             local_scheme (str): Local scheme to use.
+            endpoint_url: (str): AWS endpoint URL.
+            region_name: (str): AWS region to use. Example: us-east-1.
```
kairoaraujo, line 112:
```suggestion
            **kwargs: keyword arguments of `boto3.session.Session.client`
```

kairoaraujo (commented):
> * There might be a better option if these variables are more _application config_ than key specific secrets/variables: these systems often support environment variables and config files for something like this and looks like boto does too https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables . I think if documented, that's a fine way to support this with zero code in the library: this allows the application to configure every detail of AWS and not just the ones we expose... However, if these are key-specific variables (likely to be different if you use another AWS key), then I agree Signer attributes make sense

If you're building an application/platform that supports multiple AWS Accounts (for example, multiple account signing), using the environment variable/configuration files will be limited by the sslib.
Another scenario is that using environment variables requires you to restart your deployment or if you use configuration files need to guarantee that sslib instance is reloaded.

I suggest supporting all scenarios, env vars, configuration files, and [keyword arguments to the boto3](https://github.com/boto/boto3/blob/897f1ccbf64e2f6470fe0bd35f165bd50d3bd84f/boto3/session.py#L217), it is how the boto3 is designed as well.

jku:
That seems like a fairly maintainable approach so I certainly won't complain if you'd like to do this instead (even if I'm not a big fan of kwargs personally). Even using `botocore.config.Config` as an argument might be an option (although some annotation shenanigans might be needed in that case to allow importing without installing boto3).

So "sounds good to me" but I want to explain why I went with the design you see in current Signers:
* Using this kwargs approach  prevents getting any benefits of `from_priv_key_uri()` (because your signing key details now include arbitrary boto config which we can't possibly "serialize" into a private key uri).
* Allowing kwargs certainly makes config more flexible but the application now has to separately manage signing key details for every signing system it supports -- The `priv_key_uri` idea was intended to solve this issue be separating signing key specific config (that can be "serialized" into a private key uri) and application config (that is managed off-band). With signers that support `from_priv_key_uri` the application signing code can be completely generic and will work with all signers

Below is also just comments, not intended to take away from the "LGTM" overall comment:

> If you're building an application/platform that supports multiple AWS Accounts (for example, multiple account signing), using the environment variable/configuration files will be limited by the sslib.

I don't really think so -- I'm not suggesting defining any variables in Securesystemslib, I'm suggesting the application should use the generic (non-python specific) configuration interfaces provided by AWS. So if your application does want to use multiple accounts, it can modify the environment variables whenever it wants.

> Another scenario is that using environment variables requires you to restart your deployment or if you use configuration files need to guarantee that sslib instance is reloaded.

I really would assume you can modify the `AWS_*` environment variables whenever you want and the next boto invocation will react appropriately.

kairoaraujo:
> That seems like a fairly maintainable approach so I certainly won't complain if you'd like to do this instead (even if I'm not a big fan of kwargs personally). Even using `botocore.config.Config` as an argument might be an option (although some annotation shenanigans might be needed in that case to allow importing without installing boto3).

My comments were just general suggestion.
I also thought about using `botocore.config.Config` as an option, I just chose `**kwargs` to be simpler and also support `config` as well, but yes it can bring a breaking design in the `Signer`.

> So "sounds good to me" but I want to explain why I went with the design you see in current Signers:
> 
> * Using this kwargs approach  prevents getting any benefits of `from_priv_key_uri()` (because your signing key details now include arbitrary boto config which we can't possibly "serialize" into a private key uri).
> * Allowing kwargs certainly makes config more flexible but the application now has to separately manage signing key details for every signing system it supports -- The `priv_key_uri` idea was intended to solve this issue be separating signing key specific config (that can be "serialized" into a private key uri) and application config (that is managed off-band). With signers that support `from_priv_key_uri` the application signing code can be completely generic and will work with all signers

Thank @jku for the details.
I think implementing the parsing on `from_priv_key_uri()` is also a good solution. 
It needs a bit more maintenance but it is consistent with the design you described.


> I don't really think so -- I'm not suggesting defining any variables in Securesystemslib, I'm suggesting the application should use the generic (non-python specific) configuration interfaces provided by AWS. So if your application does want to use multiple accounts, it can modify the environment variables whenever it wants.

> I really would assume you can modify the `AWS_*` environment variables via python whenever you want and the next boto invocation will react appropriately -- but maybe I'm wrong here.

It is a solution but IMHO using environment variables with dynamic values can be confusing in case of concurrent/multiple instances. It can be a bit confusing :(

lukpueh:
FYI, Providing credentials ambiently does work pretty nicely: https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/452

I suggest to not change AWS Signer until there is a concrete use case.

Issue #631, opened 2023-08-30
Release v0.29.0


MVrachev:
**Description of issue or feature request**:
A new version of securesystemslib was released 4 months ago on 18-th of April.
There are many changes from then up to today:
https://github.com/secure-systems-lab/securesystemslib/compare/v0.28.0...main

**Is there anything  that stops the release @jku @lukpueh?** 

I will help the maintainers by submitting a possible `CHANGELOG` as suggested by @jku.

### Added
* Add `SigstoreKey` and `SigstoreSigner` #522.
* Add `SpxKey` and  `SpxSigner` #568.
* Add `AzureSigner` #588.
* Add `AWS KMS Signer` #609.
* Add `CryptoSigner` as a replacement for `SSLibSigner`. #604

### Changed
* Moved `spx keygen` to `securesystemslib/signer/_spx_signer.py` #568
* Use `sigstore` v. 1.1.2 #583 
* Move signature verification for "securesystemslib keys" inside `SSlibKey`. #585

### Removed
* Support for spx keygen. #568.

### Depricated
* SSlibSigner and sslib keydict #604

### Fixed
* Fix CI permission error #578

jku:
release is a good idea.

potential blockers:
* #624, #622 
* #630 -- we should not treat this as hard blocker as the sigstore release has not happened but recent commits look like release activity :) and I'd really like to have decent sigstore support
* anything else?

jku:
> Add SpxKey and SpxSigner https://github.com/secure-systems-lab/securesystemslib/pull/568

Hmm, the signer is not enabled by default but it's also not documented as unstable or experimental... I think it probably should: I'm thinking mostly about the key and signature container formats: they really should be documented in a TAP or something and I don't think they are currently

jku:
I've made #632 to address the stability status

MVrachev:
I approved it, but I am not a maintainer :D

jku:
current blocker list
[x] #622
[x] #634
[] #632
[] #630 (soft block only: let's not block if sigstore has not released)

PR #632, opened 2023-09-01
Sigstore, Spx: Add notes about metadata format stability

jku:
Both of these metadata formats (e.g. the data encoding and field names) are basically invented in securesystemslib: there is no community consensus on them yet.

I've also tweaked the sigstore docstring to be more exact

EDIT: I noticed SpxKey was part of the default key set: I'd rather err on the side of caution and not enable new keys automatically if we're not sure there is a consensus about the metadata format: I removed the key from default set

### DIFF
--- a/securesystemslib/signer/_sigstore_signer.py
+++ b/securesystemslib/signer/_sigstore_signer.py
@@ -28,10 +28,9 @@ logger = logging.getLogger(__name__)
 class SigstoreKey(Key):
     """Sigstore verifier.
 
-    NOTE: The Sigstore key and signature metadata formats are not part of the
-    TUF specification and are not considered stable in securesystemslib. They
-    may change in future releases and may not be supported by other
-    implementations.
+    NOTE: The Sigstore key and signature serialization formats are not yet
+    considered stable in securesystemslib. They may change in future releases
+    and may not be supported by other implementations.
     """
 
     DEFAULT_KEY_TYPE = "sigstore-oidc"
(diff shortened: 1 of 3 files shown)


securesystemslib/signer/_sigstore_signer.py
```diff
@@ -28,7 +28,10 @@
 class SigstoreKey(Key):
     """Sigstore verifier.
 
-    NOTE: unstable API - routines and metadata formats may change!
+    NOTE: The Sigstore key and signature metadata formats are not part of the
+    TUF specification and are not considered stable in securesystemslib. They
```
jku, line 32: maybe I shouldn't have specifically named TUF here but I guess the idea is clear: the formats are not "standardized" anywhere

MVrachev (approved):

jku:
SpxKey is currently used by the default as well... as in a client will accept SPHINCS keys by default right now. I'm tempted to remove that so that it's comparable to SigstoreKey: both client and repository would explicitly have to enable SPHINCS for those keys to work.

jku:
I can't actually see who the maintainers are (GitHub is a bit silly in that) but I believe at least @adityasaky and @mnm678: could either of you review this?

this is preparation for #631

securesystemslib/signer/_sigstore_signer.py
adityasaky: I'd include that Sigstore is uncertain until TAP-18 is accepted into the TUF spec. We don't have an in-toto spec for Sigstore keys specifically but I'm also of the opinion it doesn't entirely belong in the spec but rather in implementation docs. The spec / ITE only needs to identify the parameters that must be used to match the sigstore identity without worrying about other details.

There's generally more work we need to do unify the sslib specific documentation across in-toto and TUF, perhaps this is the home for it and both projects point here for what's supported once the signer interface is accepted.

jku: I don't think new key formats need to be added in the spec: even TAP-18 has no reason to be part of the spec as far as I'm concerned.  I am saying that adding support for new keys in a single implementation without clear community consensus is something we should be cautious about: I'm worried about both feature disparity and plain incompatibilities between different implementations of TUF (or intoto), e.g. python-tuf vs go-tuf.

I can try rephrasing this if it helps.

jku: I've removed the mention of TUF spec here

adityasaky (approved):
LGTM, thanks!

PR #635, opened 2023-09-05
Release 0.29.0

jku:
Release new securesystemslib
* Fixes: #631
* Last release was in April, we now support loads of new signing systems so downstream projects are waiting for a release
* I've never done a securesystems release: comments are welcome
* changelog is edited from Martins proposal

I really wanted the new sigstore API (#630) in here but I guess that will be in the next release as sigstore still has blockers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,29 @@
 # Changelog
 
+## securesystemslib v0.29.0
+
+This release is reaping the rewards of the new signer API with four(!) new
+signing methods: Two cloud based KMSs, post-quantum crypto support and a
+"keyless" signing system.
+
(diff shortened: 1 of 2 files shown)


adityasaky (approved):
This looks good to me but I also haven't released sslib in the past.

For in-toto, I imagine we rework https://github.com/in-toto/in-toto/pull/612 to go straight to the new `CryptoSigner` (and others), correct?

Edit: I think it might be a future change as we're already users of SSLibSigner and we want to start deprecating the ad-hoc parameters for signing key, GPG etc. I do see the tests have started using `CryptoSigner` so that PR should be unblocked with this release of sslib!

mnm678 (approved):
I have also not done a previous release, so I might be missing something but this LGTM

PR #644, opened 2023-09-28
Drop Python 3.7 support

lukpueh:
- Python 3.7 is EOL
- Our runtime dependencies are still ok with 3.7
- Testing dependencies have started requiring 3.8
- Our dependent python-tuf just dropped 3.7 (theupdateframework/python-tuf#2460)

Unrelated change: Add missing classifier for already supported 3.11 to pyproject.toml

h/t @jku


jku (approved):
LGTM. I assume you'll make sure in-toto does not get surprised

Issue #309, opened 2020-12-09
Revise stand-alone on-disk key format


lukpueh:
**Description of issue or feature request**:
#251 discusses various consistency issues related to securesystemslib (TUF/in-toto) cryptographic keys. This issue splits out the concerns related to stand-alone on-disk key formats:

For some key types we support serialization and deserialization of standard formats, whereas for others we use a custom format (see  'Current behavior' for details), which results in an inconsistent API. E.g. the custom format includes a signing scheme, which may be assigned at key creation time, whereas standard format keys don't include that scheme and thus need to have it assigned on import.

**Current behavior**:
`securesystemslib` currently provides functions to...
- generate, export, import rsa public keys in X.509 SubjectPublicKeyInfo PEM format (RFC5280)
- generate, export, import (optionally encrypted) rsa private keys in PKCS#1/TraditionalOpenSSL PEM format (RFC8017)
- generate, export, import ed25519, ecdsa public keys and optionally encrypted private keys in a custom json format similar to "public key metadata format" (see #308)
- import rsa, dsa and eddsa public keys from GnuPG keyrings

**Expected behavior**:

Make on-disk key formats consistent for all key types (standard vs. custom, signing scheme included vs. not included). Preferably we support standard formats for all types. This would allow end-users to generate keys out-of-band and use securesystemslib only to import public keys and convert them to the in-toto/tuf metadata format (see https://github.com/secure-systems-lab/securesystemslib/issues/308 ), and similarly to import private keys to generate signatures in the in-toto/tuf metadata format.

This has the advantage that users don't need to trust our key generation functionality (standard crypto is better than custom!).

Furthermore, if users don't depend on our key generation code, we could remove or limit it to very thin wrappers around well-established tools such as `openssh-keygen`, `openssl`, `gpg`, etc. to be used mostly internally (e.g. for tests and demos), which would be a huge maintenance relief.

PR #658, opened 2023-10-16
add migrate_key cli script

lukpueh:
This is in preparation for the removal of legacy key modules and formats, in favor of the new Signer API. It allows users to convert their old rsa, ed25519 and ecdsa key files, generated with the `interface` or `keys` module, and using an outdated standard or sslib proprietary format (see #309), to a consistent new standard format, which can be used with the file-based signer (`CryptoSigner`) of the new Signer API.

NOTE: The script uses legacy code and should thus be removed with them, from the repo tree, while remaining available to users of securesystemslib for some time. We could keep pointing to it in docs after its removal (users would need to check-out the repo at a specified tag), or move it to a different place.

**Change details**
* Add cli script to convert key files.
* Add private/private encrypted/public test key files for each supported algorithm in the legacy format. *The key pairs were generated with `interface`, minimally modified to allow writing an encrypted and non-encrypted version of the same private key.*
* Add comprehensive tests, includes backwards/forward compatibility of signatures.


jku:
I don't object to the functionality but I have some questions:
* does this need to be part of the API? does it need to be part of the wheel or could it just be in git tree without getting packaged?
* if it's a CLI script and should be part of the wheel, should it be packaged as `script` or `entry_point` (and so the module could be internal)

I'm a little surprised I can't get mypy to complain about the lack of typing (even if I add the file to the config) but I don't know if that's something you should spend time on...

lukpueh:
> * does this need to be part of the API? does it need to be part of the wheel or could it just be in git tree without getting packaged?

Git tree only is fine. It certainly does not have to be part of the public API/CLI, or even packaged. I really just wanted to add it to the "securesystemslib" namespace, piggyback on its ci, and show that it passes at least once.

My idea was to tag it, remove it, but keep it in the docs for a while, including the tag at which it can be used.

In reality, the script is unlikely to have a lot of users, and it's simple enough, so I might as well just throw it in a personal gist. What do you think?

jku:
I don't have strong opinions, any of those sounds fine. Could drop it in docs/ or examples/ as well :shrug: .

lukpueh:
@jku, would you kindly approve?

I moved the script to docs/ and added a small pointer to README.

jku (approved):

Issue #651, opened 2023-10-03
ed25519 workflow is broken


jku:
There's a workflow that tries to check that ed25519 git tree has not changed. The workflow seems broken:

```
2023-10-02T13:13:33.3812667Z ed25519 upstream main has changed!
2023-10-02T13:13:33.3823451Z Expected [main](https://github.com/pyca/ed25519/commits/main) to be commit f98354b, found 11b3d81 instead ([diff](https://github.com/pyca/ed25519/compare/f98354b09ffde8444a04c71f8e3ac7b948c89a02...main)).
2023-10-02T13:13:33.3873615Z TypeError: Cannot read properties of undefined (reading 'issuesAndPullRequests')
2023-10-02T13:13:33.3874991Z     at eval (eval at callAsyncFunction (/home/runner/work/_actions/actions/github-script/d7906e4ad0b1822421a7e6a35d5ca353c962f410/dist/index.js:15143:16), <anonymous>:6:36)
2023-10-02T13:13:33.3876232Z     at callAsyncFunction (/home/runner/work/_actions/actions/github-script/d7906e4ad0b1822421a7e6a35d5ca353c962f410/dist/index.js:15144:12)
2023-10-02T13:13:33.3877293Z     at main (/home/runner/work/_actions/actions/github-script/d7906e4ad0b1822421a7e6a35d5ca353c962f410/dist/index.js:15236:26)
2023-10-02T13:13:33.3878159Z     at /home/runner/work/_actions/actions/github-script/d7906e4ad0b1822421a7e6a35d5ca353c962f410/dist/index.js:15217:1
2023-10-02T13:13:33.3879251Z     at /home/runner/work/_actions/actions/github-script/d7906e4ad0b1822421a7e6a35d5ca353c962f410/dist/index.js:15268:3
2023-10-02T13:13:33.3880584Z     at Object.<anonymous> (/home/runner/work/_actions/actions/github-script/d7906e4ad0b1822421a7e6a35d5ca353c962f410/dist/index.js:15271:12)
2023-10-02T13:13:33.3881161Z     at Module._compile (node:internal/modules/cjs/loader:1198:14)
2023-10-02T13:13:33.3881597Z     at Object.Module._extensions..js (node:internal/modules/cjs/loader:1252:10)
2023-10-02T13:13:33.3882024Z     at Module.load (node:internal/modules/cjs/loader:1076:32)
[...]

PR #666, opened 2023-10-25
Fix ed25519 upstream check

jku:
Fixes: #651 

Upgrading to github-script v5 broke this (should have read the changelog more carefully...) but the fix seems to be easy


lukpueh:
Sorry, I missed this PR. @jku, do you mind fixing the conflicts? Then I'll review right away.

jku:
I removed the improvements to uses lines, should be a clean merge now

lukpueh (approved):
Cross-checked with [Changelog](https://github.com/actions/github-script#v5). Should be fine.

jku:
this does seem fixed but the workflow still fails with 'Resource not accessible by integration' when it tries to create an issue... I wonder if some setting was changed meanwhile?

jku:
oh it was already broken since commit d082614d4d: permssions were set to none, which prevents filing an issue when needed

PR #672, opened 2023-11-08
Prepare Debian release 0.28.0-1

lukpueh:
* Update downstream files
* Add Dockerfile to make future downstream packaging easier See doc header for usage details.

NOTE: Can't package latest upstream version (0.30.0), because of missing pyca/cryptography version in Debian.

Package files have been uploaded to mentors: https://mentors.debian.net/package/python-securesystemslib/


lukpueh:
NOTE: This branch will not be merged into `main` but into `debian` instead. This is because it is currently based off `v0.28.0` so that the source tree matches the `debian/*` files for this downstream release. It is okay that it  lags behind `main` and as a consequence does not pass all checks (e.g. docs).

lukpueh:
Released: https://tracker.debian.org/news/1479525/python-securesystemslib-0280-1-migrated-to-testing/

Issue #679, opened 2023-11-23
fix permisssions on check-upstream-ed25519 workflow


jku:
https://github.com/secure-systems-lab/securesystemslib/actions/runs/6969907734/job/18966770592

github.rest.issues.create() fails with  `Error: Unhandled error: HttpError: Resource not accessible by integration`

The job in .github/workflows/check-upstream-ed25519.yml that files the issue needs permissions to do that.

PR #706, opened 2024-01-05
github: Fix check-upgstream workflow permission

jku:
I believe this broke in commit d082614 already, I think "issues" is the only permission we need here.

Fixes: #679


lukpueh (approved):
Thanks, let's try again.

PR #711, opened 2024-01-09
CryptoSigner: Support ecdsa keytype that is no longer in spec

jku:
We may be signing older metadata that has keys with keytype "ecdsa-sha2-nistp256": support that case.

I may have missed or forgotten plans about ECDSA, feel free to remind me... The related question here would be: the only supported scheme seems to be "ecdsa-sha2-nistp256", even though SSlibKey supports "ecdsa-sha2-nistp384" as well. Should that be added to CryptoSigner for balance?


tests/test_signer.py
```diff
@@ -791,16 +796,22 @@ def test_from_priv_key_uri(self):
         SIGNER_FOR_URI_SCHEME[CryptoSigner.FILE_URI_SCHEME] = CryptoSigner
 
         for keytype, scheme, public_key_value in test_data:
+            if keytype == "ecdsa-sha2-nistp256":
+                # special case for keytype that is deprecated in spec
+                keytype_for_filename = "ecdsa"
+            else:
+                keytype_for_filename = keytype
```
jku, line 803: Maybe I should just add the filename into the test data structure

lukpueh (approved):
👍

Issue #612, opened 2023-08-01
signer: test AWS signer with LocalStack


lukpueh:
#609 adds an AWS signer including a test case (`check_aws_signer.py`), which is currently not part of our CI. @kairoaraujo suggested to try out [LocalStack](https://docs.localstack.cloud/user-guide/aws/kms/).

lukpueh:
cc @ianhundere

lukpueh:
Can be copied from https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/452/commits/c4b26c0178ed95c3f5981cdea89863f10c8387a5

kairoaraujo:
I will take this one

lukpueh:
> I will take this one

I took it ... after coordinating with Kairo :) 

👉  #777

PR #724, opened 2024-01-30
refactors AWSSigner; scheme no longer required.

### LINKED ISSUE #668: Does `AWSSigner.import_()` really require scheme string?
I imagined there would be two good ways to implement `import_()` style methods:
* either ask caller for a key identifier and then request the key details from external service
* or ask caller for all key details

Both ways have some pros and cons.... AWSSigner seems to be half-way between these: it seems to require being logged in to AWS but still requires the user to provide a scheme string.

I would expect that the scheme is in some fashion provided in the 
 `boto3_client.get_public_key()` return value, so we could improve `import_()` so that user would not actually have to provide the scheme. Does this sound correct?

**Current behavior**:

` def import_(cls, aws_key_id: str, local_scheme: str)`

**Expected behavior**:

` def import_(cls, aws_key_id: str)`
(or maybe leave local_scheme as optional for backwards compat)

cc @ianhundere

ianhundere:
resolves #668

This refactors `AWSSigner` to no longer require `local_scheme`, and instead, gets `scheme` from AWS' response (e.g. `request["SigningAlgorithms"]`).

**ecdsa**:
```
"SigningAlgorithms": [
    "ECDSA_SHA_256"
]
```

The caveat is that RSA keys have multiple `SigningAlgorithms`, so if `local_scheme` is not provided and the `keytype` is rsa then `rsassa-pss-sha256` is chosen by default as is the case throughout `securesystemslib`.

**rsa**:
```
"SigningAlgorithms": [
    "RSASSA_PKCS1_V1_5_SHA_256",
    "RSASSA_PKCS1_V1_5_SHA_384",
    "RSASSA_PKCS1_V1_5_SHA_512",
    "RSASSA_PSS_SHA_256",
    "RSASSA_PSS_SHA_384",
    "RSASSA_PSS_SHA_512"
]
```

Also includes updated docs and error handling.

cc @jku @lukpueh

### DIFF
--- a/securesystemslib/signer/_aws_signer.py
+++ b/securesystemslib/signer/_aws_signer.py
@@ -123,10 +123,7 @@ class AWSSigner(Signer):
             raise UnsupportedLibraryError(AWS_IMPORT_ERROR)
 
         try:
-            if (
-                local_scheme
-                and local_scheme in cls.aws_signing_algorithms.keys()
-            ):
+            if local_scheme and local_scheme in cls.aws_signing_algorithms:
                 scheme = local_scheme
         except KeyError as e:
             raise ValueError(f"Unsupported scheme: {local_scheme}") from e
--- a/securesystemslib/signer/_aws_signer.py
+++ b/securesystemslib/signer/_aws_signer.py
@@ -147,21 +147,6 @@ class AWSSigner(Signer):
             else:
                 raise ValueError(f"Unsupported key type: {keytype}")
 
-        if keytype == "ecdsa" and local_scheme != cls._get_ecdsa_scheme(
-            aws_scheme
-        ):
-            raise ValueError(
-                f"The AWS KMS key does not support the scheme: {local_scheme}"
-            )
-
-        if (
-            keytype == "rsa"
-            and local_scheme not in cls.aws_signing_algorithms.values()
-        ):
-            raise ValueError(
-                f"The AWS KMS key does not support the scheme: {local_scheme}"
-            )
-
         kms_pubkey = serialization.load_der_public_key(request["PublicKey"])
 
         public_key_pem = kms_pubkey.public_bytes(
--- a/securesystemslib/signer/_aws_signer.py
+++ b/securesystemslib/signer/_aws_signer.py
@@ -148,6 +148,22 @@ class AWSSigner(Signer):
                 local_scheme = "rsassa-pss-sha256"
             else:
                 raise ValueError(f"Unsupported key type: {keytype}")
+        else:
+            aws_scheme = request["SigningAlgorithms"][0]
+            if (
+                keytype == "ecdsa"
+                and cls._get_ecdsa_scheme(aws_scheme) != local_scheme
+            ):
+                raise ValueError(
+                    f"The AWS KMS key does not support the scheme: {local_scheme}"
+                )
+            elif (
+                keytype == "rsa"
+                and local_scheme not in cls.aws_signing_algorithms.values()
+            ):
+                raise ValueError(
+                    f"The AWS KMS key does not support the scheme: {local_scheme}"
+                )
 
         public_key_pem = kms_pubkey.public_bytes(
             encoding=serialization.Encoding.PEM,
--- a/securesystemslib/signer/_aws_signer.py
+++ b/securesystemslib/signer/_aws_signer.py
@@ -122,9 +122,8 @@ class AWSSigner(Signer):
         if AWS_IMPORT_ERROR:
             raise UnsupportedLibraryError(AWS_IMPORT_ERROR)
 
-        if local_scheme:
-            if local_scheme not in cls.aws_signing_algorithms:
-                raise ValueError(f"Unsupported scheme: {local_scheme}")
+        if local_scheme and local_scheme not in cls.aws_signing_algorithms:
+            raise ValueError(f"Unsupported scheme: {local_scheme}")
 
         try:
             client = boto3.client("kms")
@@ -140,21 +139,6 @@ class AWSSigner(Signer):
         keytype = cls._get_keytype_from_aws_response(request)
         aws_scheme = request["SigningAlgorithms"][0]
 
-        if (
-            keytype == "ecdsa"
-            and cls._get_ecdsa_scheme(aws_scheme) != local_scheme
-        ):
-            raise ValueError(
-                f"The AWS KMS key does not support the scheme: {local_scheme}"
-            )
-        if (
-            keytype == "rsa"
-            and local_scheme not in cls.aws_signing_algorithms.values()
-        ):
-            raise ValueError(
-                f"The AWS KMS key does not support the scheme: {local_scheme}"
-            )
-
         if not local_scheme:
             if keytype == "ecdsa":
                 local_scheme = cls._get_ecdsa_scheme(aws_scheme)
@@ -162,6 +146,21 @@ class AWSSigner(Signer):
                 local_scheme = "rsassa-pss-sha256"
             else:
                 raise ValueError(f"Unsupported key type: {keytype}")
+
+        if keytype == "ecdsa" and local_scheme != cls._get_ecdsa_scheme(
+            aws_scheme
+        ):
+            raise ValueError(
+                f"The AWS KMS key does not support the scheme: {local_scheme}"
+            )
+
+        if (
+            keytype == "rsa"
+            and local_scheme not in cls.aws_signing_algorithms.values()
+        ):
+            raise ValueError(
+                f"The AWS KMS key does not support the scheme: {local_scheme}"
+            )
 
         kms_pubkey = serialization.load_der_public_key(request["PublicKey"])
 
--- a/securesystemslib/signer/_aws_signer.py
+++ b/securesystemslib/signer/_aws_signer.py
@@ -122,12 +122,9 @@ class AWSSigner(Signer):
         if AWS_IMPORT_ERROR:
             raise UnsupportedLibraryError(AWS_IMPORT_ERROR)
 
-        try:
-            if local_scheme:
-                if local_scheme not in cls.aws_signing_algorithms:
-                    raise ValueError(f"Unsupported scheme: {local_scheme}")
-        except KeyError as e:
-            raise ValueError(f"Unsupported scheme: {local_scheme}") from e
+        if local_scheme:
+            if local_scheme not in cls.aws_signing_algorithms:
+                raise ValueError(f"Unsupported scheme: {local_scheme}")
 
         try:
             client = boto3.client("kms")
--- a/securesystemslib/signer/_aws_signer.py
+++ b/securesystemslib/signer/_aws_signer.py
@@ -122,10 +122,6 @@ class AWSSigner(Signer):
         if AWS_IMPORT_ERROR:
             raise UnsupportedLibraryError(AWS_IMPORT_ERROR)
 
-        if local_scheme:
-            if local_scheme not in cls.aws_signing_algorithms:
-                raise ValueError(f"Unsupported scheme: {local_scheme}")
-
         try:
             client = boto3.client("kms")
             request = client.get_public_key(KeyId=aws_key_id)
@@ -139,16 +135,18 @@ class AWSSigner(Signer):
 
         kms_pubkey = serialization.load_der_public_key(request["PublicKey"])
         keytype = cls._get_keytype_from_aws_response(request)
+        aws_scheme = request["SigningAlgorithms"][0]
 
         if not local_scheme:
             if keytype == "ecdsa":
-                aws_scheme = request["SigningAlgorithms"][0]
                 local_scheme = cls._get_ecdsa_scheme(aws_scheme)
             elif keytype == "rsa":
                 local_scheme = "rsassa-pss-sha256"
             else:
                 raise ValueError(f"Unsupported key type: {keytype}")
         else:
+            if local_scheme not in cls.aws_signing_algorithms:
+                raise ValueError(f"Unsupported scheme: {local_scheme}")
             aws_scheme = request["SigningAlgorithms"][0]
             if (
                 keytype == "ecdsa"
--- a/securesystemslib/signer/_aws_signer.py
+++ b/securesystemslib/signer/_aws_signer.py
@@ -143,9 +143,9 @@ class AWSSigner(Signer):
         if not local_scheme:
             if keytype == "ecdsa":
                 aws_scheme = request["SigningAlgorithms"][0]
-                scheme = cls._get_ecdsa_scheme(aws_scheme)
+                local_scheme = cls._get_ecdsa_scheme(aws_scheme)
             elif keytype == "rsa":
-                scheme = "rsassa-pss-sha256"
+                local_scheme = "rsassa-pss-sha256"
             else:
                 raise ValueError(f"Unsupported key type: {keytype}")
 
@@ -155,8 +155,8 @@ class AWSSigner(Signer):
         ).decode("utf-8")
 
         keyval = {"public": public_key_pem}
-        keyid = compute_default_keyid(keytype, scheme, keyval)
-        public_key = SSlibKey(keyid, keytype, scheme, keyval)
+        keyid = compute_default_keyid(keytype, local_scheme, keyval)
+        public_key = SSlibKey(keyid, keytype, local_scheme, keyval)
         return f"{cls.SCHEME}:{aws_key_id}", public_key
 
     @staticmethod
--- a/securesystemslib/signer/_aws_signer.py
+++ b/securesystemslib/signer/_aws_signer.py
@@ -122,6 +122,10 @@ class AWSSigner(Signer):
         if AWS_IMPORT_ERROR:
             raise UnsupportedLibraryError(AWS_IMPORT_ERROR)
 
+        if local_scheme:
+            if local_scheme not in cls.aws_signing_algorithms:
+                raise ValueError(f"Unsupported scheme: {local_scheme}")
+
         try:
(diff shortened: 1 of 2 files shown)


securesystemslib/signer/_aws_signer.py
```diff
@@ -203,32 +219,60 @@ def _get_hash_algorithm(public_key: Key) -> str:
...
-            Signature.
+            Signature: A signature object containing the key ID and the signature.
         """
         try:
-            request = self.client.sign(
+            pubkey_request = self.client.get_public_key(KeyId=self.aws_key_id)
+            keytype = self._get_keytype_from_aws_response(pubkey_request)
+            signing_algo = self._determine_signing_algorithm(
+                keytype, pubkey_request
+            )
```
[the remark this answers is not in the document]
ianhundere: ah, thanks for catching that. i definitely was overthinking it. 😅

securesystemslib/signer/_aws_signer.py
```diff
@@ -206,18 +229,20 @@ def _get_hash_algorithm(public_key: Key) -> str:
     def sign(self, payload: bytes) -> Signature:
         """Sign the payload with the AWS KMS key
 
+        This method sends the payload to AWS KMS, where it is signed using the specified
+        key and algorithm using the raw message type.
```
jku, line 233: Good addition, thanks: I had not realized this is not prehashed.

This is not for this PR but just as a note:
* Users probably expect prehashing (instead of sending potentially large files to AWS)
* Confusingly this method does prehash the content... the result is just not used

jku (approved):
Thanks. I think this is good to merge: it improves `import_()` in that it can be used without the other argument.


I'll leave some notes below but I am **not** requesting changes: I have no access to AWS signer so I can't really tell what is smart and what is not.

So just as a comment: the lookup (from-AWS-response-to-securesystemslib-key-type-and-scheme) still looks more complicated than it probably is 
* RSA users can now choose the hash length but ECDSA users for some reason cannot?
* the real import_() argument probably should have been `hash_size: int = 256` if that's the only real variable. This would be an API change for AWSSigner
* then the single lookup function  would be `def _get_keytype_and_scheme(aws_response, hash_size) -> Tuple[str, str]`

lukpueh (changes requested):
Sorry for commenting late. I am not a big fan of the new key-type specific argument. It breaks the API and feels inconsistent.

I suggest to keep the `local_scheme`, make it optional, and use it like so:
```
get key from aws by keyid

if scheme is passed:
  if scheme is not allowed for key
    raise

else:
  assign default scheme for key
```

I don't know exactly what schemes are supported for what key, but I expect that for any rsa key in the kms you should be able to use any of the securesystemslib schemes, whereas for ecdsa there is only one scheme per key.

A very brief look at the [docs](https://docs.aws.amazon.com/kms/latest/developerguide/asymmetric-key-specs.html#key-spec-rsa) seems to support my assumption.

ianhundere:
@jku @lukpueh thanks for y'all's input, i've gone ahead and renamed `rsa-scheme` back to `local_scheme`.

securesystemslib/signer/_aws_signer.py
```diff
@@ -101,65 +104,85 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
...
-                f"{local_scheme} is not a supported signing algorithm"
-            ) from e
+            client = boto3.client("kms")
+            request = client.get_public_key(KeyId=aws_key_id)
+            kms_pubkey = serialization.load_der_public_key(request["PublicKey"])
+            keytype = cls._get_keytype_from_aws_response(request)
+
+            if keytype == "ecdsa":
+                aws_scheme = request["SigningAlgorithms"][0]
+                scheme = cls._get_ecdsa_scheme(aws_scheme)
```
lukpueh, line 121: It looks like `local_scheme` is ignored now, if the key type is ecdsa. I think we should fail, if the user passes an unsupported scheme.

securesystemslib/signer/_aws_signer.py
lukpueh: > Confusingly this method does prehash the content... the result is just not used

If you are not sending the prehashed content, I suggest to  remove the code that hashes the content. This means you no longer need `self.hash_algorithm`, which means that you no longer need `self._get_aws_signing_algo` (except maybe for catching unsupported schemes).

securesystemslib/signer/_aws_signer.py
```diff
@@ -101,65 +104,85 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
...
+            elif keytype == "rsa":
+                scheme = local_scheme if local_scheme else "rsassa-pss-sha256"
```
lukpueh, line 123: Same here. If `local_scheme` is unsupported for the rsa key you won't fail before calling `_get_aws_signing_algo` for the constructor, which is actually an unnecessary operation (see comment above in `sign`).

What do you think about the following strategy, if `local_scheme` is passed:
1. get aws representation (e.g. RSASSA_PSS_SHA_256) for passed securesystemslib-style scheme (e.g. rsassa-pss-sha256)
2. fail if there it doesn't translate 
3. fail if the scheme is not in the returned SigningAlgorithms

lukpueh (changes requested):

lukpueh:
fyi @ianhundere, I've tested AWSSigner with Localstack for the first time today and it works like a charm! 🚀 ❤️ 
https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/452

Thanks for your efforts!

securesystemslib/signer/_aws_signer.py
ianhundere: sounds good, i'm on holiday but will implement your feedback when i'm back.

securesystemslib/signer/_aws_signer.py
```diff
@@ -101,65 +102,95 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
...
-                f"{local_scheme} is not a supported signing algorithm"
-            ) from e
+            client = boto3.client("kms")
+            request = client.get_public_key(KeyId=aws_key_id)
+            kms_pubkey = serialization.load_der_public_key(request["PublicKey"])
+            keytype = cls._get_keytype_from_aws_response(request)
+
+            if local_scheme:
+                try:
+                    scheme = cls._get_aws_signing_algo(local_scheme)
```
lukpueh, line 119: This is the wrong scheme to be used for SSlibKey below.

I ran my [localstack test from RSTUF](https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/452) against your PR and it failed with this error message:

```
.tox/local-aws-kms/lib/python3.10/site-packages/securesystemslib/signer/_aws_signer.py:67: in __init__
    self.aws_algo = self._get_aws_signing_algo(self.public_key.scheme)
...
E       KeyError: 'RSASSA_PSS_SHA_256'
```

securesystemslib/signer/_aws_signer.py
```diff
@@ -101,65 +102,95 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
...
+                except KeyError as e:
+                    raise ValueError(
```
lukpueh, line 121: nit: since you raise here without having looked at the fetched key, you could consider doing the check before fetching the key.

securesystemslib/signer/_aws_signer.py
```diff
@@ -101,65 +102,95 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
...
+                        f"Unsupported scheme: {local_scheme}"
+                    ) from e
+            else:
+                if keytype == "ecdsa":
+                    aws_scheme = request["SigningAlgorithms"][0]
+                    scheme = cls._get_ecdsa_scheme(aws_scheme)
+                elif keytype == "rsa":
+                    scheme = "rsassa-pss-sha256"
```
lukpueh, line 129: Same as above, you could check for this case too, if the fetched key actually supports the default scheme.

lukpueh (changes requested):
Thanks for the updates! There is one small bug in `import_` which needs to be fixed before we can merge this. The rest is polishing.

securesystemslib/signer/_aws_signer.py
lukpueh: It doesn't look like you fixed this bug. When I pass `"rsassa-pss-sha256"` to `import_`, it returns an `SSlibKey` with  `"RSASSA_PSS_SHA_256"` as `scheme`. Later, when I pass that key to  `AWSigner.__init__`, I get the above error.

securesystemslib/signer/_aws_signer.py
lukpueh: >  you might also want to check, if the fetched key actually supports the passed scheme

Did you decide to not check it? I'm fine with that, maybe we can write an issue.

securesystemslib/signer/_aws_signer.py
lukpueh: see https://github.com/secure-systems-lab/securesystemslib/pull/724#discussion_r1526029353

securesystemslib/signer/_aws_signer.py
ianhundere: doh, rushed it yesterday.

securesystemslib/signer/_aws_signer.py
ianhundere: ah, thanks for the heads up. i think that's fixed now.

securesystemslib/signer/_aws_signer.py
```diff
@@ -101,136 +116,119 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
...
         if AWS_IMPORT_ERROR:
             raise UnsupportedLibraryError(AWS_IMPORT_ERROR)
 
-        client = boto3.client("kms")
-        request = client.get_public_key(KeyId=aws_key_id)
+        try:
+            if local_scheme and local_scheme in cls.aws_signing_algorithms:
+                scheme = local_scheme
+        except KeyError as e:
+            raise ValueError(f"Unsupported scheme: {local_scheme}") from e
```
lukpueh, line 129: You are mixing EAFP and LBYL here with the consequence that your ValueError will never be raised. :) 

I suggest to write:

```suggestion
        if local_scheme:
            if local_scheme not in cls.aws_signing_algorithms:
                raise ValueError(f"Unsupported scheme: {local_scheme}") from e
                
        # no need to rename `local_scheme`, you can just use this name below
```

lukpueh (commented):
Thanks for your persistence! It looks like you are still not consistently checking, if the key supports the chosen scheme according to `SigningAlgorithms` in the response.

As I said, I'm okay, if we improve that in a follow-up PR. 

Please address my inline comment and we can merge this (or let me know, if I should address it).

securesystemslib/signer/_aws_signer.py
ianhundere: ah, thank you.

lukpueh: Very welcome!

Sorry if my comment about `local_scheme` was unclear. If you don't rename it here, you will need to use `local_scheme` below. See my code suggestions...

securesystemslib/signer/_aws_signer.py
```diff
@@ -101,136 +116,120 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
...
+
         kms_pubkey = serialization.load_der_public_key(request["PublicKey"])
+        keytype = cls._get_keytype_from_aws_response(request)
+
+        if not local_scheme:
+            if keytype == "ecdsa":
+                aws_scheme = request["SigningAlgorithms"][0]
+                scheme = cls._get_ecdsa_scheme(aws_scheme)
+            elif keytype == "rsa":
+                scheme = "rsassa-pss-sha256"
```
lukpueh, line 151:
```suggestion
                local_scheme = "rsassa-pss-sha256"
```

securesystemslib/signer/_aws_signer.py
```diff
@@ -101,136 +116,120 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
...
-        except KeyError as e:
-            raise exceptions.UnsupportedAlgorithmError(
-                f"{local_scheme} is not a supported signing algorithm"
-            ) from e
 
         keyval = {"public": public_key_pem}
-        keyid = compute_default_keyid(keytype, local_scheme, keyval)
-        public_key = SSlibKey(keyid, keytype, local_scheme, keyval)
+        keyid = compute_default_keyid(keytype, scheme, keyval)
+        public_key = SSlibKey(keyid, keytype, scheme, keyval)
```
lukpueh, line 162:
```suggestion
        keyid = compute_default_keyid(keytype, local_scheme, keyval)
        public_key = SSlibKey(keyid, keytype, local_scheme, keyval)
```

securesystemslib/signer/_aws_signer.py
```diff
@@ -101,136 +116,120 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
...
             raise UnsupportedLibraryError(AWS_IMPORT_ERROR)
 
-        client = boto3.client("kms")
-        request = client.get_public_key(KeyId=aws_key_id)
+        try:
+            if local_scheme:
+                if local_scheme not in cls.aws_signing_algorithms:
+                    raise ValueError(f"Unsupported scheme: {local_scheme}")
+        except KeyError as e:
+            raise ValueError(f"Unsupported scheme: {local_scheme}") from e
```
lukpueh, line 130: Oh and the try/except block should also go away.

ianhundere:
> Thanks for the updates! There is one small bug in `import_` which needs to be fixed before we can merge this. The rest is polishing.

okay, i think that about does it; thanks for all your feedback. 🙇

securesystemslib/signer/_aws_signer.py
```diff
@@ -101,136 +116,133 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
...
+            raise ValueError(
+                f"The AWS KMS key does not support the scheme: {local_scheme}"
+            )
+        if (
+            keytype == "rsa"
+            and local_scheme not in cls.aws_signing_algorithms.values()
+        ):
+            raise ValueError(
+                f"The AWS KMS key does not support the scheme: {local_scheme}"
+            )
```
lukpueh, line 156: Don't these checks always raise when `local_scheme` is not passed?

ianhundere: ah, you're correct. don't think these checks are necessary
```
        if (
            keytype == "ecdsa"
            and cls._get_ecdsa_scheme(aws_scheme) != local_scheme
        ):
            raise ValueError(
                f"The AWS KMS key does not support the scheme: {local_scheme}"
            )
        if (
            keytype == "rsa"
            and local_scheme not in cls.aws_signing_algorithms.values()
        ):
            raise ValueError(
                f"The AWS KMS key does not support the scheme: {local_scheme}"
            )
```

lukpueh: Let's add something like this in a follow-up PR.

ianhundere: ah, i just pushed new commit.

lukpueh: no need to revert. I think, now that you swapped the blocks, it works. let me give it one final look.

securesystemslib/signer/_aws_signer.py
```diff
@@ -101,136 +116,132 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
...
+        if keytype == "ecdsa" and local_scheme != cls._get_ecdsa_scheme(
+            aws_scheme
+        ):
+            raise ValueError(
+                f"The AWS KMS key does not support the scheme: {local_scheme}"
+            )
+
+        if (
+            keytype == "rsa"
+            and local_scheme not in cls.aws_signing_algorithms.values()
```
lukpueh, line 159: 🙈 this compares a string like "rsassa-pss-sha256"  to strings like "RSASSA_PSS_SHA_256"...

securesystemslib/signer/_aws_signer.py
```diff
@@ -101,136 +116,132 @@ def import_(cls, aws_key_id: str, local_scheme: str) -> Tuple[str, Key]:
...
+        ):
+            raise ValueError(
+                f"The AWS KMS key does not support the scheme: {local_scheme}"
+            )
```
lukpueh, line 163: Let's remove these checks after all. And add them after #612...

```suggestion
```

lukpueh (changes requested):

securesystemslib/signer/_aws_signer.py
ianhundere: oops, should have been:
```
        if keytype == "ecdsa" and local_scheme != cls._get_ecdsa_scheme(
            aws_scheme
        ):
            raise ValueError(
                f"The AWS KMS key does not support the scheme: {local_scheme}"
            )

        if (
            keytype == "rsa"
            and local_scheme not in cls.aws_signing_algorithms
        ):
            raise ValueError(
                f"The AWS KMS key does not support the scheme: {local_scheme}"
            )
```

securesystemslib/signer/_aws_signer.py
ianhundere: ah okay, i can i add what i mentioned above after https://github.com/secure-systems-lab/securesystemslib/issues/612.

lukpueh (approved):

PR #743, opened 2024-03-09
Change securesystemslib.dsse.Envelope.signatures to dict upstream

NicholasTanz:
Fixes: https://github.com/theupdateframework/python-tuf/issues/2564 

### DIFF
--- a/securesystemslib/dsse.py
+++ b/securesystemslib/dsse.py
@@ -17,7 +17,7 @@ class Envelope:
     Attributes:
         payload: Arbitrary byte sequence of serialized body.
         payload_type: string that identifies how to interpret payload.
-        signatures: list of Signature.
+        signatures: dict of Signature key id and Signatures.
 
     """
 
@@ -26,7 +26,7 @@ class Envelope:
     ):
         self.payload = payload
         self.payload_type = payload_type
-        self.signatures = signatures
+        self.signatures = {sig.keyid: sig for sig in signatures}
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, Envelope):
@@ -69,7 +69,7 @@ class Envelope:
         """Returns the JSON-serializable dictionary representation of self."""
 
         signatures = []
-        for signature in self.signatures:
+        for signature in list(self.signatures.values()):
             sig_dict = signature.to_dict()
             sig_dict["sig"] = b64enc(bytes.fromhex(sig_dict["sig"]))
             signatures.append(sig_dict)
@@ -101,7 +101,7 @@ class Envelope:
         """
 
         signature = signer.sign(self.pae())
-        self.signatures.append(signature)
+        self.signatures[signature.keyid] = signature
 
         return signature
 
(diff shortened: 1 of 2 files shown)


securesystemslib/dsse.py
```diff
@@ -26,7 +26,7 @@ def __init__(
     ):
```
NicholasTanz, line 26: I wasn't sure if I should change the last param from signatures: List[Signature] to signatures: Dict[str, Signature], so any information is appreciated

NicholasTanz:
Hey @lukpueh , I was wondering if you could take a look at this pr when you get a chance? thanks

securesystemslib/dsse.py
lukpueh: Good question! Please do change the constructor argument to `Dict[str, Signature]` and implement the translation in `from_dict` and `to_dict`. This is also what we do with [python-tuf  `Metadata`](https://github.com/theupdateframework/python-tuf/blob/e1b52e7fd828c11159a12aef4259133c3f05accf/tuf/api/metadata.py#L125).

lukpueh: I also just saw that we raise, if there are [duplicate (by keyid)](https://github.com/theupdateframework/python-tuf/blob/e1b52e7fd828c11159a12aef4259133c3f05accf/tuf/api/metadata.py#L189-L195) signatures there. I suggest to do the same in `Envelope.from_dict`.

lukpueh (changes requested):
Thanks for the clean patch! I replied to your question inline. Please address and we can merge this. FYI I already tried out what changes this will need in tuf and in-toto, and it looks reasonable (see referenced commits above).

securesystemslib/dsse.py
```diff
@@ -58,18 +61,23 @@ def from_dict(cls, data: dict) -> "Envelope":
...
-        signatures = []
+        signatures = {}
         for signature in data["signatures"]:
             signature["sig"] = b64dec(signature["sig"]).hex()
-            signatures.append(Signature.from_dict(signature))
+            signature = Signature.from_dict(signature)
+            if signature.keyid in signatures:
+                raise ValueError(
+                    f"Multiple signatures found for keyid {signature.keyid}"
+                )
```
lukpueh, line 71: Thanks for the check. Would you mind adding a small test?

securesystemslib/dsse.py
```diff
@@ -58,18 +61,23 @@ def from_dict(cls, data: dict) -> "Envelope":
...
+            signatures[signature.keyid] = signature
 
         return cls(payload, payload_type, signatures)
 
     def to_dict(self) -> dict:
         """Returns the JSON-serializable dictionary representation of self."""
 
         signatures = []
-        for signature in self.signatures:
+        for signature in list(self.signatures.values()):
```
lukpueh, line 80: Is there a reason to convert to `list`? You could iterate directly over  `dict_values`. Same comment applies below.

NicholasTanz: ah no - I just followed the format here [comment](https://github.com/theupdateframework/python-tuf/pull/2436#discussion_r1482773245) . Will change to just iterating over dict_values

lukpueh: Yeah, that PR has a few such conversions for type checking reasons.

lukpueh (approved):
LGTM!

PR #749, opened 2024-03-13
Add signer API code docs introductory text

lukpueh:
The idea was to provide just enough detail to "get" the signer api at a glance.  Finding the right words and the right level of detail was quite hard. But it's a start.

My feeling is that the intro could use a little bit more detail, and the docstrings a little less.

Plus there should be examples, and public methods in specific Signer and Key implementations need to be included

### DIFF
--- a/docs/signer.rst
+++ b/docs/signer.rst
@@ -3,12 +3,64 @@ Signer API
 
 .. currentmodule:: securesystemslib.signer
 
+At its core the signer API defines abstract interfaces to create and verify
+cryptographic signatures:
+
+* ``Signer.sign``
+* ``Key.verify_signature``
+
+These interfaces can be implemented using arbitrary PKI technology. Many
+default implementations are included.
+
+In addition, the API provides generic *load* methods for signers and public
+keys:
+
+* ``Signer.from_priv_key_uri`` - to load a signer from a URI
+* ``Key.from_dict`` - to load a public key from a serialized format
+
+These methods allow an application to use the exact same code, in order to load
+any signer or public key, which implements above interface, independently of
+the specific implementation.
+
+For a signer or public key implementation to become discoverable, it needs
+to be *registered* in the corresponding lookup table:
+
+* ``SIGNER_FOR_URI_SCHEME``
+* ``KEY_FOR_TYPE_AND_SCHEME``
+
+Usage
+-----
+The signer API is streamlined for the following series of user events,
+which may happen on different systems and at different points in time:
+
+1. **Generate** key pair (signature provider -specific)
+
+   Typically, this is done outside of the signer API, e.g. by using a Cloud KMS
+   web UI or an HSM console.
+
+2. **Configure** public key and signer access URI
+
+   The public key for a signer must be available in the signing context,
+   in order to express its eligibility.
+
+   Some of the existing signer implementations have methods to import a public
+   key from a signature provider and to build the related signer access URI.
+
+   The public key can then be serialized with interface methods for use
+   in the signing context and in the verification context.
+
+3. **Sign**, given a configured public key and signer access URI
+
+4. **Verify**, given a configured public key
+
 .. warning::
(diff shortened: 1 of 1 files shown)


docs/signer.rst
```diff
@@ -3,12 +3,64 @@ Signer API
 
 .. currentmodule:: securesystemslib.signer
 
+At its core the signer API defines abstract interfaces to create and verify
+cryptographic signatures:
+
+* ``Signer.sign``
+* ``Key.verify_signature``
+
+These interfaces can be implemented using arbitrary PKI technology. Many
```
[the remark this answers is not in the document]
lukpueh: I was trying to get away with not having the right term here.  Originally, I wanted to say something like public key / asymmetric cryptography, but the sigstore signer made me feel unsure about it.  So I opted for the more vague term PKI.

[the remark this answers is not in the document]
lukpueh: Maybe it's okay to just say "using arbitrary asymmetric signing technology". What do you think?

docs/signer.rst
```diff
@@ -3,12 +3,64 @@ Signer API
...
+default implementations are included.
+
+In addition, the API provides generic *load* methods for signers and public
+keys:
+
+* ``Signer.from_priv_key_uri`` - to load a signer from a URI
```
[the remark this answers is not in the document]
lukpueh: I'll try to expand on URIs here.

docs/signer.rst
```diff
@@ -3,12 +3,71 @@ Signer API
...
+In addition to sign and verify interface methods, the signer API provides
+generic *load* methods:
+
+* ``Signer.from_priv_key_uri`` -  Loads any specific signer from a URI. The
+  specific signer implementation itself is responsible for the URI format and
+  resolution. To become discoverable, signers and their URI schemes are
+  registered in the ``SIGNER_FOR_URI_SCHEME`` lookup table.
+
+* ``Key.from_dict`` - Loads any specific key from a serialized format. The
+  specific key implementation is responsible public key format and
```
joshuagl, line 27: something's missing in the description here, perhaps
> specific key implementation is responsible __for the__ public key format and deserialization

docs/signer.rst
```diff
@@ -3,12 +3,71 @@ Signer API
 
 .. currentmodule:: securesystemslib.signer
 
+At its core the signer API defines abstract interfaces to create and verify
+cryptographic signatures:
```
jku, line 7: I would love it if the first sentence was about the value/service provided to reader, not the mechanism.

```suggestion
The signer module provides a simple and unified cryptographic signing and signature verification API designed to work consistently regardless of the underlying signing technology. This is achieved with abstract interfaces backed by multiple implementations:  
```

docs/signer.rst
```diff
@@ -3,12 +3,71 @@ Signer API
...
+An application can use these same methods to uniformly load any signer or public
+key, regardless of the specific implementation. Many signers and keys are
+already included in the signer API. And an application can also create and
+register its own.
+
+
+Usage
+-----
+The signer API is streamlined for the following series of user events,
+which may happen on different systems and at different points in time:
```
jku, line 41: I wonder if it's useful to make the connection to TUF explicitly: that the importance of separating these steps comes from the core use case (could maybe even explain the back-and-and-forth of a yubikey holder and a TUF repository when becoming a signer).

Maybe this should be a separate section that explains the design decisions?

jku (approved):
Looks like a great start
* left a suggestion for how to start
* I think once we have some example code we can make this more concrete
  * either link to example app from the usage section
  * or include snippets of the example in this document
  * or some combination

docs/signer.rst
lukpueh: I was going back and forth about mentioning TUF or not, and ended up without because I didn't want to explain what TUF is, nor assume knowledge about it.

Let's reconsider in a second iteration.

PR #756, opened 2024-03-22
add public key attribute to Signer interface

### LINKED ISSUE #605: signer API: add public key attribute to Signer interface
Currently, the Signer interface does not prescribe the internal data structure of a signer implementation. This is because a signer is responsible for one thing only -- signing, and can be completely decoupled from private key data, and mostly decoupled from public key data.

The latter actually is part of the interface method [`from_priv_key_uri`](https://github.com/secure-systems-lab/securesystemslib/blob/5a2d43951d2be7fdb084a40cb723420bf345f54a/securesystemslib/signer/_signer.py#L168-L174), which is the canonical way of loading signers, and is usually based on information from the public key (e.g. keytype/scheme).

While signers don't necessarily require the full public key object once loaded, all current implementations [^1] do cache and use it for signing, usually to dispatch on scheme, and assign keyid to the resulting signature.

If we add the de-facto interface field `public_key: Key` indeed to the interface, signers could also be used to verify their signatures, for which use cases exist [^2].

The alternative to this proposed change would be to introduce a container format for "signer, public key"-tuples. Although, this seems unnecessary, if most signer already contain the public key.

[^1]: With the exception of `SSlibSigner`, which could easily provide an alias for `public_key` using the internal data structure, and might even become obsolete (see #604)

[^2]: `runlib.in_toto_record_stop` needs to verify preliminary link metadata, and then re-sign it.

terrycain:
Added a abstract property of `public_key` to the Signer abstract base class. As `self.public_key` was being used by the signers inheriting from Signer, they were updated to `self._public_key` and the property added. 

The Signer property itself doesn't contain an implementation as that class doesn't define a constructor so you cannot ensure `self._public_key` would exist. 

Fixes #605

### DIFF
--- a/securesystemslib/signer/_sslib_signer.py
+++ b/securesystemslib/signer/_sslib_signer.py
@@ -48,6 +48,7 @@ class SSlibSigner(Signer):
     def __init__(self, key_dict: Dict):
         self.key_dict = key_dict
         self._crypto_signer = CryptoSigner.from_securesystemslib_key(key_dict)
+        self._public_key = SSlibKey.from_securesystemslib_key(key_dict)
 
     @classmethod
     def from_priv_key_uri(
@@ -98,6 +99,10 @@ class SSlibSigner(Signer):
         keydict["keyval"]["private"] = private
 
         return cls(keydict)
+
+    @property
+    def public_key(self) -> Key:
+        return self._public_key
 
     def sign(self, payload: bytes) -> Signature:
         """Signs a given payload by the key assigned to the SSlibSigner instance.
--- a/securesystemslib/signer/_aws_signer.py
+++ b/securesystemslib/signer/_aws_signer.py
@@ -64,9 +64,13 @@ class AWSSigner(Signer):
 
         self.hash_algorithm = self._get_hash_algorithm(public_key)
         self.aws_key_id = aws_key_id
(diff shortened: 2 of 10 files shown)


securesystemslib/signer/_sslib_signer.py
```diff
@@ -99,6 +100,10 @@ def from_priv_key_uri(
 
         return cls(keydict)
 
+    @property
+    def public_key(self) -> Key:
+        return self._public_key
```
jku, line 105: Since sslibsigner is just a wrapper it could just return the underlying `self._crypto_signer.public_key` without storing a copy itself... but I think this is fine too: the whole class is deprecated anyway, we'll probably remove it very soon.

jku (approved):
Looks great, thanks

lukpueh (approved):

terrycain:
Awesome, thanks :)

jku:
@terrycain if you were in the room on friday, sorry for not noticing: too many people for my monkey brain. Also congrats for being the only one who got their work actually merged on the day, the rest of us are still wrangling branches...

terrycain:
@jku Yeah though only for long enough to submit the PR as then I had to disappear to deal with some work stuff 😄

PR #759, opened 2024-03-26
CryptoSigner: Use envvar as private key path prefix

jku:
* Introduce new "file2:" scheme for private key uris. This is different from both "file:" and "fn:" (used in RSTUF) to make sure it's not used accidentally
* encryption is not supported unlike in "file:"
* environment variable CRYPTO_SIGNER_PATH_PREFIX is introduced as a prefix for all private key lookups: this way signing system can choose the key directory at runtime but private key uris can still refer to a specific file within the directory
* Prefix could also be supported via a class variable but this is not currently implemented

### DIFF
--- a/securesystemslib/signer/_crypto_signer.py
+++ b/securesystemslib/signer/_crypto_signer.py
@@ -114,14 +114,6 @@ class CryptoSigner(Signer):
           create a signer: use this when you already have a private
           key  (and a private key URI) you can use.
 
-          URI has the format "file2:<PATH>", where PATH is a filesystem
-          path to the private key file. If CRYPTO_SIGNER_PATH_PREFIX
-          environment variable set, the private key will be read from
-          ``CRYPTO_SIGNER_PATH_PREFIX + <SEPARATOR> + PATH``. The
-          purpose of this is to allow PATH to only encode an identifier
-          (e.g. filename) while allowing the signing system to store
-          the private keys where ever it wants at runtime.
-
         b. newly generated key pair -- see ``CryptoSigner.generate_*()``
 
           Use this when you need a brand new private key pair.
@@ -217,7 +209,7 @@ class CryptoSigner(Signer):
         is set, the private key will be read from
         ``CRYPTO_SIGNER_PATH_PREFIX + <SEPARATOR> + PATH``. The purpose of this
         is to allow PATH to only encode an identifier (e.g. filename) while allowing
-        the signing system to store the private keys where ever it wants at runtime.
+        the signing system to store the private keys whereever it wants at runtime.
 
         Additionally raises:
             UnsupportedLibraryError: pyca/cryptography not installed
--- a/securesystemslib/signer/_crypto_signer.py
+++ b/securesystemslib/signer/_crypto_signer.py
@@ -1,6 +1,7 @@
 """Signer implementation for pyca/cryptography signing. """
 
 import logging
+import os
 from dataclasses import astuple, dataclass
 from typing import Optional, Union
 from urllib import parse
@@ -26,7 +27,6 @@ try:
     from cryptography.hazmat.primitives.asymmetric.padding import (
         MGF1,
         PSS,
-        AsymmetricPadding,
         PKCS1v15,
     )
     from cryptography.hazmat.primitives.asymmetric.rsa import (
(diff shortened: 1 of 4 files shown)


jku:
@lukpueh I'm trying to make sense of the docs and I'm wondering if I should make private key (serialization) public as well?

This would allow key generation like this (without using cryptography directly).  
```python
signer = CryptoSigner.generate_ed25519()
with open("new_private_key.pem", "wb") as f:
    f.write(signer.private_key)
# store "file:new_private_key.pem" as the uri that allows accessing the private key later
```

I think this would make CRYPOSIGNER.md a little more reasonable?

lukpueh:
>I'm trying to make sense of the docs and I'm wondering if I should make private key (serialization) public as well?


Yes, could be handy, but maybe not necessary. I'd rather simplify on the pubkey side of things, because the way it is currently shown is not really how the signer API is intended. What if we ignore pubkey transmission altogether ...


1. Prepare
```python
from cryptography.hazmat.primitives import asymmetric, serialization

# Generate key pair 
private_key = asymmetric.ed25519.Ed25519PrivateKey.generate()

# Deploy private key
private_pem = private_key.private_bytes(
    encoding=serialization.Encoding.PEM,
    format=serialization.PrivateFormat.PKCS8,
)
with open("private.pem", "wb") as f:
    f.write(private_pem)

# Authorize/configure signer
# TODO: mention how public key and uri may end up in the signing environment (e.g. with "x-signer-uri", and json.dumps(public_key.to_dict())
public_key = SSlibKey.from_crypto(private_key.public_key())
uri = "file2:private.pem"


```

2. Sign
```python
from securesystemslib.signer import Signer

# TODO: mention where public key and uri may come from
signer = Signer.from_priv_key_uri(uri, public_key)
signer.sign(b"data")
```

jku:
Yeah I suppose that's the minimal change needed. Then we can file another issue to enable this for the prepare step


```
# Generate key pair 
signer = CryptoSigner. generate_ed25519()

# Store private key 
with open("private.pem", "wb") as f:
    f.write(signer.private_key)

# Authorize/configure signer
# TODO: mention how public key and uri may end up in the signing environment 
public_key = signer.public_key
uri = "file2:private.pem"

```

jku:
Maybe this is now good for review.

We really should add short generic doc about supported keys/signers but I guess it's not part of this PR?

securesystemslib/signer/_crypto_signer.py
```diff
@@ -108,24 +108,32 @@ class CryptoSigner(Signer):
...
 
         c. existing pyca/cryptography private key object -- ``CryptoSigner()``
 
+          Use this if you need a brand new private key pair and option
+          b is not flexible enough for your case.
     """
 
-    FILE_URI_SCHEME = "file"
+    SCHEME = "file2"
+    PREFIX_ENV_VAR = "CRYPTO_SIGNER_PATH_PREFIX"
```
jku, line 136: the environment variable name could have _securesystemslib_ in it but I suppose it's long enough already

securesystemslib/signer/_crypto_signer.py
```diff
@@ -108,24 +108,32 @@ class CryptoSigner(Signer):
...
+          key  (and a private key URI) you can use.
 
-          A related public key must be passed.
+          URI has the format "file2:<PATH>", where PATH is a filesystem
+          path to the private key file. If CRYPTO_SIGNER_PATH_PREFIX
+          environment variable set, the private key will be read from
+          ``CRYPTO_SIGNER_PATH_PREFIX + <SEPARATOR> + PATH``. The
+          purpose of this is to allow PATH to only encode an identifier
+          (e.g. filename) while allowing the signing system to store
+          the private keys where ever it wants at runtime.
```
lukpueh, line 123: spelling nit:

```suggestion
          the private keys wherever it wants at runtime.
```

securesystemslib/signer/_crypto_signer.py
lukpueh: I think it's fine. An application can subclass CryptoSigner and override `PREFIX_ENV_VAR`.

tests/test_signer.py
```diff
@@ -657,27 +658,21 @@ def test_from_priv_key_uri(self):
...
         for keytype, scheme, public_key_value, fname in test_data:
-            for encrypted in [True, False]:
-                if encrypted:
-                    priv_fname = fname.replace(".pem", "_encrypted.pem")
-                    uri = f"file:{PEMS_DIR / priv_fname}?encrypted=true"
-
-                    def handler(_):
-                        return "hunter2"
-
+            for use_prefix in [True, False]:
```
lukpueh, line 662: Nice re-purposing of the inner loop! :)

securesystemslib/signer/_crypto_signer.py
```diff
@@ -204,6 +212,13 @@ def from_priv_key_uri(
         expected (and tested) encoding/format is PEM/PKCS8. Other formats may
         but are not guaranteed to work.
 
+        URI has the format "file2:<PATH>", where PATH is a filesystem path to the
+        private key file. If CRYPTO_SIGNER_PATH_PREFIX environment variable
+        is set, the private key will be read from
+        ``CRYPTO_SIGNER_PATH_PREFIX + <SEPARATOR> + PATH``. The purpose of this
+        is to allow PATH to only encode an identifier (e.g. filename) while allowing
+        the signing system to store the private keys where ever it wants at runtime.
```
lukpueh, line 220: spelling nit2:
```suggestion
        the signing system to store the private keys wherever it wants at runtime.
```

Regardless, do you think it's necessary to have the same block here and in the class docstring?

lukpueh (approved):
Very nice! I only left some minor docstrings comments. Otherwise this looks great.

Issue #669, opened 2023-11-02
"ecdsa-sha2-nistp256" wrongly used as default scheme for any "ecdsa" key


lukpueh:
**Description of issue or feature request**:
*ecdsa* has repeatedly been a source of confusion for securesystemslib and its users:
https://github.com/secure-systems-lab/securesystemslib/issues/239, https://github.com/secure-systems-lab/securesystemslib/issues/363, https://github.com/secure-systems-lab/securesystemslib/issues/513, https://github.com/theupdateframework/specification/pull/272,  https://github.com/theupdateframework/python-tuf/issues/1388, https://github.com/theupdateframework/go-tuf/pull/508

 This is because supported ecdsa signing *schemes* contain additional details, which either **can** vary for a given key (*hash*, e.g. `sha2`) or **cannot** (*curve*, e.g. `nistp256`).

For comparison, rsa signing *schemes* only contain details, which can vary (*padding*, e.g. `pss`, and *hash*, eg. `sha256`). 

**Current behavior**:
There is at least one case, where securesystemslib code uses "ecdsa-sha2-nistp256" as the default scheme for any "ecdsa" key, even if that key uses a different curve:

```python
>>> from securesystemslib.signer import SSlibKey
>>> from cryptography.hazmat.primitives.asymmetric import ec
>>> 
>>> nistp384 = ec.SECP384R1()
>>> private = ec.generate_private_key(ec.SECP384R1())
>>> public = SSlibKey._from_crypto_public_key(private.public_key(), None, None)
>>> public.scheme
'ecdsa-sha2-nistp256'

```

**Expected behavior**:

* Fix above
* Bonus: Check if there are other occurrences, where this happens

PR #763, opened 2024-04-02
SSlibKey: fix ecdsa nistp384 default scheme

lukpueh:
fixes https://github.com/secure-systems-lab/securesystemslib/issues/669

Fix `SSlibKey.from_crypto` to assign the correct default scheme for ecdsa nistp384 keys.

This includes refactoring from_crypto to use a single switch over pyca/cryptography public key object type, in order to infer securesystemslib "keytype" and default scheme.

Previously, only "keytype" was inferred from the key object, and the default scheme then from the keytype. Given that for the same keytype (ecdsa) there can be different default schemes, this needed to be changed.

The refactoring also moves the keytype-specific public key value serialization to the same switch.


jku (approved):
seems right to me

Issue #731, opened 2024-02-07
Remove legacy interfaces/implementation


lukpueh:
The lack of a clear API in securesystemslib has been described in #270 (also see below for related issues).

The new Signer API (`secureystemslib.signer`) provides a by far more powerful replacement API for signing and verifying digital signatures.

This issue calls for the removal of legacy interfaces/implementation, in order to finalise securesystemslib's transition to the Signer API.

IMPORTANT: The removal and a subsequent release must be coordinated with securesystemslib users, most notably in-toto, python-tuf, and RSTUF. cc @adityasaky, @SantiagoTorres, @trishankatdatadog, @jku, @joshuagl, @kairoaraujo, @MVrachev.

**Modules to remove**
* keys.py
* {rsa, ecdsa, ed25519}_keys.py
* interface.py
* schema.py
* _sslib_signer.py
* settings.py

**Other**
* formats.py (remove all but `encode_canonical`)
* remove deprecated methods in SSlibKey

**Maybe revise**
* util.py (e.g. [#245](https://github.com/secure-systems-lab/securesystemslib/issues/245))
* hash.py
* storage.py (consider move to python-tuf)

closes #183,  #172 #219, #270, #280, #286, #309, #310, #367

jku:
cc @woodruffw: This should have no effect on sigstore-python (as `encode_canonical()` remains available) but FYI.

woodruffw:
Thanks for the heads up!

lukpueh:
> **Maybe revise**
> 
> * hash.py
> * storage.py (consider move to python-tuf)

That's all what's left here. I suggest to move the discussion to #270 and close here.

PR #767, opened 2024-04-03
Release v???

lukpueh:
- bump version in __init__.py
- add changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## securesystemslib v0.32.0
+
+### Added
+Signer: Add public key attribute to interface (#756)
+CryptoSigner: Support ecdsa keytype that is no longer in spec (#711)
+
+### Changed
+* dependabot: Use groups, update weekly #735
+* dsse: Change Envelope.signatures type to dict (#743)
+* docs: Improve user and contributor docs (#744, #745, #746, #749)
+
+### Fixed
+* Fix check-upstream-ed25519 workflow permission #706
+
 ## securesystemslib v0.31.0
 
 ### Added
--- a/securesystemslib/__init__.py
+++ b/securesystemslib/__init__.py
@@ -1,7 +1,7 @@
 # pylint: disable=missing-module-docstring
 import logging
 
-__version__ = "0.31.0"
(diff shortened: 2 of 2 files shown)


jku:
Just so we remember:

This release should not happen 
* until users of python-tuf have had time to update to `tuf 4.0`
* and maybe even until the main users of python-tuf have made a release

This is because especially tuf.ngclient users may not be direct securesystemslib users so may not pin the securesystemslib version.

jku:
Just documenting some users:
* python-tuf 4.0 caps securesystemslib < 0.32
* tuf-on-ci 0.9 caps securesystemslib < 0.32
* root-signing-staging has upgraded to tuf-on-ci 0.9
* sigstore-python main uses python-tuf 4.0 (but no release yet)
* rstuf apparently already pins securesystemslib

TODO:
* There is a sigstore-python 2.1.5 point release coming up hopefully today that caps securesystemslib < 0.32

lukpueh:
> Just documenting some users:

- in-toto 2.3.0 caps securesystemslib < 0.32

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Changelog
 
+## securesystemslib v0.32.0
```
lukpueh, line 3: Should we make this 1.0.0 already?

I submitted a host of PRs yesterday to remove all legacy code (#731). If we merge those, I think we are ready.

lukpueh: Before we tag 1.0.0 we should remove disclaimers like this:

https://github.com/secure-systems-lab/securesystemslib/blame/main/docs/signer.rst#L62-L67 

(there may be others)

jku:
This is something I would have like before 1.0: **securesystemslib/py.typed and complete annotations for public api**

Notes on that 
* leave _gpg and _vendor unannotated, 
* this is still a lot of work because of how `hash` and `formats` are written -- I guess we could just annotate them with "Any" (with the idea that it's better to get `signer` annotations used even if hash and formats are not).
* there's still some missing annotations  because we don't currently enable 

        disallow_untyped_defs = True
        disallow_incomplete_defs = True

I think I would rather get a release out and then start improving that situation

jku:
superseded by #807

PR #771, opened 2024-04-10
Remove deprecated SSlibSigner and from/to_securesystemslib_key methods

lukpueh:
- addresses part of [#731](https://github.com/secure-systems-lab/securesystemslib/issues/731)
- please review commit by commit and see commit messages for details

---

This PR removes the deprecated SSlibSigner.
CryptoSigner can be used as replacement. 
The most interesting (review-worthy) bits of this PR are the two initial test change commits.

In addition, this PR removes three deprecated / obsolete translation methods between Signer API objects and the legacy secureysystemslib key dictionary format. These are not directly related to SSlibSigner but still need to be removed as part of #731.


jku (approved):
:1st_place_medal:

PR #775, opened 2024-04-10
Remove schema checks in securesystemslib.gpg

lukpueh:
- addresses part of [#731](https://github.com/secure-systems-lab/securesystemslib/issues/731)
- ~blocks on [#772](https://github.com/secure-systems-lab/securesystemslib/pull/772) (commits are included in this PR)~

---

In preparation for the removal of schema.py (https://github.com/secure-systems-lab/securesystemslib/issues/183), this patch removes schema checks in the following modules of the `securesystemslib.gpg` subpackage:

* internal modules `rsa`, `dsa`, `eddsa`, `common`. These checks are redundant with schema checks that are already performed in the calling functions of the `functions` module.

* in previously public `functions` module: 
  * keyid in `create_signature` and `export_pubkey` functions 
  * public key and signature dict in `verify_signature` function

This is okay for two reasons:

1. the preferred way of interacting with `securesystemslib.gpg.functions` is via `GPGSigner`, which controls the format of the passed arguments to some extent

2. securesystemslib.gpg still raises meaningful and even more consistent errors for invalid arguments anyway, than it did before. E.g. a keyid passed to `export_pubkey` that doesn't conform to the previously checked hex schema, now raises a `KeyNotFoundError`.

Other changes include:

* move string literal `GPG_HASH_ALGORITHM_STRING` from `securesystemslib.schema` to the better suited `secureystemslib.gpg.constants` module.
* remove mentions of schema definitions in docstrings
* adopt changes in tests

PR #773, opened 2024-04-10
Remove mostly unused util.py

lukpueh:
- addresses part of [#731](https://github.com/secure-systems-lab/securesystemslib/issues/731)
- ~~blocks on [#772](https://github.com/secure-systems-lab/securesystemslib/pull/772) (commits are included in this PR)~~

closes #245 

---

The util.py module provides a mix of minimal functions related to local files and directories. These functions are mostly unused, exceptions are (according to sourcegraph.com):
- persist_temp_file in `python-tuf` (cc @jku)
- get_file_hashes in `scai-demos` (cc @marcelamelara, @adityasaky)
- get_file_details, load_json_file in `taf` (cc @renatav, @adityasaky)

It seems feasible to reimplement the desired functionality downstream.

The benefits of removing util.py are:
- minimizes the API surface
- removes one dependent of the about to be removed schema.py ([#183](https://github.com/secure-systems-lab/securesystemslib/issues/183))
- makes unittest_toolbox obsolete

PR #776, opened 2024-04-10
Remove schema facility and most of formats

### LINKED ISSUE #183: Revise schema and formats facility

lukpueh:
- addresses part of [#731](https://github.com/secure-systems-lab/securesystemslib/issues/731)
- ~blocks on [#773](https://github.com/secure-systems-lab/securesystemslib/pull/773),  [#774](https://github.com/secure-systems-lab/securesystemslib/pull/774),  [#775](https://github.com/secure-systems-lab/securesystemslib/pull/775) (commits are included in this PR)~

close #183 

---

The usage of the securesystemslib schema/formats facility has been little by little phased out and is now nowhere used anymore. Several reasons to not use it are listed in https://github.com/secure-systems-lab/securesystemslib/issues/183.

This patch removes schema.py along with all securesystemslib schema definitions in formats.py.

Additionally, it removes unrelated and unused helper functions in formats.py.


lukpueh:
Some windows runners seem to calculate test coverage differently than the rest and thus drop below the already lowered constraint. I'll fix this, when this PR is no longer blocked by other PRs.

lukpueh:
> Some windows runners seem to calculate test coverage differently than the rest and thus drop below the already lowered constraint.

9be1638 further lowers the coverage requirement. As before, this does not mean that there are fewer tests (see commit message).

jku:
> this does not mean that there are fewer tests (see commit message).

Some of the bad coverage is quite expected (aws, azure, gcp, sigstore). I was surprised that HSMSigner is the worst though -- I wonder what that's about.

jku (approved):

lukpueh:
> > this does not mean that there are fewer tests (see commit message).
> 
> Some of the bad coverage is quite expected (aws, azure, gcp, sigstore). I was surprised that HSMSigner is the worst though -- I wonder what that's about.

https://github.com/secure-systems-lab/securesystemslib/blob/37089c53cde6842a5224b0b2cac90644a8d753df/.github/workflows/_test.yml#L62-L64

PR #777, opened 2024-04-11
Test AWSSigner with localstack

### LINKED ISSUE #612: signer: test AWS signer with LocalStack

lukpueh:
Add test setup to test AWSSigner locally or on GitHub CI using localstack.

**Change details**

* Add independent tox environment to init/cleanup localstack, configure ambient AWS KMS credentials, create test keys, and run the test.

* Adds aws requirements file with pinned runtime requirements to trigger tox run via Dependabot. Note: test requirements are not included, but defined directly in tox.ini without pinning (we don't really need to trigger of those update).

* Adds shell script to generate test keys for all supported schemes in tox setup.

* Update AWSSigner tests to match localstack setup. The previous setup looked like it needed manual intervention in order to run. The new tests runs the full Signer API flow  -- import (with and without passed scheme), load, sign, verify (pass and fail) --  automatically and for each scheme supported by AWSSigner.

* Adds independent GitHub Action workflow to run tox aws test

closes #612


lukpueh:
> * Adds independent GitHub Action workflow to run tox aws test

And it works too :)

https://github.com/secure-systems-lab/securesystemslib/actions/runs/8648633608/job/23712912574?pr=777

tox.ini
```diff
@@ -78,3 +78,38 @@ commands =
     pylint -j 0 --rcfile=pylintrc securesystemslib tests
     bandit --recursive securesystemslib --exclude _vendor
     mypy
+
+# Requires docker running
```
jku, line 82: is this why the test is not part of the default set?

jku (approved):
Nice, this looks like a solid improvement

tox.ini
lukpueh: Yes. I thought, we shouldn't require devs to run docker just for this. And it takes quite a while to start the container, especially the first time. 

On GitHub, there doesn't seem to be any default caching. Maybe, we could improve this, by pulling the images explicitly with docker or maybe by using the localstack GitHub Action. Right now docker stuff happens behind the scenes by running the `localstack` command from within tox, which has the advantage that we can run the same `tox` command locally and on GitHub.

Issue #765, opened 2024-04-03
AWSSigner: may return *incorrect* ecdsa-sha2-nistp512 scheme  


lukpueh:
`AWSSigner.import_`  supports importing a public key on curve nistp**521**, and returns it as SSlibKey with scheme "ecdsa-sha2-nistp**512**" (note the transposed digits!):

https://github.com/secure-systems-lab/securesystemslib/blob/185aa9cedeb6597c40be62d938297c75d1f930a0/securesystemslib/signer/_aws_signer.py#L174

A nistp512 cure does not exist, so this is likely a typo and should be "ecdsa-sha2-nistp521", which, however, is not supported by SSlibKey (see #764).

**Possible fixes**
* remove support for "ecdsa-sha2-nistp512" in AWSSigner
* change scheme name to "ecdsa-sha2-nistp521"
  - requires adding support for "ecdsa-sha2-nistp521" to SSlibKey (see [#764](https://github.com/secure-systems-lab/securesystemslib/issues/764))
  - may require other changes in AWSSigner, e.g `_get_hash_algorithm` won't work anymore ([nistp521 keys use sha512](https://docs.aws.amazon.com/kms/latest/developerguide/asymmetric-key-specs.html#key-spec-ecc)  for the message digest)

PR #778, opened 2024-04-11
Simplify and strengthen AWSSigner.import_

lukpueh:
**Simplify**
- remove redundant keytype/scheme switches
- remove try/except, which catches an exception only to re-raise it

**Strengthen**
- choose default scheme based on schemes supported by the key

Roughly implements the algorithm suggested in: https://github.com/secure-systems-lab/securesystemslib/pull/724#pullrequestreview-1864774479

**Unrelated changes**
- fix [#765](https://github.com/secure-systems-lab/securesystemslib/issues/765) fix typo AND disable scheme until SSlibKey adds support
- fail in `__init__` if public_key.scheme is not in aws_algos

**TODO:**
* add tests

Many thanks to @ianhundere for the solid ground work!

### DIFF
--- a/securesystemslib/signer/_aws_signer.py
+++ b/securesystemslib/signer/_aws_signer.py
@@ -1,11 +1,13 @@
 """Signer implementation for AWS Key Management Service"""
 
 import logging
-from typing import Optional, Tuple
+from typing import List, Optional, Tuple
 from urllib import parse
 
-from securesystemslib import exceptions
-from securesystemslib.exceptions import UnsupportedLibraryError
+from securesystemslib.exceptions import (
+    UnsupportedAlgorithmError,
+    UnsupportedLibraryError,
+)
 from securesystemslib.signer._key import Key, SSlibKey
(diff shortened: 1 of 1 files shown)


lukpueh:
>TODO:
> add tests

Test coverage is actually pretty good, especially compared to other cloud KMS signers. It tests the  full Signer API flow -- import (with and without passed scheme), load, sign, verify (pass and fail) for all supported schemes.

What's missing are tests for some failure cases, like AWS returning an unsupported key, etc. I think we can improve this in a future PRs.

jku (approved):
LGTM. 

Re-raising the error wasn't entirely pointless as it added the AWS keyid in the error message but ... I suppose it's reasonable to assume that caller would know what key they were trying to import: it's not like this happens in unexpected situations.

lukpueh:
> Re-raising the error wasn't entirely pointless

You are right, it was a bit sneaky of me to not mention this. Apologies!

PR #779, opened 2024-04-12
tests: Improve SigstoreSigner test

jku:
* The test now works locally and in PRs
* Test does require git and uses it to fetch a test token: because of this the test is still not included  in the default set
* Since this now runs in PRs, I've removed the "file issue on failure" functionality

### DIFF
--- a/tests/check_sigstore_signer.py
+++ b/tests/check_sigstore_signer.py
@@ -1,13 +1,24 @@
 """
 Test SigstoreSigner API.
 
-NOTE: The filename prefix is check_ instead of test_ so that tests are
-only run when explicitly invoked in a suited environment.
+These tests require git and will use it to fetch a testing identity token from
+https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon
 
+Because of these unusual requirements (and because sometimes the fetch may take
+a longer time) the test file is named check_* and is not included in the default
+tests.
 """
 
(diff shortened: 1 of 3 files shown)


tests/check_sigstore_signer.py
```diff
@@ -17,26 +28,76 @@
...
-class TestSigstoreSigner(unittest.TestCase):
-    """Test public key parsing, signature creation and verification.
 
-    Requires ambient credentials for signing (e.g. from GitHub Action).
+def identity_token() -> str:
+    """Return identity token for TEST_IDENTITY"""
+    # following code is modified from extremely-dangerous-public-oidc-beacon download-token.py.
+    # Caching can be made smarter (to return the cached token only if it is valid) if token
+    # starts going invalid during runs
+    min_validity = timedelta(seconds=20)
```
jku, line 43: since we run a single test, 20 secs is overkill: I'll shorten this

lukpueh (approved):
Very cool! And it seems to work. I can also confirm that the "download token" implementation matches the one upstream.

jku:
It looks like sigstore signer might be out of order for some time (seuresystemslib depends on sigstore which  depends on python-tuf that depends on previous version of securesystemslib)...

I'll merge this but we'll probably have to disable the test  pretty much right away, and wait for a sigstore release that is compatible

PR #792, opened 2024-04-23
Mark securesystemslib.gpg subpackage as internal

### LINKED ISSUE #270: Revise module architecture and define clear API modules and functions

lukpueh:
~TODO: `CommandError` and `KeyNotFoundError` raised in public `GPGSigner` methods are now internal. This doesn't seem right, even though I doubt that anyone needs to handle those errors. I think we can re-raise the latter as `ValueError`, because the key is not found for a passed keyid. `CommandError` could maybe be an `OSError`?~

done

---

The `gpg` subpackage provides a vaguely defined API (`gpg.functions`) to create signatures, export public keys, and verify signatures. This API and the used formats are incompatible with the securesystemslib signer API.

For the sake of a consistent API, the `gpg` subpackage is marked as internal (renamed to `_gpg`) and the above mentioned functionality is exposed via the new signer API. Replacement methods are:

- `GPGSigner.import_` (replaces `export_pubkey`)
- `GPGSigner.sign`
- `GPGKey.verify_signature`

Note that public key and signature formats also change, in order to match `Key` and `Signature` interfaces. This means:

- signature field `signature` is renamed to `sig`
- public key fields `type`, `method` and `hashes` are replaced by `keytype` and `scheme` fields, and
- public keys no longer include `subkeys` or key expiration infos. This means that the signature verification function no longer needs to decide, if a key is authorized or valid to verify a given signature.

See discussion for context: https://github.com/secure-systems-lab/securesystemslib/pull/488#issuecomment-1398169012, https://github.com/secure-systems-lab/securesystemslib/pull/488#issuecomment-1400190844

Closes #270


jku:
The exception ideas seem reasonable to me

jku:
Huh, mac tests seem to be running on arm now but python < 3.10 is not available...

jku:
> Huh, mac tests seem to be running on arm now but python < 3.10 is not available...

yeah macos-latest just changed from 12 to 14, and apparently 14 runs on arm now and does not have all python versions.

jku:
> I think it's okay to not test on <3.10 on Mac.

Yeah the whole matrix might be a little overkill anyway: would likely be fine to test all pythons on linux and just latest on mac and windows...

But the alternative is to pin macos-12 instead of macos-latest

jku (approved):
this feels right: all of this code is more or less an implementation detail of GPGSigner.

Issue #766, opened 2024-04-03
SSlibKey: consider stronger validation of keytype/scheme for keyval 


lukpueh:
The consistency of SSlibKey instances is mainly "validated" at usage time. That is, in `SSlibKey.verify_signature`, when keyval is deserialised based on scheme,  [unknown schemes](https://github.com/secure-systems-lab/securesystemslib/blob/185aa9cedeb6597c40be62d938297c75d1f930a0/securesystemslib/signer/_key.py#L397-L398) and undeserializable keyvals are filtered out. This validation is incomplete and also happens late.

For comparison, at creation time inputs are hardly validated (see e.g. in [from_dict](https://github.com/secure-systems-lab/securesystemslib/blob/185aa9cedeb6597c40be62d938297c75d1f930a0/securesystemslib/signer/_key.py#L233-L234) or in the [base constructor](https://github.com/secure-systems-lab/securesystemslib/blob/185aa9cedeb6597c40be62d938297c75d1f930a0/securesystemslib/signer/_key.py#L96-L99)). Note that an additional safeguard exists in the Key.from_dict deserialisation interface, which filters out unregistered keytype, scheme pairs.

Let's consider:
- adding more comprehensive validation, most notably check consistency of keytype, scheme, and keyval, and
- validating earlier, e.g. already in the constructor

See related issues related to invalid SSlibKey instances and validation: #764 #765, #669, #559

lukpueh:
An argument against more comprehensive and/or earlier validation is that a key might be deserialized but never used for signature verification. In that case it might not matter if the key is invalid.

Similarly, a key might be inconsistent, but still manage to verify a signature. This is likely the case for ecdsa keys, e.g. if the keyval contains a nistp256 key, but scheme is "ecdsa-sha2-nistp384" (or vice-versa).


I lean towards checking early and comprehensively, and failing hard if an SSlibKey is inconsistent. If we want to support deserialisation of unknown keys, we should implement this in the Key.from_dict interface.

PratGit1606:
I'd like to work on this. I noticed PRs #794 and #795 were opened but not merged,is there a preferred approach or direction you'd like a contributor to follow here?

jku:
This is a little tricky. lazy validation is a feature as well: in some cases the application deserializing a key might not actually need that key (in the TUF metadata case they might deserialize TUF metadata that contains keys that are not needed for whatever the application is doing). In this case failing at deserialize time could be annoying.

[...]

PR #795, opened 2024-04-24
Improve SSlibKey validation in _verify

lukpueh:
Added checks as requested in #766:

- keytype matches scheme
- scheme matches deserialized key type (only for pem formatted keyvals; for ed25519 the check happens implicitly on deserialization)
- scheme matches deserialized key curve (only for ecdsa)

Note that this PR does not move the checks to the constructor as suggested in #766. This may or may not be addressed in a follow-up PR.

failing tests fixed in #794

### DIFF
--- a/securesystemslib/signer/_key.py
+++ b/securesystemslib/signer/_key.py
@@ -341,9 +341,18 @@ class SSlibKey(Key):
 
     def _verify(self, signature: bytes, data: bytes) -> None:
         """Helper to verify signature using pyca/cryptography (default)."""
+
+        def _validate_type(key, type_):
+            if not isinstance(key, type_):
+                raise ValueError(f"bad key {key} for {self.scheme}")
+
+        def _validate_curve(key, curve):
+            if not isinstance(key.curve, curve):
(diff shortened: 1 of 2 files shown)


securesystemslib/signer/_key.py
```diff
@@ -341,9 +341,18 @@ def _verify_ed25519_fallback(self, signature: bytes, data: bytes) -> None:
 
     def _verify(self, signature: bytes, data: bytes) -> None:
```
lukpueh, line 342: The function has become a bit unwieldy. But I did resist the temptation to refactor anything, mostly because want this PR to be easy to review ... and also because I was unsure how to make it better. :)

One idea I had was to factor out the long deserialization if/else switch to something like a `_to_crypto` method: 504ffaea84ff6a6062de93cc9d277adac9cffe73

...

I'll think about this some more.

jku (approved):
Looks good to me. It's true that some things might make sense to do earlier but I think there is a good chance for mistakes in there so this feels right for now.

PR #797, opened 2024-04-24
ci: test macOS and Windows on latest Python only

lukpueh:
CI started failing recently, because new macos-latest runs on arm, which does not include all older Python versions.

As workaround and for the sake of a slimmer test matrix, we drop all but latest Python tests on macOS and Windows.

The remaining matrix should still give us reasonable coverage.

Related discussion in: https://github.com/secure-systems-lab/securesystemslib/pull/792#issuecomment-2072359562

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -8,13 +8,19 @@ jobs:
   test:
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


jku (approved):

jku:
You'll have to fix the expected checks in GitHub UI

PR #799, opened 2024-04-25
Add CryptoSigner.private_bytes

### LINKED ISSUE #798: CryptoSigner should expose private key bytes
Based on a couple of examples (the CRYPTOSIGNER.md doc and the example code in python-tuf) I think we should have a 
```
@property
def private_bytes(self) -> bytes
```
in CryptoSigner.

I'll make a PR

jku:
IMO this completes the MVP CryptoSigner API

Fixes #798

This is what the docs will look like: https://github.com/secure-systems-lab/securesystemslib/blob/cc68f4d1961da2bf274eab8b64baf4339ee06827/docs/CRYPTO_SIGNER.md

### DIFF
--- a/tests/test_signer.py
+++ b/tests/test_signer.py
@@ -703,6 +703,23 @@ class TestCryptoSigner(unittest.TestCase):
             with self.assertRaises(UnverifiedSignatureError):
                 signer.public_key.verify_signature(sig, b"NOT DATA")
 
+    def test_private_bytes(self):
+        """Test full cycle of generate -> private_bytes -> from_priv_key_uri"""
+        generate_funcs = [
+            CryptoSigner.generate_rsa,
+            CryptoSigner.generate_ecdsa,
+            CryptoSigner.generate_ed25519,
+        ]
+        for generate in generate_funcs:
+            signer = generate()
+            with open("privkey.pem", "wb") as f:
+                f.write(signer.private_bytes)
+
+            signer2 = Signer.from_priv_key_uri(
+                "file2:privkey.pem", signer.public_key
+            )
+            self.assertEqual(signer.private_bytes, signer2.private_bytes)
+
     def test_custom_crypto_signer(self):
         # setup
         key = self.keys[0]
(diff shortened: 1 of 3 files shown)


jku:
rebased on main

tests/test_signer.py
```diff
@@ -703,6 +703,23 @@ def test_generate(self):
             with self.assertRaises(UnverifiedSignatureError):
                 signer.public_key.verify_signature(sig, b"NOT DATA")
 
+    def test_private_bytes(self):
+        """Test full cycle of generate -> private_bytes -> from_priv_key_uri"""
+        generate_funcs = [
+            CryptoSigner.generate_rsa,
+            CryptoSigner.generate_ecdsa,
+            CryptoSigner.generate_ed25519,
+        ]
```
lukpueh, line 712: nit: Generate is already tested above. If we want to optimize test speed, we could use the existing test keys in PEMS_DIR / f"{keytype}_private.pem" here.

jku: yeah I guess some RSA keygen does take some time...

tests/test_signer.py
```diff
@@ -703,6 +703,21 @@ def test_generate(self):
...
 
+    def test_private_bytes(self):
+        """Test private_bytes -> from_priv_key_uri"""
+        for pem in ["rsa", "ecdsa", "ed25519"]:
+            with open(PEMS_DIR / f"{pem}_private.pem", "rb") as f:
+                privkey = load_pem_private_key(f.read(), None)
+                signer = CryptoSigner(privkey)
+
+            with open("privkey.pem", "wb") as f:
+                f.write(signer.private_bytes)
```
lukpueh, line 714: Should we write this to a temp/test dir? Otherwise this litters cwd.

jku: oh true. I wish this was a default unittest thing...

lukpueh (approved):
Nice! I think you should fix the test, to not leave behind a pem. Otherwise this looks great

PR #800, opened 2024-04-26
Add VaultSigner and tests

### LINKED ISSUE #307: Enable signing and verifying using keys stored in Hashicorp Vault

@trishankatdatadog did a preliminary implementation of Hashicorp Vault integration with the TUF reference implementation in theupdateframework/tuf#1060, but such functionality seems a better fit for securesystemslib. Particularly as having such functionality in securesystemslib would also make it more widely usable, such as by TUF's sibling project in-toto.

One model for implementing Vault support would be to mimic the patterns in the proposed securesystemslib PKCS11-based HSM interface in PR #229.

**Current behavior**:

Unable to sign or verify with keys stored in Vault.

**Expected behavior**:

Able to sign and verify with keys stored in Vault.

lukpueh:
Adds fully functional VaultSigner and basic "import_ -> from_priv_key_uri
-> sign -> verify" test loop using a local vault server.

Tests run on CI and can be run locally with tox, if vault is installed.

Note also that vault supports all key types and signing schemes supported by
SSlibKey. The here added VaultSigner, however, only supports `ed25519`
keys.

Other key types and schemes may be added in follow-up PRs

Fixes #307


jku:
> * fix ci failure (the test actually passes, but the vault tear down script seems to exit with non-zero and thus fails the job).

143 is what docker does when killed I think... does the initial `vault server` command somehow affect this?

lukpueh:
> 143 is what docker does when killed I think...

AFAIU not only a docker, but generally on linux. If a command receives SIGTERM it should exit with 143.
> 
> My guess is the initial `vault server` command has a signal handler that does exit 143


I think so too. But I cannot reproduce. I don't currently have a linux box at hand, and on my Mac the vault command does not exit with 143 when killed. (Interestingly, a different test command (`tail -f /dev/null`) indeed exits with 143 when killed).

I tried two independent workarounds to no avail:

https://github.com/secure-systems-lab/securesystemslib/blob/85c8d0ca46b16f2b1ef63c48277ac5f2a33733ed/tests/scripts/init-vault.sh#L3-L5

https://github.com/secure-systems-lab/securesystemslib/blob/85c8d0ca46b16f2b1ef63c48277ac5f2a33733ed/tests/scripts/stop-vault.sh#L6-L11


Any ideas, @jku?

lukpueh:
> > My guess is the initial `vault server` command has a signal handler that does exit 143
> 
> I think so too. But I cannot reproduce. I don't currently have a linux box at hand, and on my Mac the vault command does not exit with 143 when killed. (Interestingly, a different test command (`tail -f /dev/null`) indeed exits with 143 when killed).

I tried to debug this inside the GHA runner (with `tmate`) and `vault` exits with 0 there too when killed. This at least explains why my workarounds did not work.

jku:
Now that I think about it, I was probably wrong and this actually kills tox instead of "vault server":
```
pid=$(pgrep -f vault)
kill $pid
```

lukpueh:
> ... actually kills tox

🤦 or the stop-vault.sh

lukpueh:
@jku, this is ready for review. But we don't need to include it in 1.0.

tests/check_vault_signer.py
```diff
@@ -0,0 +1,38 @@
...
+            self.assertEqual(uri, f"{VaultSigner.SCHEME}:{name}/{version}")
+            self.assertEqual(public_key.scheme, scheme)
+
+            # Test load
+            signer = Signer.from_priv_key_uri(uri, public_key)
+            self.assertIsInstance(signer, VaultSigner)
+
+            # Test sign and verify
+            signature = signer.sign(b"DATA")
+            self.assertIsNone(public_key.verify_signature(signature, b"DATA"))
```
jku, line 32: nit: assert seems unnecessary
```suggestion
            public_key.verify_signature(signature, b"DATA")
```

jku (approved):
:+1:

jku:
> we don't need to include it in 1.0.

Up to you, no strong feelings here.

tests/check_vault_signer.py
lukpueh: 🤷 I wanted to make it clear that this is a test scenario, and not some random function call that could return e.g. True or False and pass the test either way.

lukpueh:
> > we don't need to include it in 1.0.
> 
> Up to you, no strong feelings here.

In that case, let's ship it.

PR #807, opened 2024-05-02
Release 1.0.0

### LINKED ISSUE #801: 1.0 release
I suppose we need an actual issue for this and not just the PR #767 .

Some possible blockers:
* docs: Remove pre 1.0  warnings (at least `signer` has one) _EDIT: done #802_
* docs: Document the Signer implementations (import methods, supported priv key urls, etc) in the reference docs #640
* docs: example app as documentation #633
* annotations: actually annotate the API, add py,typed  (see https://github.com/secure-systems-lab/securesystemslib/pull/767#issuecomment-2082169008)
* vaultsigner PR

I would say only the first one is *really* required.

lukpueh:
Fixes #801
Supersedes #767

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,38 @@
 # Changelog
 
+## securesystemslib v1.0.0
+
+### Added
+* Signer: add public_key attribute to interface (#756)
+* CryptoSigner: support ecdsa keytype that is no longer in spec (#711)
+* CryptoSigner: add private_bytes property (#799)
+* CryptoSigner: add `"file2"` signer uri (#759)
+* test: use localstack to test AWSSigner (#777)
+
+### Removed
+* CryptoSigner: remove `"file"` signer uri (#759)
+* migration script for legacy keys  (#770)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,38 @@
 # Changelog
 
+## securesystemslib v1.0.0
+
```
jku, line 4:
```suggestion
## securesystemslib v1.0.0

Securesystemslib API is now considered stable. The core functionality is provided in the _Signer_ interface and the half a dozen integrated _Signer_ implementations that can be found in the `signer` module. Smaller helper modules `dsse`, `formats`, `hash` and `storage` are also part of the API. Several legacy modules have been removed.
```

Could add a short overview maybe? I don't know how much value this brings

jku (approved):
It's a hard release to write changelogs for... but it looks good to me. I made one suggestion but I don't think it's necessary.

For the release announcement we should include a paragraph that explains what the library now is -- maybe the first paragraph from README is fine

lukpueh:
Rebased on main, included and re-formatted Jussi's open paragraph, and added a line about the just merged VaultSigner.

PR #829, opened 2024-06-26
Moving from pylint+black+isort+bandit to Ruff Linter

### LINKED ISSUE #440: Fix linter issues where needed
blocked on #270, partially addressed by #432 

**Current behavior**:
#439 turned on automatic linting with `pylint` and `bandit`, but inline-disabled them for all existing issues, above all to not waste time on code whose future is uncertain (see #270).

**Expected behavior**:
When resolving #270, we should not forget to address any of the remaining linter issues. They can be found by searching for the following inline code comments:

- `# nosec`
- `# pylint: disable=<msg>[, <msg>, ...]`, where `<msg>` is one of:

  ```
  abstract-method
  bad-classmethod-argument
  broad-except
  consider-using-enumerate
  consider-using-f-string
  consider-using-from-import
  consider-using-with
  dangerous-default-value
  global-statement
  global-variable-not-assigned
  implicit-str-concat
  import-error
  inconsistent-return-statements
  invalid-name
  logging-format-interpolation
  logging-not-lazy
  missing-class-docstring
  missing-function-docstring
  missing-module-docstring
  no-else-raise
  no-else-return
  no-member
  no-name-in-module
  protected-access
  raise-missing-from
  redefined-builtin
  redefined-outer-name
  redundant-unittest-assert
  singleton-comparison
  super-init-not-called
  super-with-arguments
  too-many-branches
  too-many-locals
  too-many-statements
  unnecessary-pass
  unneeded-not
  unspecified-encoding
  unsubscriptable-object
  unused-argument
  unused-import
  unused-variable
  use-implicit-booleaness-not-len
  useless-object-inheritance
  wrong-import-position
  ```

L77H:
- pylint, bandit, black and isort calls replaced by ruff calls in tox.ini
- ruff requirement added to requirements-lint.txt
- ruff config added to pyproject.toml: this covers currently only the functionality as before (isort, pylint and bandit), to be extended in the next PRs.
- code fixes made to comply with ruff rules
- _vendor exclusion added to mypy.ini
- black and isort config removed from pyproject.toml
- inline ruff error suppression added where necessary (noqa)

Fixes #440

(First ever PR, so feedback on mistakes more than welcome)


jku:
Looks like a good start, thanks.

I'll leave at least some comments today, might not have time for a full review . Just as a heads-up: I originally suggested doing small iterative PRs mostly because there might be quite a few comments if everything is done in one go... but we'll see how it goes, it looks manageable so far.

pyproject.toml
```diff
@@ -70,12 +70,35 @@ include = [
...
-extend-exclude="_vendor"
+# Ruff section
+[tool.ruff]
+lint.select = [
+    "I",      # isort: all
+    "PL",     # pylint: all
+    "S",      # flake8-bandit: all
+    "N",      # pep8-naming: all
+    "RUF100"  # ruff: find unused noqa
+]
```
jku, line 81: I would like to see at least E and F rulesets (the ruff default ones) enabled as well. 

Adding them leads to a bunch of `line-too-long` errors (apparently ruff is stricter than black when it comes to line length).  I wonder if we should just use the default 88 line-length since it looks like we have been fine with a lot of lines past 80... You could add E501 to ignore list for now: we can re-enable that afterwards or discuss changing the default line length next week, @lukpueh might have an opinion

pyproject.toml
```diff
@@ -70,12 +70,35 @@ include = [
...
+quote-style = "double"
+
+# Like Black, indent with spaces, rather than tabs.
+indent-style = "space"
+
+# Like Black, respect magic trailing commas.
+skip-magic-trailing-comma = false
+
+# Like Black, automatically detect the appropriate line ending.
+line-ending = "auto"
```
jku, line 99: do we need all of this configuration?  If not, I would rather not have it clutter pyproject.toml

jku (changes requested):
Looks pretty good and the speed is awesome. Thanks for the work,

If isort, black, pylint and bandit are no longer used, then most references should be removed as well. could do some grepping to find all cases but at least 
  * remove them from requirements-lint.txt
  * remove all `# pylint: disable=` comments in code 
  * remove references in `.github/dependabot.yml` (and add ruff instead)

I left some specific comments in-line as well

mypy.ini
```diff
@@ -5,6 +5,8 @@ files =
     securesystemslib/storage.py,
     securesystemslib/_gpg/constants.py
 
+exclude = securesystemslib/_vendor
```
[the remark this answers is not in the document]
jku: answering myself: 
* we have a `files` list mypy.ini already (just above this line): this was used before this PR.
* this PR adds files list as command line input in tox.ini, likely overriding the selection in mypy.ini

You could remove the files list from mypy.ini, that seems like the best option.

pyproject.toml
L77H: I will test it, and remove if not necessary

mypy.ini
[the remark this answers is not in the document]
L77H: I will remove the lines from mypy.ini. This makes it so that mypy will cover more files as well.

jku:
> After that in a new PR I can start adding more and more Rules to the Rules selector until we reach ["ALL"]. Now with all rules there are more than 2400 errors popping up, so I suggest to do it iteratively.

Yeah "ALL" is a bit demanding (even assuming a lengthy ignore list): it's a good goal to have but not required in any way. There's certainly rulesets that would be nice to enable so if you're willing to do more than this PR, that sounds great.

L77H:
Sure, if you can let me know which rule sets would be nice to have, I will be working on adding those! I also noticed that not all files in _gpg have type annotations yet, maybe in a next PR I can also work on completing that.

jku (commented):
Looks great. I have a couple of small requests:
* "/pylintrc" should still be removed from the sdist include in pyproject.toml -- this might break the release build otherwise
* `.pre-commit-config.yaml` looks like it should be modified based on this: I would say it's not required as part of this PR but could you file an issue so we don't forget?
* I filed an issue for the line length decision (so we don't just ignore E501 forever) #830

After this it looks good to merge

jku (approved):
LGTM, thank you

PR #836, opened 2024-06-27
pyupgrade "UP" linting rules integration, code fixes to comply

### LINKED ISSUE #834: linter: ruff coverage
The various linters have been replaced by ruff (thanks to @L77H).

* We should have a look at the ruff rulesets and decide which are most interesting (and best value for effort): https://docs.astral.sh/ruff/rules/
* Also should look at current ignores (both noqa comments and the ignore list in pyproject) and see if we can improve code 

As a initial list, these contain useful rules
* RUF
* UP: this contains "use f-string instead of format" error which we're going to see a lot of. `flynt` tool may allow automated conversion to f-strings but we need to be very careful with that
* pylint extended rulesets PLR, PLE, PLC, PLW
* various flake rulesets: ANN at least

I have not tested how much work enabling them would be -- starting with the easiest may be a good idea

L77H:

- `"UP"` added to selected rules in pyproject.toml
- fixed 44 code errors (UP032, UP012, UP008, UP004, UP031) to comply with UP rules

Fixes #834

### DIFF
--- a/securesystemslib/_gpg/common.py
+++ b/securesystemslib/_gpg/common.py
@@ -105,9 +105,8 @@ def parse_pubkey_payload(data):
     ptr += 1
     if version_number not in SUPPORTED_PUBKEY_PACKET_VERSIONS:
         raise PacketVersionNotSupportedError(
-            "Pubkey packet version '{}' not supported, must be one of {}".format(
-                version_number, SUPPORTED_PUBKEY_PACKET_VERSIONS
-            )
+            f"Pubkey packet version '{version_number}' not supported, "
+            f"must be one of {SUPPORTED_PUBKEY_PACKET_VERSIONS}"
         )
 
     # NOTE: Uncomment this line to decode the time of creation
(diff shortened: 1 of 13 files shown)


securesystemslib/_gpg/common.py
```diff
@@ -886,7 +877,7 @@ def parse_signature_packet(
     signature = handler.get_signature_params(data[ptr:])
 
     signature_data = {
-        "keyid": "{}".format(keyid),
```
jku, line 889: I suppose this could be just `"keyid": keyid,` but I I think as a pure conversion commit this is fine too

securesystemslib/formats.py
```diff
@@ -42,7 +42,7 @@ def _canonical_string_encoder(string):
       A string with the canonical-encoded 'string' embedded.
     """
 
-    string = '"%s"' % string.replace("\\", "\\\\").replace('"', '\\"')
+    string = '"{}"'.format(string.replace("\\", "\\\\").replace('"', '\\"'))
```
jku, line 45: this seems to be the important line in this patch -- everythings else is error messages or otherwise non-critical. This seems correct to me.

I assume the reason for using format() here is the "quote-quoting" that would be tricky with f-strings?

jku (approved):
This looks good to me and was not as scary as I thought it would be: almost every str.format() call happens when building error and log messages. The only interesting changes are 
* one critical line being changed in the canonicalization code: I left a comment there
* some gpg command line building: these seem ok too

@lukpueh I wouldn't mind you having a look as well

securesystemslib/_gpg/common.py
L77H: Yes, this line makes no sense like this, will change to `"keyid": keyid`

securesystemslib/formats.py
L77H: Yes the issue is with the `.replace('"', '\\"')`. I guess we can turn it into an f-string by escaping these quotes, using format might be more readable in the end though.

L77H: I am mistaken, backslashes are not allowed inside the f-string expression. Keeping it as is will be the most readable solution and is accepted by the linter.

lukpueh (approved):
Looks great!

PR #837, opened 2024-06-28
Abstract methods added to Key for signing scheme dissection

### LINKED ISSUE #594: signer: deduplicate signing scheme dissection
There is some duplicate code in `securesystemslib.signer` to dissect signing schemes, in order to figure out the 
* appropriate hash algorithm: e.g. "rsassa-pss-sha384" and "ecdsa-sha2-nistp384" use "sha384"), or
* padding: e.g. "rsa-pkcs1v15-sha224 and "rsa-pkcs1v15-sha512" use "pkcs1v15"

It would be nice to provide a signer-wide helper function or map.

hint: the code can be found by grepping for `split`, also in pending PRs: #585, #588, #590. Note that we usually first extract the substring (which is the same for all schemes) and then map them e.g. to specific hash constants or objects (which differs between signers).

related: #593

L77H:
- `get_hash_algorithm_str` abstract method added to Key class
- `get_hash_algorithm` abstract method added to Key class
- `get_padding_name_str` abstract method added to Key class
- `get_padding_name` abstract method added to Key class
- removed duplicate code

This PR tries to make classes using the Key instance independent of the specific Key implementation. 
It also includes the changes from PR #836.

Fixes #594


securesystemslib/signer/_key.py
```diff
@@ -196,6 +203,32 @@ def verify_signature(self, signature: Signature, data: bytes) -> None:
...
+        """Return payload padding name used for this key as a str"""
+
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_padding_name(self, hash_algorithm: Any) -> Any:
+        """Return payload padding name used for this key as a AsymmetricPadding"""
+
+        raise NotImplementedError
+
```
lukpueh, line 231: I can't really see a benefit of adding these to the Key interface. The only subclass, which implements them, is SSlibKey. Why don't make these SSlibKey-only methods?

L77H: In `_azure_signer.py` and `_gcp_signer.py` both `AzureSigner` and `GCPSigner` take in a public key of type `Key` on which they use the get hash method. Can this be changed to SSlibKey?

lukpueh: Hm. Good question. It looks like all `Signer` implementations take `Key`s, even though they can only handle specific Key implementations. That's probably okay. And I definitely don't want to disrupt anyone, by changing it.

But adding new behaviour to an abstract base class, which is specific to one subclass only, and raising NotImplementedError everywhere else, does not feel right.

Would it be okay to add regular functions, e.g. to _utils? 

 Thoughts, @jku?

jku: I admit I didn't look at the specific uses in AzureSigner and GCPSigner yet but I would say it's fine for them to handle SSlibKey only -- that sounds correct to me, what else could the key be?

Whether that should be done in practice by type checking inside the AzureSigner and GCPSigner functions or by changing some public argument types I can't say without having a closer look...

L77H: In my last commit the methods are now SSlibKey-only methods. I added the stricter type checking on GCPSigner and AzureSigner and some of their methods including a method from the Signer class. It seems to pass all the tests, but a more detailed review by @lukpueh and @jku would be advised.

lukpueh: Looks great! If the change isn't likely to bother any user, then I'd go with it. I'll give a detailed review after #836 is merged.

securesystemslib/signer/_key.py
```diff
@@ -56,6 +58,11 @@
 
 logger = logging.getLogger(__name__)
 
+
+class UnsupportedKeyType(Exception):  # noqa: N818
+    pass
```
jku, line 63: this could be marked private (_UnsupportedKeyType) since I don't think we intend to leak it outside... but that's a nit

jku (approved):
Looks good to me, thanks!

I did leave one comment but I really don't think it's that important, this looks good to merge

securesystemslib/signer/_gpg_signer.py
```diff
@@ -55,6 +55,18 @@ def verify_signature(self, signature: Signature, data: bytes) -> None:
...
+        raise NotImplementedError
+
+    def get_hash_algorithm(self) -> None:
+        raise NotImplementedError
+
+    def get_padding_name_str(self) -> None:
+        raise NotImplementedError
+
+    def get_padding_name(self, hash_algorithm: None, salt_length: None) -> None:
+        raise NotImplementedError
```
lukpueh, line 68: These stubs are no longer needed, right?

lukpueh (changes requested):
Looks mostly good. Please address inline comment and consider updating the PR title.

lukpueh:
There are a few subtle issues related to error handling in GCPSigner and AzureSigner i this PR. I created a new PR based on this work: #981.

Thanks for your efforts, @L77H!

Issue #158, opened 2018-10-16
Doc improvement for formats._canonical_string_encoder / encode_canonical()


awwad:
@lukpueh has correctly pointed out that [`formats._canonical_string_encoder()`'s docstring](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/formats.py#L688-L709) is bad.  It would be better not to have one, but better yet for it to be clear.  It should be substantially more specific about the behavior the function intends and provides and why. It seems that [this](http://wiki.laptop.org/go/Canonical_JSON) canonicalization policy is part of the motivation.

The docstring for `formats.encode_canonical()` should also be adjusted slightly to explain this policy.

PR #843, opened 2024-07-03
type annotations for formats.py & tried to improve docstring

### LINKED ISSUE #358: Add type annotations

L77H:
- type annotations for functions in formats.py
- change to signer/_utils.py to comply with new type requirements
- tried to make the function docstring for `_canonical_string_encoder` and `encode_canonical` a bit more clear

Fixes #358 #158

### DIFF
--- a/securesystemslib/formats.py
+++ b/securesystemslib/formats.py
@@ -20,13 +20,18 @@
 
 """
 
+from typing import Callable, Optional, Union
+
 from securesystemslib import exceptions
 
 
-def _canonical_string_encoder(string):
+def _canonical_string_encoder(string: str) -> str:
     """
     <Purpose>
-      Encode 'string' to canonical string format.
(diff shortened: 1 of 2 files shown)


jku (approved):
lgtm, thanks

PR #886, opened 2024-10-08
Update Python versions

jku:
3.8 is EOL, 3.13 has been released.

* Test on 3.13, don't test on 3.8
* Test Mac on 3.9 (since that's Apples Python version)
* Do not require 3.9 yet: I would like to do this but it makes sense to update the annotations at the same time
* Stop encoding python versions in tox.ini: I think this not very useful and leads to versions in pyproject.toml and tox.ini being not in sync


jku:
Hmm, GitHub does not have 3.13 yet. Marking this draft

jku:
oh this also needs a project/org admin to fiddle with the required checks (see "6 expected checks" below)

jku:
I've updated the checks to have "3.13" in the name and filed #893 for future improvement.

This PR is ready for review

adityasaky (approved):

PR #888, opened 2024-10-12
VaultSigner test unexpectedly does not depend on environment variables

### LINKED ISSUE #820: VaultSigner test unexpectedly does not depend on environment variables
The HashicorpVault client used by VaultSigner to [`import_`](https://github.com/secure-systems-lab/securesystemslib/blob/5e734e44b1df536f0fe6205a90cfc0ec1c4e8bd7/securesystemslib/signer/_vault_signer.py#L122) a public key or [`sign`](https://github.com/secure-systems-lab/securesystemslib/blob/5e734e44b1df536f0fe6205a90cfc0ec1c4e8bd7/securesystemslib/signer/_vault_signer.py#L54), needs a vault address and a token. We expect users to provide both via specific envvars known to the client, and so do the tests:

https://github.com/secure-systems-lab/securesystemslib/blob/5e734e44b1df536f0fe6205a90cfc0ec1c4e8bd7/tox.ini#L122-L123

But it turns out that the tests work even if the envvars are not available to VaultSigner, because:
- above VAULT_ADDR is the default address, which the client falls back to, if the env var is not set, and
- `vault server -dev` (in [`init-vault.sh`](https://github.com/secure-systems-lab/securesystemslib/blob/5e734e44b1df536f0fe6205a90cfc0ec1c4e8bd7/tests/scripts/init-vault.sh#L3)) secretly writes the token to a `~/.vault-token` file, which the client falls back to, if the env var is not set

This is not necessarily an issue but a bit surprising, so I think a note in VaultSigner code doc and tox.ini might be fix enough.

Alternatively, we can set VAULT_ADDR to a non-default port and start the test server like so:
```bash
vault server -dev \
    -dev-listen-address="${VAULT_ADDR#http://}" \  # unlike VAULT_ADDR this option does not want the protocol prefix
    -dev-root-token-id="${VAULT_TOKEN}" \
    -dev-no-store-token \
    &
```

NicholasTanz:

Fixes: #820


tox.ini
```diff
@@ -118,7 +118,7 @@ allowlist_externals =
     bash
 
 setenv =
-    VAULT_ADDR = http://localhost:8200
+    VAULT_ADDR = http://localhost:8201
```
NicholasTanz, line 121: not sure if there was a specific port I should've used? I just did +1 from the default port

NicholasTanz:
Hi @jku, I'm running into an issue where the `test-vault.yaml` workflow is stuck on the `run tests` step and am not sure why this is occurring. Any input is appreciated, thanks

tests/scripts/init-vault.sh
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 
-vault server -dev -dev-root-token-id="${VAULT_TOKEN}" &
+vault server -dev \
+    -dev-listen-address="${VAULT_ADDR#http://}" \  # unlike VAULT_ADDR this option does not want the protocol prefix
+    -dev-root-token-id="${VAULT_TOKEN}" \
+    &
 
 until vault status
 do
```
NicholasTanz, line 9: yeah, I feel like I didn't setup the vault server properly, as I think it's stuck in this do-until loop

kairoaraujo:
> I have to admit I know very little about vault or the test setup here. @kairoaraujo might have more to say here?

Added to my TODO.
I will have a look.

tests/scripts/init-vault.sh
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 
-vault server -dev -dev-root-token-id="${VAULT_TOKEN}" &
+vault server -dev \
+    -dev-listen-address="${VAULT_ADDR#http://}" \
```
NicholasTanz, line 4: I figured out the issue... the comment on this line was causing the command to be interpreted incorrectly :( lol

NicholasTanz:
Hi @jku @kairoaraujo when one of you have a moment, can you take a look at this PR. thanks

PR #914, opened 2024-11-29
Use __future__.annotations module

jku:
The extra import is a little annoying but allows using even python 3.10 annotation features already, which I like.
* start using "X | None" instead of Optional[X]
* Drop the quotes from factory function return values

This also keeps us compatible with python 3.8 for the moment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ classifiers = [
     "Topic :: Security",
     "Topic :: Software Development",
 ]
-requires-python = "~=3.9"
+requires-python = "~=3.8"
 dynamic = ["version"]
 
 [project.urls]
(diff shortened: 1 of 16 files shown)


jku:
Not quite sure who is available for review: flagged @adityasaky but no problem if you're not available

jku:
would make sense to drop this back to 3.8 now :

`requires-python = "~=3.9"`

adityasaky (approved):
This looks good to me though I must note I may not qualify as an "expert reviewer" in python's type hints.

Also, is support for 3.8 a requirement somewhere?

jku:
> Also, is support for 3.8 a requirement somewhere?

No, and I would even say it's not supported: I just don't want to be the first one to explicitly break potential 3.8 users if it's easy to avoid.

PR #926, opened 2024-12-19
github status check for whole test run

### LINKED ISSUE #893: Create a Github status check for whole test run
Currently our required status checks look like this:
Screenshot from 2024-10-15 18-05-37

this is very annoying as soon as a new Python version is released as the check names change. I believe we could fix this by adding another job to the test workflow that requires the whole matrix to succeed: this should show up as a single check that won't change its name.

NicholasTanz:

Fixes #893

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -81,3 +81,10 @@ jobs:
 
       - name: Run tox
         run: tox -e ${{ matrix.toxenv }}
+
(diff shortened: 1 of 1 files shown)


jku:
Thanks! I'll play with the repository settings a bit, maybe I can test the status requirements before merging...

jku (approved):

PR #958, opened 2025-02-28
add zizmor for linting workflows.

### LINKED ISSUE #954: Use zizmor to lint workflows
It would be good if our GitHub workflows were statically analyzed with zizmor: https://github.com/woodruffw/zizmor

Similar change was recently done in python-tuf: https://github.com/theupdateframework/python-tuf/pull/2798
It's likely that the same approach will work:
* add zizmor (with current version) to requirements-lint.txt
* Add `zizmor --persona=pedantic -q` call in lint section of tox.ini
* fix issues reported by zizmor when `tox -e lint` runs: 
  * looks like zizmor reports 18 findings currently: most are easy to fix
  * please paste specific error in a comment here if it's not obvious how to deal with it

NicholasTanz:

Fixes #954


.github/workflows/cd.yml
```diff
@@ -106,13 +111,15 @@ jobs:
       - name: Finalize GitHub release
         uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
         with:
-          script: |
+          script: | # zizmor: ignore[template-injection]
```
NicholasTanz, line 114: I ended up ignoring the `needs.candidate_release.outputs.release_id` zizmor warning, since it came from a previous step and the audit confidence from zizmor was low

.github/workflows/check-upstream-ed25519.yml
```diff
@@ -30,7 +32,7 @@ jobs:
       uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea
       if: ${{ steps.test_ed25519.outputs.result == '1' }}
       with:
-        script: |
+        script: | # zizmor: ignore[template-injection]
```
NicholasTanz, line 35: same with this one here with `steps.test_ed25519.outputs.output`

jku (approved):
Nice, thanks. LGTM

PR #969, opened 2025-03-15
simplify securesystemslib.hash

### LINKED ISSUE #943: simplify hashing
Currently the hash module in securesystemslib supports multiple hash libraries:
* This seems bonkers to me and I've never seen anyone use this functionality
* It also makes type annotating the module a pain: this is pretty much the reason why securesystemslib is not yet marked as annotated (there are other unannotated modules but they are not as core the the library as `hash` is)

I think we can keep providing most of the API, but remove the cryptography support (from hashes module) and use only hashlib.

jku:
* Stop supporting multiple hash libraries. This is an API break.
* Mark the library as annotated, be more strict in mypy setup

This is likely not quite finished, more of a discussion item

Fixes #943


jku:
huh, clearly this was not ready even for discussion

lukpueh:
> huh, clearly this was not ready even for discussion

why not?

jku:
Meaning, I would rather continue with just "annotating" current hash,module as in the newer pr

PR #970, opened 2025-03-15
Annotate the library (add py.typed)

jku:
I'd like to statically analyze calls to securesystemslib in other projects
* securesystemslib needs a py.typed file for that
* securesystemslib.hash is the only public module that is not annotated: current plan is to deprecate it before next release

Changelog:
* Add py.typed
* refactor mypy config into pyproject.toml
* make mypy config stricter but don't try to lint tests with it
* Run lint on oldest supported python version to let mypy find annotation issues on that python
* Tell mypy to ignore hash module
* Fix a bunch of small annotation issues elsewhere in code


securesystemslib/storage.py
```diff
@@ -189,9 +189,9 @@ class FilesystemBackend(StorageBackendInterface):
     # objects.
     _instance = None
 
-    def __new__(cls, *args, **kwargs):
+    def __new__(cls) -> FilesystemBackend:
         if cls._instance is None:
-            cls._instance = object.__new__(cls, *args, **kwargs)
+            cls._instance = object.__new__(cls)
         return cls._instance
```
jku, line 195: if there's a neater way to handle this I'm happy to hear it -- maybe we just want to ignore mypy here since `__new__()` has some magic for silently accepting kwargs...

jku: I will do this to keep the PR minimal

jku:
Only change in this force-push is that `FilesystemBackend.__new__()` signature now remains what it was: I just used more "typing.Any".

jku:
I will modify this so hash is left untouched by this PR: based on https://github.com/theupdateframework/python-tuf/pull/2815 we can deprecate hash instead.

jku:
This should be good to go now. python-tuf lint (with this securesystemslib, with mypy excludes removed) has some complaints but only in tests and seemingly for good reason.

lukpueh (approved):

Issue #593, opened 2023-05-31
signer: add constants for key types and signing schemes


lukpueh:
Supported key types and signing scheme strings are currently hardcoded all over the library. We should define them in a single place e.g. `securesystemslib.signer._constants` and update all `secureystemslib.signer.*` modules to use them. (I suggest to not touch legacy modules such as `secureystemslib.keys`)

Publicly registered key types and schemes can be found in:
https://github.com/secure-systems-lab/securesystemslib/blob/135567fa04f10d0c6a4cd32eb45ce736e1f50a93/securesystemslib/signer/__init__.py#L39-L55

Additionally, there are a few non-registered ones, which ca be found by looking for implementations of the `Key` interface.

lukpueh:
The 3 original key types are already defined as constants in __init__, which are only used in `securesystemslib.interface`. I suggest to deprecate them in favour of constants defined in the signer sub-package.

PR #977, opened 2025-03-19
Remove internal use of securesystemslib.hash

lukpueh:
fixes https://github.com/secure-systems-lab/securesystemslib/issues/943

* Internally, the features added by the module (i.e. custom blake name
  support, and file hashing) are not needed. Builtin hashlib is used now
  directly.

* External users port the relevant features into their code base:
  https://github.com/theupdateframework/python-tuf/pull/2815
  https://github.com/in-toto/in-toto/pull/861


securesystemslib/signer/_gcp_signer.py
```diff
@@ -180,7 +180,7 @@ def _get_hash_algorithm(public_key: Key) -> str:
             )
 
         # trigger UnsupportedAlgorithm if appropriate
```
lukpueh, line 182: This is now a ValueError, which breaks the API and is also inconsistent with above, where we raise an UnsupportedAlgorithmError for some unsupported algorithms. Same is true for other signers in this PR

I'll mark the PR as draft until I fix this.

lukpueh: Turns out, other signers exhaustively check public key schemes prior to using the hash algorithm name. This menas `UnsupportedAlgorithmError` is not expected when a digest object is created in those signers.

d5016ae adds a similar scheme check to GCPSigner to make use of the already existing `raise UnsupportedAlgorithmError`, previously only used for invalid key types.

 f5f162f proposes an alternative, less invasive but IMO more clunky fix, which just re-raises hashlib's exception as `UnsupportedAlgorithmError`. 

Either way, we may want to consider a more comprehensive solution in #593 and #766.

securesystemslib/signer/_gcp_signer.py
```diff
@@ -163,24 +163,22 @@ def _get_hash_algorithm(public_key: Key) -> str:
...
+        ) or (
             # nistp256 uses sha-256, nistp384 uses sha-384
-            bits = public_key.scheme.split("-nistp")[-1]
-            algo = f"sha{bits}"
+            # TODO: Check for invalid type/scheme combinations (#766)
+            public_key.keytype
+            in ["ecdsa", "ecdsa-sha2-nistp256", "ecdsa-sha2-nistp384"]
+            and public_key.scheme.endswith(("256", "384"))
+        ):
+            algo = public_key.scheme[-3:]
```
jku, line 175: I'm a little surprised if this works since it seems to change the algo value ("sha256" -> "256"). 

The GCP tests currently only run on push so you can trigger it by pushing the branch to main repo

I'm fine with the simpler solution too.

lukpueh: Good catch. 🤦  That happens when code isn't tested right away.

jku (approved):
lgtm. We can look at improving the tests if we actually end up improving the scheme parsing globally.

PR #981, opened 2025-03-26
Refactor scheme parsing 

### LINKED ISSUE #594: signer: deduplicate signing scheme dissection

lukpueh:
Fixes #594

This is an attempt to consolidate SSlibKey type and scheme parsing, to extract hash algo and padding names. It adds two new public methods on SSlibKey: (5fbeeaf17a07aede0b0bf2215e3fdf79874c93fc)
- `def get_hash_algorithm_name(self) -> str:`
- `def get_padding_name(self) -> str:`

and an internal helper to get pyca/cryptography hash objects from hash algo names: (166e457593d33286a9a200c62ee9077d5d4986c8)
- `def get_hash_algorithm(name: str) -> HashAlgorithm:`

Unfortunately, the diff is bigger than I wanted, and we don't really cut down on lines of code either. This is because there are two important prerequisites to use the new `SSlibKey` methods in the relevant signers
1. The public key attributes on the Signers need to be narrowed down from `Key` to `SSlibKey` type (3c6a34da1cc5ac4c98521c7cc123a104533d6978)
2. AzureSigner and GCPSigner relied on their custom scheme dissection helpers to validate the passed scheme. Since those signers, support only a subset of the schemes supported by SSlibKey, additional validation had to be added. (48c7523a2209491119ed5bf4be77703f89628812)

I think there's a value in parsing the schemes in one place only, and making scheme valdiation more explicit. But I am also fine, if others don't think it's worth such a big code change.

### DIFF
--- a/securesystemslib/signer/_gcp_signer.py
+++ b/securesystemslib/signer/_gcp_signer.py
@@ -149,7 +149,7 @@ class GCPSigner(Signer):
         request = {"name": gcp_keyid}
         kms_pubkey = client.get_public_key(request)
         try:
-            keytype, scheme = cls._get_keytype_and_scheme(kms_pubkey.algorithm)
+            keytype, scheme = KEYTYPES_AND_SCHEMES[kms_pubkey.algorithm]
         except KeyError as e:
             raise exceptions.UnsupportedAlgorithmError(
                 f"{kms_pubkey.algorithm} is not a supported signing algorithm"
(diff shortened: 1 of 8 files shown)


lukpueh:
GCP tests are passing: https://github.com/secure-systems-lab/securesystemslib/actions/runs/14085915617/job/39449808178

securesystemslib/signer/_gcp_signer.py
```diff
@@ -112,81 +164,7 @@ def import_(cls, gcp_keyid: str) -> tuple[str, Key]:
     @staticmethod
     def _get_keytype_and_scheme(algorithm: int) -> tuple[str, str]:
```
jku, line 165: nit: method is probably unnecessary if it's a oneliner

jku (approved):
I think it looks right... left two comments about maybe unnecessary functions but seems fine to me.

I guess we don't have any azure tests but that's not your fault

PR #988, opened 2025-04-11
Prepare release v1.3.0

### LINKED ISSUE #987: New release?
Release is useful to actually use the annotations externally. It's also been 4 months so maybe time to do point release.

TODO:
* maybe mark `hash` module more clearly as deprecated?

lukpueh:
Fixes #987

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## securesystemslib v1.3.0
+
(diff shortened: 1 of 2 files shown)


jku (approved):
From issue:

> ... adding the note would be helpful to users, who look up the docstrings in code or in the IDE. I'm okay, to add it. If others think it's necessary.

either way works for me. I don't think there are users that will be negatively surprised ... and would actually read a docstring

PR #1030, opened 2025-09-24
Prepare release 1.3.1

jku:
Not a lot here but it's been a few months since previous release, doesn't hurt to do a release.

* There is an open PR for dependency updates but that's for pinned deps: these are not used in the released package
* Looking at git log, I used way too much "squash and merge" this summer: will avoid that in future unless there's a good reason as it seems to make git archaeology more difficult

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ keywords = [
     "ecdsa",
 ]
 classifiers = [
-    "Development Status :: 4 - Beta",
+    "Development Status :: 5 - Production/Stable",
     "Intended Audience :: Developers",
     "Natural Language :: English",
     "Operating System :: POSIX",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -18,7 +18,7 @@ keywords = [
     "ecdsa",
 ]
 classifiers = [
-    "Development Status :: 4 - Beta",
+    "Development Status :: 5 - Production/Stable",
```
JustinCappos, line 21: Moving to Prod / Stable makes sense given our user base seems quite stable.  Any concerns from anyone?  I just wanted to flag this for further comment...

JustinCappos (approved):

pyproject.toml
jku: Oh right, I didn't mention that: I figured this was just forgotten since 1.0 in May 2024 was supposed to mark stability...

PR #1107, opened 2026-03-25
feat: support multiple bytes key_id for hsm signer (cka_id)

Spinbazz:

Fix: remove keyid size limitation to 1 byte (FF)

Fixes #1105

### DIFF
--- a/securesystemslib/signer/_hsm_signer.py
+++ b/securesystemslib/signer/_hsm_signer.py
@@ -210,11 +210,16 @@ class HSMSigner(Signer):
         if key_type is None:
             key_type = PyKCS11.CKO_PUBLIC_KEY
 
+        keyid_hexstr = hex(keyid)[2::]
+        cka_id_filter = [
+            int(keyid_hexstr[i : i + 2], 16) for i in range(0, len(keyid_hexstr), 2)
+        ]
+
         keys = session.findObjects(
             [
                 (PyKCS11.CKA_CLASS, key_type),
                 (PyKCS11.CKA_KEY_TYPE, PyKCS11.CKK_ECDSA),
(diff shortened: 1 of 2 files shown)


securesystemslib/signer/_hsm_signer.py
```diff
@@ -210,11 +210,16 @@ def _find_key(
         if key_type is None:
             key_type = PyKCS11.CKO_PUBLIC_KEY
 
+        keyid_hexstr = hex(keyid)[2::]
+        cka_id_filter = [
+            int(keyid_hexstr[i : i + 2], 16) for i in range(0, len(keyid_hexstr), 2)
+        ]
```
jku: I think this bit is buggy: 
```
>>> hex(255)[2::]
'ff'
>>> hex(256)[2::]
'100'
```
that will screw up the filter AFAICT. The latter leads to `cka_id_filter = [16, 0]` which sounds wrong.

using `keyid.to_bytes()` instead might be simpler (but still requires figuring out the length).

jku (changes requested):
apologies for the silence, did not mean to leave you on unread for so long. The basic idea makes sense to me: the search filter is actually a byte array and current code fails if keyid > 0xFF.

I left one comment in code -- that looks like a hex conversion bug to me. Maybe adding a test with odd and even length hex string ids would make sense? let me know if you're not invested enough, I can look at adding that.

securesystemslib/signer/_hsm_signer.py
jku: Since you allow pushes to your branch (thanks), I'll fix this and add a test

jku (approved):
looks good to me now -- happy to get another review from a maintainer or @Spinbazz

PR #1109, opened 2026-03-25
feat: allow hsm login before import pubkey (pubkey may be inaccessibl…

Spinbazz:
…e for unlogged client)

Add a call of login pykcs11 function before searching pub key on import function

Fixes #1108


securesystemslib/signer/_hsm_signer.py
```diff
@@ -303,8 +304,9 @@ def import_(
             token_filter = cls._build_token_filter()
 
         uri = f"{cls.SCHEME}:{hsm_keyid}?{parse.urlencode(token_filter)}"
-
         with cls._get_session(token_filter) as session:
+            if pin_handler is not None:
+                session.login(pin_handler(cls.SECRETS_HANDLER_MSG))
```
jku: this sounds like if the app provides pin_handler, all users now get a pin request even though some of them don't need it.  I assume we could avoid that somehow

Spinbazz: I propose to call pin handler only if unlogged public key research has failed

Spinbazz: Hi @jku can you review my PR, i just added unittest.

jku:
Thanks. If you fix the last lint error ("tox -e fix" should just work) I think I can merge this. However...

I finally got to testing `python-pkcs` as a `pykcs11lib` replacement in #1156 and I think it looks much nicer than the current code (and is typically installable without compiling which would be great). So if you don't want to fix the last lint issue here, that's fine too: I will include the equivalent change in #1156 (I'll set co-authored-by so you'll get credit). 

Also, if you have experience with these libraries, I appreciate any comments in #1156.

jku:
Also let me know if you'd like a release with this fix soon -- I might not do one before August otherwise but I realize this PR has been sitting here for a while...

jku (approved):
Since your branch allows it I went and fixed the whitespace issue -- I usually avoid adding fixes as it means I'm then not eligible to approve the PR... but I think I can override the rules in this case. 

I think the PR looks fine, let's try it.

PR #1119, opened 2026-04-27
tests: Switch sigstore test token  source

jku:
This one should be more reliable


jku:
leaving draft until I'm 100% sure this testing token URL is the final one

jku:
Marking this ready for review: sigstore-conformance also now uses this URL so it should be good to use

PR #1120, opened 2026-04-27
Remove AWS tests from CI

jku:
Localstack now requires an account:
https://github.com/secure-systems-lab/securesystemslib/issues/1104

Disabling tests on CI until the tests are fixed.

---

I'm happy to withdraw this if someone is looking into fixing the tests


jku:
sigstore test is failing because it's missing #1119 ... which is failing because it's missing this PR :(

PR #1124, opened 2026-05-12
Add ML-DSA support

### LINKED ISSUE #1123: Support ML-DSA
* Cryptography 47/48 has added support for ML-DSA post quantum signing algorithm
* There is a TAP for the TUF keytype "mldsa" https://github.com/theupdateframework/taps/pull/195

Supporting this in securesystemslib is pretty easy, I have a draft branch already.

jku:
This implements a proposed [ML-DSA TAP](https://github.com/theupdateframework/taps/pull/195): I think it's ready for review.

This currently enables the ML-DSA algorithms by default for signing but not for verify: this should allow merging before the TAP is finalized so we can add support in tuf-conformance.

The only unusual thing about the new key is the _custom pre-hash with domain prefix_: this is documented in the TAP. 

Support is added to
* SSLibKey (Verification in general)
* Cryptosigner (Signing with keys-on-disk)
* GCPsigner (Signing with Google Cloud KMS)

No dependencies are added but minimum cryptography version is bumped to 48: ML-DSA support was added in 47 already but support via openssl only became available in 48

The test code is from AI.


jku:
Test run on GCP: https://github.com/secure-systems-lab/securesystemslib/actions/runs/25749092040/job/75620285276: looks ok.

There is a test TUF repo https://github.com/jku/pqc-test-repo2
* online ML-DSA keys using the GCP support in this PR
* offline ML-DSA keys using #1149

securesystemslib/signer/_crypto_signer.py
```diff
@@ -136,8 +157,7 @@ def __init__(
             "rsa-pkcs1v15-sha384",
             "rsa-pkcs1v15-sha512",
         ]:
-            if not isinstance(private_key, RSAPrivateKey):
-                raise ValueError(f"invalid rsa key: {type(private_key)}")
+            assert_type("rsa", private_key, RSAPrivateKey)
```
jku, line 151: all these assert tweaks are just to keep linter happy ("too many branches"): the functionality should not change

securesystemslib/signer/_crypto_signer.py
```diff
@@ -71,6 +78,14 @@ class _NoSignArgs:
     pass
 
 
+@dataclass
+class _MLDSASignArgs:
+    prefix: bytes
+
+    def __init__(self, version: int) -> None:
+        self.prefix = b"tuf" + bytes([version])
```
jku: arguably the "pre-signing-prefix" is really a public key property, not a signing argument: that way we wouldn't have to create it in several places

jku: decided to just simplify and hard code as long as version is is not actually dynamic

jku:
I'll mark this ready for review: We should make a decision on if we really enable verification by default before merging/releasing.

securesystemslib/signer/_crypto_signer.py
```diff
@@ -321,5 +338,14 @@ def generate_ecdsa(
         return CryptoSigner(private_key, public_key)
 
     def sign(self, payload: bytes) -> Signature:
-        sig = self._private_key.sign(payload, *astuple(self._sign_args))  # type: ignore
+        if isinstance(
+            self._private_key, (MLDSA44PrivateKey, MLDSA65PrivateKey, MLDSA87PrivateKey)
+        ):
```
jku: using `isinstance(...)` here instead of the clearer `self.keytype == "ml-dsa"` just means the sign() call does not need a mypy ignore

jku:
rebased on main

jku (commented):
After a quick attempt at making this work in tuf-conformance:
* there is no `CryptoSigner.generate_mldsa()` -- this is only an inconvenience
* but also `SSLibKey.from_crypto()` does not seem to support ml-dsa yet

EDIT: actually the latter is in the PR so only the inconvenience remains... I think I still want that in the PR

securesystemslib/signer/_crypto_signer.py
jku: I will switch this around: I'd rather take the code clarity and accept the ignore (since the other branch has an ignore anyway)

securesystemslib/signer/_crypto_signer.py
```diff
@@ -320,6 +337,46 @@ def generate_ecdsa(
...
+        Returns:
+            CryptoSigner
+        """
+        if CRYPTO_IMPORT_ERROR:
+            raise UnsupportedLibraryError(CRYPTO_IMPORT_ERROR)
+
+        scheme = "ml-dsa-65/1" if scheme is None else scheme
+        if scheme == "ml-dsa-44/1":
+            private_key: PrivateKeyTypes = MLDSA44PrivateKey.generate()
+        elif scheme in "ml-dsa-65/1":
```
loosebazooka: `in` -> `==`

securesystemslib/signer/_gcp_signer.py
```diff
@@ -59,6 +60,18 @@
             "rsa",
             "rsa-pkcs1v15-sha512",
         ),
+        CryptoKeyVersion.CryptoKeyVersionAlgorithm.PQ_SIGN_ML_DSA_44: (
```
loosebazooka, line 63: gemini seems to think this could break older google-cloud-kms installs that have not included these constants yet (even if they aren't using the PQ parts of this library). I'm not exactly sure how python works in this case and whatever way it works, I'm not sure it even matters. People can maybe just update.

tests/test_signer.py
loosebazooka: should we add ml-dsa tests here?

securesystemslib/signer/__init__.py
```diff
@@ -69,5 +69,8 @@
 )
 
 # Keys with currently unstable metadata formats, not supported by default:
+#       ("ml-dsa", "ml-dsa-44/1"): SSlibKey,
```
loosebazooka: should these link to https://github.com/theupdateframework/taps/blob/master/tap21.md#v1 ? My understanding is that `.../1` is a TUF specific thing and is otherwise arbitrary?

loosebazooka (commented):
Looks mostly fine to me. Is the KMS stuff not tied to your account anymore?

securesystemslib/signer/_gcp_signer.py
jku: I can set a minimum version for google-cloud-kms, that seems right (even though I'd argue you should anyway always use the most recent one of this sort of things)

securesystemslib/signer/__init__.py
jku: yeah, this is not a bad idea

loosebazooka (approved):
here's my gray lgtm.

jku:
> Is the KMS stuff not tied to your account anymore?

it's just the CI tests for the gcp signer that run on my personal GCP project: It's my 0.45€ monthly donation to the project

securesystemslib/signer/_key.py
```diff
@@ -383,6 +398,7 @@ def _verify_ed25519_fallback(self, signature: bytes, data: bytes) -> None:
 
     def _verify(self, signature: bytes, data: bytes) -> None:
         """Helper to verify signature using pyca/cryptography (default)."""
+        # ruff: noqa: PLR0915
```
lukpueh: nit: apparently this is a [file-scope exemption](https://docs.astral.sh/ruff/linter/#file-level). It probably does not matter, but if you want to only exempt the function, you could use [disable/enable markers](https://docs.astral.sh/ruff/linter/#block-level).

securesystemslib/signer/_crypto_signer.py
```diff
@@ -320,6 +337,46 @@ def generate_ecdsa(
...
+
+        public_key = SSlibKey.from_crypto(private_key.public_key(), keyid, scheme)
+        return CryptoSigner(private_key, public_key)
+
     def sign(self, payload: bytes) -> Signature:
+        if self.public_key.keytype == "ml-dsa":
+            # ml-dsa keytype specifies a domain-specific hash prefixing scheme
+            digest = Hash(SHA512())
+            digest.update(payload)
+            payload = b"tuf" + bytes([1]) + digest.finalize()
```
lukpueh: TAP-21 §Verification 4 ays "The version MUST be taken from the trusted TUF metadata's scheme parameter."

Hard-coding the version here relies on other code rejecting schemes that don't end in `/1`, which is the case today. So it's probably oaky, and I saw that the hardcode was a deliberate decision. Still worth  keeping in mind the potential for drift.

Same is true in _gcp_signer.py and _key.py.

securesystemslib/signer/_crypto_signer.py
```diff
@@ -23,6 +23,11 @@
     from cryptography.hazmat.primitives.asymmetric.ed25519 import (
         Ed25519PrivateKey,
     )
+    from cryptography.hazmat.primitives.asymmetric.mldsa import (
+        MLDSA44PrivateKey,
+        MLDSA65PrivateKey,
+        MLDSA87PrivateKey,
+    )
```
lukpueh, line 31: This will raise "'pyca/cryptography' library required" in some of the supported cryptography libs. Current floor in pyprojects.toml is 40.0.0, but mldsa doesn't exist before 48.0.0.

Could raise the floor, or use a dedicated error message for mldsa.

Same is true for imports in _key.py

securesystemslib/signer/_gcp_signer.py
lukpueh: According to [the docs](https://docs.cloud.google.com/kms/docs/reference/rpc/google.cloud.kms.v1#getpublickeyrequest) you need to pass `public_key_format` for PQ keys or else it errors. But it seems to work regardless 🤷

lukpueh (changes requested):
Full disclosure: I asked Claude Code Opus 5 to do a code review and  cross check with the TAP. All findings are their's. I just verified and shortened them.  IMO only the import error is a blocker.

Otherwise, including my own manual review, LGTM!

securesystemslib/signer/_crypto_signer.py
jku: Yeah I could add a `get_mldsa_payload(data: bytes, version: int) -> bytes` ... maybe that makes sense even if it does not have an obvious good location.

securesystemslib/signer/_crypto_signer.py
jku: I have bumped the floor for `crypto` and `gcpkms` extras

securesystemslib/signer/_gcp_signer.py
jku: Good catch. Of course it wasn't that simple: when public_key_format is passed, GCP no longer returns the pem shortcut field (even though we specified PEM as format): we need to change the response parsing as well.

jku:
@lukpueh let me know if you want to do another review, otherwise I'll just merge as is.

lukpueh (approved):

PR #1137, opened 2026-06-14
Add named constants for key types and signing schemes

### LINKED ISSUE #593: signer: add constants for key types and signing schemes

PratGit1606:
Fixes #593

Adds `securesystemslib/signer/_constants.py` with named constants for all key types (`KEY_TYPE_RSA`, `KEY_TYPE_ECDSA`, `KEY_TYPE_ED25519`) and signing schemes (`ECDSA_SHA2_NISTP256`, `RSASSA_PSS_SHA256`, etc.).
All constants are exported from the public `signer` API.

Hardcoded strings replaced across `_key.py`, `_crypto_signer.py`, `_gcp_signer.py`, `_hsm_signer.py`, `_azure_signer.py`, `_aws_signer.py`, and `__init__.py`. GPG-specific strings left untouched (separate subsystem).

Verified: `tox -e lint,py,purepy,py-no-gpg,py-test-gpg-fails` all pass.

Note: I noticed the stale `refactor-scheme-handling` branch which touches some of the same files. That branch appears to predate the current signer API and was never merged. This PR targets the current main branch. Happy to rebase or adjust if there's overlap with any planned work.


jku:
sorry, this has been sitting here for a while: I think it looks reasonable, I was just hoping to get the earlier #1123 merged before this. If it does not happen soon, I'll switch the order

jku:
The ml-dsa branch has finally landed. I won't ask you to rebase after all this time, I can do the rebase and add the ML-DSA keytype and schemes on top: I'll add a few commits on this branch  today or tomorrow (unless you insist on doing all of it).

jku:
I've merged main in here, and added a small commit with constants for ML-DSA keytype that was just added.

jku (approved):
This seems correct to me: there's no functional changes and the constants seem reasonable.

The scheme names are currently unprefixed. I don't think that leads to issues so I'm ok with that... The only annoying bit is that since they are exported in `signer/__init__.py` it's harder to see what e.g. `ED25519` is -- it's a scheme name but in the context of that file it's not obvious. I think documenting the constants (in the published docs) would help but that can be a followup

jku:
> The only annoying bit is that since they are exported in `signer/__init__.py` it's harder to see what e.g. `ED25519` is -- it's a scheme name but in the context of that file it's not obvious. I think documenting the constants (in the published docs) would help but that can be a followup

Just to document this for anyone else reviewing, I considered these options:
* Make `constants` module public, don't re-export the constants in signer: This way people who use the constants will be more likely to actually see the context in `constants.py`
* use (str, enum) classes
  ```
    class KeyType(str, Enum):
        RSA = "rsa"
        ECDSA = "ecdsa"
        ED25519 = "ed25519"
        MLDSA = "ml-dsa"
  ```
* use separate modules (`keytypes`, `schemes`) instead of generic `constants`

The latter two would make e.g. IDE autocomplete more useful... but especially the (str, Enum) option is annoying for existing code as it means you'd always have to refer to `Scheme.ECDSA_SHA2_NISTP521` which does get really long. 

So I'm leaning towards just documenting the constants, and did that in the last commit: https://python-securesystemslib--1137.org.readthedocs.build/en/1137/signer.html#api-documentation

jku:
This PR has been open for months and my latest changes only added the recently added keytype and improved docs: I will merge today.

PR #1149, opened 2026-06-26
Support Tillitis TKey as a signer

jku:
This includes support for a hardware key that supports ML-DSA. 
* The implementation depends on a tiny library I created a few months ago https://github.com/jku/keylet 
* Tillitis only made the device application available this summer: it works but might not be completely polished 
* Untested on windows and mac -- I expect it to work but I did have to add a linux specific serial hack in keylet so it's not impossible that other platforms require some as well

Tests are ai generated, otherwise this is my own work.

Links:
https://www.tillitis.se/
https://github.com/tillitis/tkey-pq-device-signer

## Design

* TKey is interesting in that it has no long term memory to store a signing key (or even a signing application). The process is
  * USB device is plugged in, host application loads a device app into the device
  * host application can also provide a passphrase given by the user
  * device application constructs a private key by combining a device secret, device app and the passphrase (if any of these change, the key changes). the key can then be used for signing
  * when the device is unplugged, the key is wiped
* This means the device binary is critical: a user *must* be able to keep using the same binary "forever". But upgrading the device binary must be possible as well
* the solution is that the binaries are "pinned" by the securesystemslib private key uri... but they are actually provided in the keylet project (so will not balloon the git repo here): this way a specific private key uri can keep using the same binary but when bugs are fixed, the keylet default device binary can still change
* keylet library supports ed25519 as well... but the 4K payload limit makes it unsuitable for our use


jku:
I'm making this ready for review: I think the code is ready and it's a separate signer implementation of only 180 SLOC.

securesystemslib/signer/_tkey_signer.py
```diff
@@ -0,0 +1,177 @@
...
+        uri = f"{cls.SCHEME}:{path}?{parse.urlencode(query)}"
+
+        return uri, key
+
+    def sign(self, payload: bytes) -> Signature:
+        """Signs payload with Tillitis TKey."""
+
+        # Use TUF-specific message prefix and digest as payload
+        digest = hashlib.sha512(payload).digest()
+        msg = b"tuf" + bytes([1]) + digest
```
jku: there is now a `get_mldsa_payload()` in utils

securesystemslib/signer/_tkey_signer.py
```diff
@@ -0,0 +1,174 @@
...
+        device_path = uri.path if uri.path not in ("", "/") else None
+
+        # Extract query parameters
+        query_params = parse.parse_qs(uri.query)
+
+        digest = None
+        if "digest" in query_params:
+            digest = query_params["digest"][0]
+
+        if digest is None:
```
jku: "if not digest" would actually be better here to catch empty parameter

securesystemslib/signer/_tkey_signer.py
```diff
@@ -0,0 +1,174 @@
...
+            raise UnsupportedLibraryError(TKEY_IMPORT_ERROR)
+
+        self._public_key = public_key
+
+        app = SignApp.load_mldsa(digest=digest)
+        self._tkey = TKeySign(app, device_path, passphrase)
+
+        # key derivation depends on passphrase: compare keys to make sure
+        raw_pubkey = self._tkey.get_pubkey()
+        if public_key.scheme == "ml-dsa-44/1":
```
jku: use the new constants

jku:
I've done self and AI review, will merge soon unless someone leaves a comment with their intent to review: this is an independent signer so bar for entry (and removal) should be fairly low.

jku:
I'm perfectly willing to get reviews/comments on this later too but I want to unblock myself in tuf-on-ci so will merge this now: as mentioned new independent signers should be low danger and low friction to add or remove.

Please open an issue if you have any complaints about this one

Issue #1136, opened 2026-06-11
vault tests are flaky


jku:
Example:
https://github.com/secure-systems-lab/securesystemslib/actions/runs/27339603548/job/80772682291

```
==> Vault server configuration:

Administrative Namespace: 
             Api Address: http://127.0.0.1:8200/
                     Cgo: disabled
         Cluster Address: https://127.0.0.1:8201/
   Environment Variables: GOTRACEBACK, HOME, LANG, LD_LIBRARY_PATH, PATH, PIP_DISABLE_PIP_VERSION_CHECK, PIP_USER, PKG_CONFIG_PATH, PWD, PYTHONHASHSEED, PYTHONIOENCODING, SHLVL, TOX_ENV_DIR, TOX_ENV_NAME, TOX_WORK_DIR, VAULT_ADDR, VAULT_TOKEN, VIRTUAL_ENV, _, __TOX_ENVIRONMENT_VARIABLE_ORIGINAL_CI
              Go Version: go1.26.4
              Listener 1: tcp (addr: "127.0.0.1:8200", cluster address: "127.0.0.1:8201", disable_request_limiter: "false", max_json_array_element_count: "10000", max_json_depth: "300", max_json_object_entry_count: "10000", max_json_string_value_length: "1048576", max_request_duration: "1m30s", max_request_size: "33554432", tls: "disabled")
               Log Level: 
                   Mlock: supported: true, enabled: false
           Recovery Mode: false
                 Storage: inmem
                 Version: Vault v2.0.2, built 2026-06-04T13:18:11Z
             Version Sha: a71d5add7ef4111ec4015e9a81b06388f652e2ac

==> Vault server started! Log data will stream in below:
``` 

```
======================================================================
ERROR: test_vault_import_sign_verify (__main__.TestVaultSigner.test_vault_import_sign_verify)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/securesystemslib/securesystemslib/.tox/local-vault/lib/python3.14/site-packages/urllib3/connection.py", line 204, in _new_conn
    sock = connection.create_connection(
        (self._dns_host, self.port),
    ...<2 lines>...
        socket_options=self.socket_options,
    )
  File "/home/runner/work/securesystemslib/securesystemslib/.tox/local-vault/lib/python3.14/site-packages/urllib3/util/connection.py", line 85, in create_connection
    raise err
[...]

PR #1154, opened 2026-07-02
tests: Make vault tests more robust

jku:
AI suggestions to improve test reliability:
* Avoid potential IPv6 issues with URLs: Use 127.0.0.1 instead of localhost
* Make sure init-vault.sh fails early and loudly

Maybe helps with #1136


lukpueh (approved):
Let's give it a try... Thanks!

Issue #1155, opened 2026-07-02
HSMSigner installation issues


jku:
Installing `securesystemslib[hsm]` is fragile:
* user has to install a pkcs11 module in a place where we can find it
* pykcs11lib does not provide binaries for basically any interesting platforms (not for linux and not for mac on arm): compiling pykcs11lib is known to be problematic so `pip install securesystemslib[hsm]` can fail in very user hostile ways

I don't have great solutions. Listing some options here:

* Use `pyscard` instead of `pykcs11lib`: This would support PIV as before, just not through PKCS11 making this a little simpler (the pkcs11 module would not longer be needed). Unfortunately the project is maintained in the same way as PyKCS11: binaries are mostly not available so pip install will still lead to compilation
 * Use `python-pkcs11` instead of `pykcs11lib`: This is still PKCS11 (so the module needs to be available) but this project seems to generally provide binaries and the API seems more reasonable for our use case
* Use `fido2`: this would be possible in pure Python but is not PIV at all and *requires a custom signature format*: this is not appealing at all due to the verification compatibility issues

Out of these `python-pkcs11` looks most appealing on paper: the API matches our usage *much* better than PyKCS11lib does: I suspect our code would look significantly better afterwards and they seem to provide pre-built binaries. They also use softHSM for testing so I suspect our tests would just work.

CC @kommendorkapten

PR #1156, opened 2026-07-03
HSMSigner: Switch to python-pkcs11

jku:
Switch PKCS11 dependency from `PyKCS11` to `python-pkcs`. The driver for this change is `PyKCS11` being a pain in the behind to install: #1155.

I did this just as a test after getting annoyed by the pykcs11lib installation troubles... but the code looks so much better that I think it will be worth it:
* Main advantage is binary releases: python-pkcs11 does those making installation much more likely to succeed
* Installing also no longer needs swig and C++ tool chain -- this is great for various reasons
* python-pkcs11 is also a *much* better API for our use case: we can simplify basically all code dealing with pkcs11 data

Patch was originally AI produced but I've pretty much rewritten it since then. 

## API Changes

`HSMSigner()` and `HSMSigner.import_()` now take an optional `token_label:str` instead of `token_filter:dict`. This is an API break but it's unlikely to affect users as typical `import_()` call is done without arguments and the constructor is not used directly.


jku:
> HSMSigner() and HSMSigner.import_() now take an optional token_label:str instead of token_filter:dict. This is an API break but it's unlikely to affect users as typical import_() call is done without arguments and the constructor is not used directly.

We could maybe avoid this change but I think it is the right call:  in practice we did not support anything else than label so this should be a safe change WRT existing privkeyuris

securesystemslib/signer/_hsm_signer.py
```diff
@@ -153,124 +121,45 @@ def public_key(self) -> SSlibKey:
...
-                slots.append(slot)
-
-        if len(slots) != 1:
-            raise ValueError(
-                f"Found {len(slots)} cryptographic tokens matching filter {filters}"
+    def _find_token(token_label: str | None = None) -> pkcs11.Token:
+        """Return the PKCS#11 token that matches token_label."""
+        lib_path = os.environ.get("PYKCS11LIB")
+        if not lib_path:
+            raise ValueError("PYKCS11LIB environment variable must be set")
```
jku, line 132: this was previously done by pykcs11lib, I preserved the mechanism (and the env var name that is now a bit odd).

The reason the library path is not just an argument is that `Signer.from_priv_key()` is the preferred way to construct signers: we can't add the argument there. We could consider other choices here:
* A class variable `securesystemslib.signer.HSMSigner.lib_path` that can be set before use would work... but I think the env variable is not much uglier (and should anyway be supported for backwards compat). 
* We could change the env variable name... but would likely want to support the old one as well
* We could also consider adding some platform specific fallbacks (as is currently done in tuf-on-ci)... 

but in the end these decisions seem mostly unrelated to switching the dependency -- the only  reason to do something in this PR is that the env var name is now a little weird.

jku:
I've added an extra feature from #1109 (try authenticating if pubkey is not found without it): we can still merge #1109 to main before this one: I added the changes here already since the code is different enough that actual merging is not going to work right

jku:
Eh, I'm happy with this: marking ready for review

tests/test_hsm_signer.py
```diff
@@ -90,31 +88,35 @@ def setUpClass(cls):
...
+                "--init-token",
+                "--slot",
+                "0",
+                "--label",
+                cls.token_label,
+                "--so-pin",
+                "abcd",
+                "--pin",
+                cls.hsm_user_pin,
+            ],
```
jku, line 104: We shell out "softhsm2-util"  in the test setup now because unlike PyKCS11, python-pkcs only wraps methods that are needed for actual client functionality (and this does not include things like `C_InitToken`)

jku:
If anyone would like to test on Mac, here's some instructions  (I know the equivalent of this succeeds on CI ... but that is true for current main as well and users have had problems installing that because of PyKCS11):

**install test**
```
git clone https://github.com/jku/securesystemslib.git
cd securesystemslib
git checkout pkcs-switcheroo

# create virtualenv, install securesystemslib as editable with "hsm" extra
python3 -m venv .venv
source .venv/bin/activate
pip install -e ".[hsm]"
```
this should succeed, should not compile anything and should not take long

**running the HSM test**

(I'm assuming you have homebrew)

```
# install softhsm for HSM emulation
brew install softhsm

# In the same virtualenv, run hsm tests with the softhsm PKCS module in use
PYKCS11LIB=$(brew --prefix softhsm)/lib/softhsm/libsofthsm2.so python3 -m unittest tests/test_hsm_signer.py
```
this should run 3 tests, should not skip any of them and should succeed.

securesystemslib/signer/_hsm_signer.py
```diff
@@ -389,22 +262,14 @@ def sign(self, payload: bytes) -> Signature:
...
-        r = int.from_bytes(signature[:r_s_len], byteorder="big")
-        s = int.from_bytes(signature[r_s_len:], byteorder="big")
-
-        # Create an ASN.1 encoded Dss-Sig-Value to be used with pyca/cryptography
-        dss_sig_value = binascii.hexlify(encode_dss_signature(r, s)).decode("ascii")
-
-        return Signature(self.public_key.keyid, dss_sig_value)
+        token = self._find_token(self.token_label)
+
+        with token.open(rw=True, user_pin=pin) as session:
```
jku: AI tells me  signing is supported in a Read-Only User session

securesystemslib/signer/_hsm_signer.py
```diff
@@ -159,125 +120,60 @@ def public_key(self) -> SSlibKey:
...
+            raise ValueError("PYKCS11LIB environment variable must be set")
+        try:
+            return pkcs11.lib(lib_path).get_token(
+                token_label=token_label, token_flags=pkcs11.TokenFlag.TOKEN_INITIALIZED
             )
-
-        return slots[0]
+        except NoSuchToken:
+            label_str = f" for label {token_label}" if token_label else ""
+            raise ValueError(f"No PKCS#11 token found{label_str}")
```
jku, line 139: AI says MultipleTokensReturned is possible too, same goes for MultipleObjectsReturned when calling get_key()

securesystemslib/signer/_hsm_signer.py
jku: should update this

securesystemslib/signer/_hsm_signer.py
jku: this seems to be true

jku:
So I did self review with AI: Found minor issues that have now been fixed. 

I will merge this now bypassing usual review rules (it's been six weeks) please file new issues if you see something wrong

PR #1158, opened 2026-07-03
Drop chmod +x execute bit on *.py scripts

jas4711:
Hi! I help maintain this for Debian, and our QA tooling noticed this. Thanks.


jku (approved):
Thanks

PR #1163, opened 2026-07-08
Make Key, Signature and Envelope actually hashable

arpitjain099:

While reading through the signer code I noticed that the `__hash__` methods added in "lint: Make ruff happy" (d87eb73) never actually work. They drop the objects dict fields (`Key.keyval`/`unrecognized_fields`, `Signature.unrecognized_fields`, `Envelope.signatures`) straight into the hashed tuple, so any `hash()` call raises `TypeError: unhashable type: 'dict'`. That means Key, Signature and Envelope still cant go in a set or be used as dict keys, which is the thing defining `__hash__` was meant to enable.

The fix adds `make_hashable` to `securesystemslib/_internal/utils.py`, which recursively converts dicts to frozensets of their items and lists to tuples, and uses it on the dict fields of all three classes. Converting structurally rather than serializing means the result inherits Python's own equality for the leaf values, so objects that compare equal still hash equal.

An earlier revision of this PR used `json.dumps(..., sort_keys=True)` instead. That was wrong: `1`, `1.0` and `True` compare equal in Python but serialize differently, so equal objects came out with different hashes, and `json.dumps` raises on a field value that is not JSON serializable. Thanks to @jku for catching it.

Tests cover all three types plus the equal-values-hash-equally cases and a non-JSON field value; they fail on main with the TypeError and pass with the change, and the rest of test_signer.py / test_dsse.py stay green.

No linked issue, happy to open one if you would prefer to track it separately.

### DIFF
--- a/securesystemslib/_internal/utils.py
+++ b/securesystemslib/_internal/utils.py
@@ -2,6 +2,29 @@
 
 import base64
 import binascii
+from typing import Any
+
+
+def make_hashable(value: Any) -> Any:
+    """Return a hashable equivalent of a dict or list, for use in __hash__
+
+    Dicts become frozensets of their items and lists become tuples, both
+    recursively. Values that are already hashable are returned as they are, so
+    the result compares equal whenever the input does, which is what __hash__
+    needs.
+
+    Arguments:
+        value: Value to convert
+
+    Returns:
+        A hashable equivalent of the value
+    """
+
+    if isinstance(value, dict):
+        return frozenset((k, make_hashable(v)) for k, v in value.items())
+    if isinstance(value, (list, tuple)):
+        return tuple(make_hashable(v) for v in value)
+    return value
 
 
 def b64enc(data: bytes) -> str:
--- a/securesystemslib/dsse.py
+++ b/securesystemslib/dsse.py
@@ -43,7 +43,13 @@ class Envelope:
         )
 
     def __hash__(self) -> int:
-        return hash((self.payload, self.payload_type, self.signatures))
+        return hash(
+            (
+                self.payload,
+                self.payload_type,
+                frozenset(self.signatures.items()),
+            )
+        )
 
     @classmethod
     def from_dict(cls, data: dict) -> Envelope:
--- a/securesystemslib/dsse.py
+++ b/securesystemslib/dsse.py
@@ -6,7 +6,7 @@ import logging
 from typing import Any
 
 from securesystemslib import exceptions
-from securesystemslib._internal.utils import b64dec, b64enc
+from securesystemslib._internal.utils import b64dec, b64enc, make_hashable
 from securesystemslib.signer import Key, Signature, Signer
 
 logger = logging.getLogger(__name__)
@@ -47,7 +47,7 @@ class Envelope:
             (
                 self.payload,
                 self.payload_type,
-                frozenset(self.signatures.items()),
+                make_hashable(self.signatures),
             )
         )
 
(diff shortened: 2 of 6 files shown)


jku (changes requested):
sorry I only got to this now.

I agree the issue is real.

That said, I don't like the json use here. it feels unnecessary and theoretically even wrong (think of the differences in number types for example).

Some possible routes:
1. Just don't implement hash at all -- if it has never worked maybe it was not that important? I'm sure we can deal with the linter issue some other way
2. just hash the identifying parts: IIRC there is no requirement for hash to cover 100% of what `eq` covers. Leaving unrecognised_fields out seems reasonable, but leaving keyval out seems a bit sketchy... so this is likely not a solution
3. have a helper to convert mutable containers to hashable containers

AI says this would work for point 3:

```
def _make_hashable(val: Any) -> Any:
    """Recursively convert unhashable containers (dict, list) to hashable equivalents."""
    if isinstance(val, dict):
        return frozenset((k, _make_hashable(v)) for k, v in val.items())
    if isinstance(val, (list, tuple)):
        return tuple(_make_hashable(v) for v in val)
    return val

```

Do you have an opinion?

arpitjain099:
No worries, and yes: you are right that the json use is wrong, not just inelegant. I have pushed option 3.

The concrete problem is that Python compares `1`, `1.0` and `True` as equal, so `{"extra": 1}` and `{"extra": True}` are equal dicts that `json.dumps` renders as `1` and `true`. Two equal objects with different hashes, which is the one rule `__hash__` has to keep. Measured on `Signature`:

```
                equal   json hash equal   helper hash equal
int vs bool     True    False             True
int vs float    True    False             True
nested in list  True    False             True
```

There is a second problem I had missed: `json.dumps` raises `TypeError: Object of type bytes is not JSON serializable` for a non-JSON value in `unrecognized_fields`, so my patch swapped one `TypeError` for another on that path.

On your three routes:

1 is defensible and I would not argue against it. Nothing can depend on the current behaviour since it has never worked, and these objects are mutable, so any `__hash__` is a footgun if someone puts a `Key` in a set and then mutates `keyval`. If you would rather drop `__hash__` and silence the linter, say so and I will redo the PR that way.

2 does not work for the reason you suspected. `keyval` is the identity of the key, so leaving it out means two different keys hash the same, and `keyval` is exactly the dict that makes hashing fail today.

3 is what I went with. The helper is correct for this because it converts structurally rather than textually, so it inherits Python's own equality for the leaves and the invariant holds by construction. I put it in `securesystemslib/_internal/utils.py` as `make_hashable` since all three classes need it.

One detail worth flagging: `Envelope.signatures` is a dict, not a list, so the helper's frozenset-of-items is the right conversion there. `Envelope.__eq__` compares those dicts, which is order-insensitive, and frozenset matches that. Lists become tuples, which is order-sensitive and matches list equality.

Known limit: a value that is itself unhashable and not a dict, list or tuple (a `set`, say) still raises. That cannot come from `from_dict` parsing, only from someone assigning it directly, and I would rather not add `isinstance` branches for cases the deserializer cannot produce. Easy to extend later if it ever comes up.

Added two regression tests for the cases above: the equal-values-hash-equally table and a `bytes` field value. Both fail with the json version.

`tests/test_signer.py` and `tests/test_dsse.py` pass, `ruff check`, `ruff format --check` and `mypy` are clean. Two pre-existing failures in my environment are unrelated to the change and reproduce on a pristine `main`: `test_sphincs` needs `pyspx`, and mypy cannot find the `sigstore` stubs.

I also rebased onto `main`, since the branch had drifted since July.

jku (approved):
lgtm, thank you

PR #1171, opened 2026-07-29
Drop SHA-224 from default verification registry

### LINKED ISSUE #1164: Drop sha-224 from default verify set
NIST is taking their time with _Transitioning the Use of Cryptographic Algorithms and Key Lengths_ but current revision makes SHA-224 "legacy" by 2030 (https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-131Ar3.ipd.pdf) and other similar orgs have set similar deadlines.

I propose we drop `rsassa-pss-sha224` and `rsa-pkcs1v15-sha224` from the default set of supported verification keys (`KEY_FOR_TYPE_AND_SCHEME`): securesystemslib should offer a smart default installation and as most users will just use the default set, we should actively avoid having a bloated default key registry. We can leave the support in code so the theoretical users who really need the keys can add them back to the supported set.

ECD5A:

This removes `rsassa-pss-sha224` and `rsa-pkcs1v15-sha224` from the default `KEY_FOR_TYPE_AND_SCHEME` registry.

As a result, `Key.from_dict()` no longer accepts SHA-224 RSA schemes by default. The underlying SHA-224 implementation remains available through `SSlibKey`, so callers that still require these legacy schemes can explicitly register or construct them.

The tests verify that:

- both SHA-224 schemes are rejected by the default registry;
- explicitly registering either scheme restores `Key.from_dict()` support;
- existing SHA-224 signature vectors continue to verify through `SSlibKey`;
- invalid payloads still fail signature verification.

Closes #1164


jku:
Cheers, this looks good to me. We have a bit of a PR merge bottleneck right now and I'd rather merge in order so it might be a while until we get to this one but I don't see any issues here, thanks.

jku:
ok, PR bottleneck is finally moving: there is now a conflict here as the string constants are now actual constants. 

No need to change every instance of those strings in the test code but resolving the conflict in `__init__.py` would be appreciated.

jku (approved):
lgtm, thank you

PR #1181, opened 2026-08-10
CryptoSigner: support key generation for nistp384 and nistp521 ecdsa schemes

### LINKED ISSUE #1153: CryptoSigner onlys supports ecdsa-sha2-nistp256 as ecdsa scheme
* CryptoSigner.generate_ecdsa() only supports ecdsa-sha2-nistp256, this is not too bad since it's just convenient helper, however
* CryptoSigner() will throw if the scheme is not "ecdsa-sha2-nistp256": I think this should not happen, no reason not to support the larger hashes too 

For background, I was planning to extend the TUF Conformance tests on key support... But that's going to be tricky if securesystemslib doesn't support those keys

ArockiaRajamanickam:

`SSlibKey._verify()` already handles all three NIST ecdsa schemes, `SSlibKey.from_crypto()` already derives `ecdsa-sha2-nistp384` and `ecdsa-sha2-nistp521` from the curve, and both combinations are registered in `signer/__init__.py`. `CryptoSigner` was the only piece that refused them, so a key that `from_crypto()` produced could not be used to sign:

```python
priv = generate_private_key(SECP384R1())
pub = SSlibKey.from_crypto(priv.public_key())   # scheme: ecdsa-sha2-nistp384
CryptoSigner(priv, pub)
# ValueError: unsupported public key ecdsa/ecdsa-sha2-nistp384
```

Both bullets of the issue are addressed, in one commit but two independent hunks:

1. `CryptoSigner()` accepts `ecdsa-sha2-nistp384` and `ecdsa-sha2-nistp521`, using the curve and hash that `SSlibKey._verify()` uses for each.
2. `generate_ecdsa()` takes an optional `scheme`, defaulting to `ecdsa-sha2-nistp256` so existing calls are unaffected. This follows `generate_rsa()` and `generate_mldsa()`, which already take a `scheme`.

Happy to drop the second one if you would rather keep this to the part you called out as the real problem.

### A related bug this closes

@lukpueh's point on the issue, that the scheme fixes the key type rather than being freely combinable, turns out to be load-bearing here. `assert_type()` only checked for `EllipticCurvePrivateKey`, not for the curve, so on current main this is accepted:

```python
p384 = generate_private_key(SECP384R1())
k = SSlibKey.from_crypto(p384.public_key())
signer = CryptoSigner(p384, SSlibKey(k.keyid, "ecdsa", "ecdsa-sha2-nistp256", k.keyval))
sig = signer.sign(b"data")   # signs happily
```

The signature can never verify, because `SSlibKey._verify()` calls `_validate_curve()`. Extending the scheme check without also checking the curve would have widened that from one mismatched pair to six, so the constructor now validates the curve and raises `bad curve secp384r1 for ecdsa-sha2-nistp256`, matching the wording `_verify()` already uses.

For the same reason the deprecated keytype check is now `keytype in ["ecdsa", scheme]` rather than a fixed list. That mirrors `_verify()` exactly: it accepts a legacy keytype only when it equals the scheme, so `keytype=ecdsa-sha2-nistp256` with `scheme=ecdsa-sha2-nistp521` stays rejected.

### Notes on the implementation

The scheme to curve/hash mapping is built inside `_get_ecdsa_curve_and_hash()` rather than at module scope on purpose. Those symbols come from the guarded `try:` import block, so a module-level dict would raise `NameError` on import and break the `purepy` environment. I checked that case specifically.

I did not add a `CHANGELOG.md` entry, since there is no `Unreleased` section and the entries carry PR numbers, so it looks like you write those at release time. Glad to add one if you want it in the PR.

### Tests

`tests/data/pems/` already had `ecdsa_secp384r1_private.pem` and `ecdsa_secp521r1_private.pem`, so the two curves are added to the existing `key_info` table in `TestCryptoSigner`, which extends `test_init` and `test_sign` to cover them. Two new tests cover the `scheme` argument and the curve mismatch.

Reverting only `_crypto_signer.py` and keeping the tests fails 4 tests in `TestCryptoSigner`, including both new ones. With the change, `tests/` goes from 78 passed to 80 passed against the same single pre-existing failure in my environment (`TestSphincs::test_sphincs`, no pyspx installed; `test_hsm_signer.py` also fails to collect here, both identically on a clean checkout). `ruff format`, `ruff check` and `mypy securesystemslib` are clean, mypy back to the same 6 pre-existing sigstore stub errors as on main. `tests.check_public_interfaces` passes in a venv with no cryptography installed.

I used an AI assistant while working on this. The reproduction, the curve mismatch finding and the decision to flag the `generate_ecdsa()` hunk as separable are mine, and I ran every check above myself.

Fixes #1153

### DIFF
--- a/securesystemslib/signer/_crypto_signer.py
+++ b/securesystemslib/signer/_crypto_signer.py
@@ -1,13 +1,18 @@
 """Signer implementation for pyca/cryptography signing."""
 
+from __future__ import annotations
+
 import logging
 import os
 from dataclasses import astuple, dataclass
+from typing import cast
 from urllib import parse
 
 from securesystemslib.exceptions import UnsupportedLibraryError
 from securesystemslib.signer._constants import (
     ECDSA_SHA2_NISTP256,
+    ECDSA_SHA2_NISTP384,
+    ECDSA_SHA2_NISTP521,
     ED25519,
     KEY_TYPE_ECDSA,
     KEY_TYPE_ED25519,
@@ -35,6 +40,9 @@ try:
     from cryptography.hazmat.primitives.asymmetric.ec import (
         ECDSA,
         SECP256R1,
+        SECP384R1,
+        SECP521R1,
+        EllipticCurve,
         EllipticCurvePrivateKey,
     )
     from cryptography.hazmat.primitives.asymmetric.ec import (
@@ -63,6 +71,8 @@ try:
     from cryptography.hazmat.primitives.asymmetric.types import PrivateKeyTypes
     from cryptography.hazmat.primitives.hashes import (
         SHA256,
+        SHA384,
+        SHA512,
         HashAlgorithm,
     )
     from cryptography.hazmat.primitives.serialization import (
@@ -82,13 +92,13 @@ logger = logging.getLogger(__name__)
 
 @dataclass
 class _RSASignArgs:
-    padding: "AsymmetricPadding"
-    hash_algo: "HashAlgorithm"
+    padding: AsymmetricPadding
+    hash_algo: HashAlgorithm
 
 
 @dataclass
 class _ECDSASignArgs:
-    sig_algo: "ECDSA"
+    sig_algo: ECDSA
 
 
 @dataclass
@@ -96,12 +106,34 @@ class _NoSignArgs:
     pass
 
 
-# for backwards compat: use when spec-deprecated keytype ecdsa-sha2-nistp256
-# should be accepted in addition to "ecdsa"
-_ECDSA_KEYTYPES = [KEY_TYPE_ECDSA, ECDSA_SHA2_NISTP256]
+# keep in sync with _get_ecdsa_curve_and_hash() below
+_ECDSA_SCHEMES = [
+    ECDSA_SHA2_NISTP256,
+    ECDSA_SHA2_NISTP384,
+    ECDSA_SHA2_NISTP521,
+]
 
 
-def _get_rsa_padding(name: str, hash_algorithm: "HashAlgorithm") -> "AsymmetricPadding":
+def _get_ecdsa_curve_and_hash(
+    scheme: str,
+) -> tuple[type[EllipticCurve], HashAlgorithm]:
+    """Helper to return curve and hash algorithm for an ecdsa scheme.
+
+    An ecdsa scheme fixes both, and the pairs must agree with the ones
+    SSlibKey._verify() uses, or signatures will not verify.
+    """
+    # built here and not at module scope, so that importing this module
+    # still works when pyca/cryptography is not installed
+    curves_and_hashes: dict[str, tuple[type[EllipticCurve], HashAlgorithm]] = {
+        ECDSA_SHA2_NISTP256: (SECP256R1, SHA256()),
+        ECDSA_SHA2_NISTP384: (SECP384R1, SHA384()),
+        ECDSA_SHA2_NISTP521: (SECP521R1, SHA512()),
+    }
+
+    return curves_and_hashes[scheme]
+
+
+def _get_rsa_padding(name: str, hash_algorithm: HashAlgorithm) -> AsymmetricPadding:
     """Helper to return rsa signature padding for name."""
     padding: AsymmetricPadding
     if name == "pss":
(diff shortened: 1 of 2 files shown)


securesystemslib/signer/_crypto_signer.py
```diff
@@ -77,9 +83,31 @@ class _NoSignArgs:
...
+_ECDSA_SCHEMES = [
+    "ecdsa-sha2-nistp256",
+    "ecdsa-sha2-nistp384",
+    "ecdsa-sha2-nistp521",
+]
+
+
+def _get_ecdsa_curve_and_hash(
+    scheme: str,
+) -> "tuple[type[EllipticCurve], HashAlgorithm]":
```
jku: I *think* you can avoid the quotes here by adding `from __future__ import annotations` to top of file, I think it's just missing from this specific file.

(if I remember incorrectly then disregard)

securesystemslib/signer/_crypto_signer.py
```diff
@@ -77,9 +83,31 @@ class _NoSignArgs:
...
 
-# for backwards compat: use when spec-deprecated keytype ecdsa-sha2-nistp256
-# should be accepted in addition to "ecdsa"
-_ECDSA_KEYTYPES = ["ecdsa", "ecdsa-sha2-nistp256"]
+# keep in sync with _get_ecdsa_curve_and_hash() below
+_ECDSA_SCHEMES = [
+    "ecdsa-sha2-nistp256",
+    "ecdsa-sha2-nistp384",
+    "ecdsa-sha2-nistp521",
+]
```
jku, line 114: we just added constants for these:
```
from securesystemslib.signer._constants import (
    KEY_TYPE_ECDSA,
    ECDSA_SHA2_NISTP256,
    ECDSA_SHA2_NISTP384,
    ECDSA_SHA2_NISTP521,
)
```
would you mind using those? I don't really mind if tests use the plain strings but this file should not

jku (changes requested):
Thanks, this looks nice. I just merged a PR that switches the key type and scheme strings to constants and that obviously conflicts with this PR: sorry about that.

If you can fix the conflicts and use the constants, that would be appreciated. Rebase and force push is fine if you prefer that.

securesystemslib/signer/_crypto_signer.py
ArockiaRajamanickam: Done. `_ECDSA_SCHEMES` and the curve/hash mapping now use `ECDSA_SHA2_NISTP256/384/521`, and the keytype check uses `KEY_TYPE_ECDSA`. No plain scheme strings left in this file. I left the tests on plain strings as you said you did not mind.

securesystemslib/signer/_crypto_signer.py
ArockiaRajamanickam: You remembered correctly. `_crypto_signer.py` was the only module in `signer/` without it, apart from `__init__.py`, `_constants.py` and `_crypto_utils.py`, so I added it and dropped the quotes.

One consequence worth flagging: once the import is there, ruff UP037 requires removing the quotes from the other annotations in the file too, so the diff picks up 11 lines that are otherwise unrelated to this fix. Happy to pull that out into its own PR and keep this one to the ecdsa change if you would rather review them separately.

I checked that the no-cryptography path still imports cleanly, since PEP 563 is what makes the unquoted `EllipticCurve` annotations safe there.

ArockiaRajamanickam:
Rebased onto main and force pushed, no conflicts left. No need to apologise, the constants are a clear improvement to read against.

Both review points are addressed, each answered inline:

- the file now uses `KEY_TYPE_ECDSA` and `ECDSA_SHA2_NISTP256/384/521` throughout, no plain scheme strings left in `_crypto_signer.py`
- added `from __future__ import annotations`, which you were right about, it was the only module in `signer/` missing it

One thing to look at: adding that import means ruff UP037 also requires dropping the quotes from the 11 other annotations already in the file, so the diff is a bit wider than the ecdsa change alone. Say the word and I will move that into a separate PR.

Re-verified after the rebase on Python 3.10, 3.12 and 3.14, with the pinned ruff 0.16.0 and mypy 2.3.0 from `requirements-lint.txt`: `ruff format`, `ruff check` and `mypy securesystemslib` clean, 87 tests passing, and `tests.check_public_interfaces` still passes in a venv with no cryptography installed. Reverting only `_crypto_signer.py` and keeping the tests fails 4 tests in `TestCryptoSigner`.

jku (approved):
This looks complete to me, will be useful to extend the key type tests in tuf-conformance. Thanks for the new tests.

PR #1182, opened 2026-08-12
Remove SpxSigner, SpxKey

### LINKED ISSUE #1141: Remove or update SpxSigner
We could consider dropping or improving SpxSigner:
* current dependency `pyspx` looks unmaintained. It also does not seem to work on windows 
* SPHINCS+ became SLH-DSA (FIPS 205), I'm unsure if pyspx is compatible with that
* `cryptography` does not support SLH-DSA but this might change now that latest openssl does support it
* There are alternatives like https://github.com/colinxu2020/slhdsa (LGPL) and https://github.com/open-quantum-safe/liboqs-python (C code), I have not researched them

I think the baseline for algorithms like this should probably be support in `cryptography` not in a 3rd party dependency...

jku:
The SPHINCS signer was added as an experimental post-quantum signer some years ago. Situation is now a little different:
* The underlying `pyspx` library seems to be unmaintained **<- this is the main argument for removing**
* SPHINCS did end up in NIST algorithms as SLH-DSA but so did ML-DSA -- the signature and key sizes in ML-DSA are much more acceptable for this use case
* ML-DSA is now available directly from `cryptography` (and is implemented in CryptoSigner/SSLibKey)

There should be no users for SpxSigner and SpxKey was never part of the default verify keyset so I feel comfortable just removing them without deprecation: anyone who wants to can easily copy them and use them as external signer/key. I'm not against supporting SLH-DSA but would like the support to come from a supported library, preferably `cryptography`.

Fixes #1141


lukpueh (approved):

jku:
I think I'll leave this after 1.5 release

PR #1190, opened 2026-08-20
TKeySigner: raise KeyMismatchError for mismatched keys

### LINKED ISSUE #1189: TKeySigner: better error for wrong keys
TKeySigner does this if signer does not match given pubkey

```
        if key.keyval != self.public_key.keyval:
            raise RuntimeError(
                "TKey public key does not match: This could mean incorrect Passphrase."
            )
```

This really should be something more useful than a RuntimeError since it's one of the few errors apps will want to catch.

ValueError would be the basic choice but maybe a specific new error makes sense (if we don't have a KeyMismatchError or something yet)

ECD5A:

Add a dedicated `KeyMismatchError` for TKey public-key mismatches. The exception subclasses `ValueError`, so callers can catch the specific mismatch without losing the conventional invalid-value behavior.

The existing mismatch test now verifies the specific exception type.

Testing:
- `python -m unittest tests.test_tkey_signer`
- `ruff format --check securesystemslib tests`
- `ruff check securesystemslib tests`
- `mypy securesystemslib`
- core test suite: 98 passed, 41 skipped

Fixes #1189

### DIFF
--- a/securesystemslib/exceptions.py
+++ b/securesystemslib/exceptions.py
@@ -34,6 +34,10 @@ class UnsupportedLibraryError(Error):
     """Indicate that a supported library could not be located or imported."""
 
 
+class KeyMismatchError(ValueError):
+    """Indicate that a private key does not match the expected public key."""
+
+
 class StorageError(Error):
     """Indicate an error occured during interaction with an abstracted storage
     backend."""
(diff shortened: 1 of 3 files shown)


jku (approved):
Yeah, I agree with this: it's especially useful for TKeySigner because of its design... but could be useful in other Signers too:
* CryptoSigner could also do similar check in __init__()
* HSMSigner *could* get the public key from device to check -- but this likely needs careful testing
* likewise the cloud KMSs signer could fetch public key at init time... but this is an extra HTTP request, I'm not completely  sure we want it

I'll approve this and suggest we leave the other possible signers to a followup issue

PR #1191, opened 2026-08-20
ci: publish GitHub releases from drafts

### LINKED ISSUE #1098: Use Immutable Releases
copy-pasting from python-tuf

---

There's a checkbox in the settings Enable release immutability: we should check it.

The docs are just vague enough that I'm not 100% sure if our current released workflow works with it:
* Description is "Disallow assets and tags from being modified once a release is published"
* We don't modify assets but do add them after the initial publish: I think that might still be a violation

Maybe best to tweak the release workflow so the initial publish makes the release a draft (and the final publish makes it non-draft): then we should be good to go.

Doing this also gives us  release attestations for free

ECD5A:
**Description...**:

Prepare the release workflow for immutable releases by creating the GitHub release as a draft, awaiting every asset upload, and publishing it only after the PyPI step succeeds.

Pass the release ID through the step environment instead of interpolating it into the JavaScript body. Await the final API call so publication failures are reported by the workflow.

Testing:
- actionlint .github/workflows/cd.yml
- zizmor --persona=regular -q .
- YAML parsing and JavaScript syntax checks for both github-script blocks

Repository administrators still need to enable release immutability after this workflow change is merged.

Fixes #1098

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -64,24 +64,25 @@ jobs:
         uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
         with:
           script: |
-            fs = require('fs')
-            res = await github.rest.repos.createRelease({
+            const fs = require('fs')
+            const res = await github.rest.repos.createRelease({
               owner: context.repo.owner,
               repo: context.repo.repo,
               name: process.env.REF_NAME + '-rc',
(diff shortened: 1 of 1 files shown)


jku (approved):
lgtm, thanks. 

I'm sure there's still something I'm overlooking and we find out when release fails but... only one way to find out. I'll switch on immutable releases in settings, we'll see results on next release

Issue #1192, opened 2026-08-20
Use KeyMismatchError in more signers


jku:
TKeySigner uses KeyMismatchError when public key and private do not match on `__init__()`. We could do the same elsewhere. This is not a complete design but as notes:
* [x] CryptoSigner could do similar check in `__init__()`
* HSMSigner *could* get the public key from device just like TKeySigner -- but this likely needs careful testing with a yubikey or similar
* likewise the cloud KMSs signers could fetch public key at init time... but this is an extra HTTP request and KMS signing rarely goes wrong because of human error (unlike most other mechanisms), so I'm not sure we want that extra request

_Originally posted by @jku in https://github.com/secure-systems-lab/securesystemslib/pull/1190#pullrequestreview-4982301180_

PR #1196, opened 2026-08-25
CryptoSigner: reject mismatched key pairs

ECD5A:

Validate that a `CryptoSigner` private key matches the supplied `SSlibKey` before the signer becomes usable. A mismatch now raises `KeyMismatchError`, matching the existing `TKeySigner` behavior.

Regression coverage exercises RSA, ECDSA, Ed25519, and ML-DSA. HSM and cloud-KMS signers remain outside this focused change because checking those keys can require device access or network requests.

Tests:

- `python -m pytest tests/test_signer.py tests/test_dsse.py -q`
- `python -m ruff format --check securesystemslib tests`
- `python -m ruff check securesystemslib tests`

Addresses the `CryptoSigner` portion of #1192.

### DIFF
--- a/securesystemslib/signer/_crypto_signer.py
+++ b/securesystemslib/signer/_crypto_signer.py
@@ -8,7 +8,7 @@ from dataclasses import astuple, dataclass
 from typing import cast
 from urllib import parse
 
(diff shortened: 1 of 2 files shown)


jku (approved):
Thanks!

PR #1197, opened 2026-08-25
Tweak cryptosigner key check

jku:
#1196 added a check to CryptoSigner to verify the given public and private keys are actually a match: python-tuf test suite revealed that the check is a bit too simple: multiple keyvals can actually be the same cryptographic key.

* Use the cryptographic key for comparison instead: this means the check is a little more expensive but also more correct
* Also fix a related Ed25519 key bug (the ed25519 keyval is special, `_crypto_key()` needs to take that into account)

@ECD5A review welcome if you have any comments

### DIFF
--- a/securesystemslib/signer/_crypto_signer.py
+++ b/securesystemslib/signer/_crypto_signer.py
@@ -244,8 +244,7 @@ class CryptoSigner(Signer):
                 f"unsupported public key {public_key.keytype}/{public_key.scheme}"
             )
 
-        derived_public_key = SSlibKey.from_crypto(private_key.public_key())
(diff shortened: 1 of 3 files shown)


ECD5A (approved):
Reviewed against #1196 and python-tuf #3000. Comparing the parsed keys fixes the serialization mismatch without weakening the key-pair check, and the Ed25519 handling looks correct. I also ran the signer tests, Ruff checks, and the affected python-tuf metadata tests locally. No blocking issues from my side.

jku:
Thanks, I'll merge based on this to make python-tuf CI green again.

PR #1200, opened 2026-08-30
Signer: lazy-load optional implementations

### LINKED ISSUE #1198: Assess dynamic/lazy imports for Signers
Currently we import Signers greedily in `securesystemslib.signer`: this is a no-op if the signer-specific extras are not installed...  but when the extras are installed, the dependencies get imported as well. 

Now that we have three cloud signers (all with gigantic SDKs) this is affecting startup time if all three are enabled. This is especially visible in tuf-on-ci e2e tests where the startup cost is paid often, and many signers are available.

We could investigate lazy imports: only import specific signers (or signer dependencies) when the URI for that signer is used with `Signer.from_priv_key_uri()`
* one option is to use PEP562 mechanism (module level `__getattr__()`)
* another is to do in-method imports in every signer -- not impossible since we already have the import error checks at right places

ECD5A:

Load optional signer implementations only when their public class or private-key URI scheme is first used. Direct and wildcard imports remain available, custom registry entries continue to override defaults, and `Signer.from_priv_key_uri()` now loads and caches built-in implementations in a plain dictionary. Importing `Signer` alone no longer loads optional SDKs.

The regression tests cover lazy base imports, direct and wildcard imports, built-in loading and caching, custom overrides, and failure paths.

**Local performance check**

Command: `python -c "from securesystemslib.signer import Signer"`

Measured on Windows with Python 3.14 and the optional signer packages available in the development environment. Seven fresh-process runs were used; the values below are medians and include interpreter startup:

- `main` (parent of this PR): ~0.455 s
- this PR: ~0.167 s
- reduction: ~63%

Absolute timings vary by machine, but this confirms that the optional signer imports were removed from the base import path.

Fixes #1198

### DIFF
--- a/docs/signer.rst
+++ b/docs/signer.rst
@@ -19,8 +19,9 @@ generic *load* methods:
 
 * ``Signer.from_priv_key_uri`` -  Loads any specific signer from a URI. The
   specific signer implementation itself is responsible for the URI format and
-  resolution. To become discoverable, signers and their URI schemes are
-  registered in the ``SIGNER_FOR_URI_SCHEME`` lookup table.
+  resolution. Built-in signers are loaded on demand. Applications can register
+  custom signers or override built-ins in ``SIGNER_FOR_URI_SCHEME``. This table
+  contains custom entries and cached implementations, not all supported schemes.
 
 * ``Key.from_dict`` - Loads any specific key from a serialized format. The
   specific key implementation is responsible for the public key format and
--- a/securesystemslib/signer/__init__.py
+++ b/securesystemslib/signer/__init__.py
@@ -5,9 +5,9 @@ This module provides extensible interfaces for public keys and signers:
 Some implementations are provided by default but more can be added by users.
 """
 
-# ruff: noqa: F401
-from securesystemslib.signer._aws_signer import AWSSigner
-from securesystemslib.signer._azure_signer import AzureSigner
+from importlib import import_module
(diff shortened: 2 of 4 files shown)


jku (commented):
I'll do a proper review later but as initial comment: `_LazySignerRegistry` feels quite complicated.

It's possible I'm overlooking something but this feels simpler:
* Use normal dicts:
    ```
    SIGNER_FOR_URI_SCHEME: dict[str, type] = {}
    _DEFAULT_SIGNERS = {
        "awskms": (<modulename>, "AWSSigner"),
        ...
    }
    ```
* and then do a slightly special lookup in `from_priv_key_uri()`
    ```
    signer = SIGNER_FOR_URI_SCHEME.get(scheme)
    if signer is None:
        if scheme not in _DEFAULT_SIGNERS:
            raise ValueError(f"Unsupported private key scheme {scheme}")
        
        mod_name, cls_name = _DEFAULT_SIGNERS[scheme]
        signer = getattr(importlib.import_module(mod_name), cls_name)
        # Cache signer for future lookups
        SIGNER_FOR_URI_SCHEME[scheme] = signer        
    ```

Feel free to wait until I have time to review properly or consider the above already now

ECD5A:
Thanks — I went with this approach. The registry is now a plain dict, and from_priv_key_uri() loads and caches default signers on demand. I also updated the tests and documentation to reflect the narrower registry semantics.

PR #1204, opened 2026-09-01
Cryptography compat fix

### LINKED ISSUE #1203: dependency confusion wrt cryptography
* We require users to select the extras they need, including `crypto` for cryptography.
* Many users already use cryptography so likely don't include `crypto` extra because it seems to work as is
* but there is a difference:
  * installing `securesystemslib[crypto]` and `cryptography<47` leads to securesystemslib==1.4.0
  * installing `securesystemslib` and `cryptography<47` leads to securesystemslib==1.5.0 **this will not work with the installed cryptography**

This just happened for old versions of sigstore-python: they are all broken now. I think we need to fix this even if it's not strictly our fault -- using extras for cryptography was maybe not a good idea to begin with.

jku:
### Support older cryptography as  we can

Problem:
* We require users to select the extras they need, including `crypto` for cryptography.
* Many users already use cryptography themselves so may not include `crypto` extra because it seems to work without
* but there is a difference:
  * installing `securesystemslib[crypto]` and `cryptography<47` leads to securesystemslib==1.4.0
  * installing `securesystemslib` and `cryptography<47` leads to securesystemslib==1.5.0 **this will not work with the installed cryptography**

Solution:
* Only require ML-DSA types if ml-dsa keys are requested: this make s the code work with older cryptography
* Keep officially requiring cryptography 48: so anyone using the extra gets correct dependency resolution
* Add test env with with cryptography 44 -- the idea is to understand when we break this, not to promise that we support it

The real code change is 50 lines (but they are rather annoying). The rest of it is just test changes.

Fixes #1203

### DIFF
--- a/tests/test_hsm_signer.py
+++ b/tests/test_hsm_signer.py
@@ -7,18 +7,27 @@ import tempfile
 import unittest
 from unittest import mock
 
-import pkcs11
-from asn1crypto.keys import (
-    ECDomainParameters,
-    NamedCurve,
-)
+try:
+    import pkcs11
+    from asn1crypto.keys import (
+        ECDomainParameters,
+        NamedCurve,
+    )
+
+    HAVE_PKCS11 = True
+except ImportError:
+    HAVE_PKCS11 = False
+
 from cryptography.hazmat.primitives.asymmetric.ec import SECP256R1, SECP384R1
 
 from securesystemslib.exceptions import UnverifiedSignatureError
 from securesystemslib.signer import HSMSigner, Signer
 
 
-@unittest.skipUnless(os.environ.get("PYKCS11LIB"), "set PYKCS11LIB to SoftHSM lib path")
+@unittest.skipUnless(
+    HAVE_PKCS11 and os.environ.get("PYKCS11LIB"),
+    "set PYKCS11LIB to SoftHSM lib path and install python-pkcs11",
+)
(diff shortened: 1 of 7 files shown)


tests/test_hsm_signer.py
```diff
@@ -7,18 +7,27 @@
...
+try:
+    import pkcs11
+    from asn1crypto.keys import (
+        ECDomainParameters,
+        NamedCurve,
+    )
+
+    HAVE_PKCS11 = True
+except ImportError:
+    HAVE_PKCS11 = False
```
jku, line 19: this is here because otherwise the test fails in the new min-crypto env

jku:
I've done self- and AI-review as substitute of an actual review -- lgtm. I plan to merge soon as I want to get especially 1205 backport merged and released since it looks like we're technically unable to yank the 1.5.0 release

Phantom9869:
Tested locally with cryptography 50.0.0 on the PR branch — all 95 tests pass. One minor suggestion: a short inline comment on the MLDSA44PublicKey is not None checks in _from_crypto() would help future readers understand why the guard is needed. Otherwise looks good to me.

jku:
We did manage to yank 1.5.0 so this is is not critical to release immediately... but I don't want to keep a yanked release without a fix release out for long so I am merging this

